PR #138, opened 2015-01-29
add subscription to InvoiceLineItem

pandacalculus:
noticed this was missing, and we are assuming we need it to determine the subscription when an `InvoiceLineItem` is a proration adjustment for a subscription (the api doc implies in this case the type of the line item will be `invoiceitem`, the `id` will be an invoice item id, `proration` will be true, and `subscription` will contain the id of the subscription that caused the proration).

### DIFF
--- a/src/main/java/com/stripe/model/InvoiceLineItem.java
+++ b/src/main/java/com/stripe/model/InvoiceLineItem.java
@@ -13,7 +13,8 @@ public class InvoiceLineItem extends StripeObject {
 	Integer quantity;
 	Plan plan;
 	String description;
-  Map<String, String> metadata;
+	Map<String, String> metadata;
+	String subscription;
 
 	public String getId() {
 		return this.id;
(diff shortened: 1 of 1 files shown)


kyleconroy:
@pandacalculus thanks for the PR, but it looks like this change has broken the build. Can you take a look at fixing the `NullPointerException`s?

pandacalculus:
@kyleconroy I can't reproduce these test errors on my fork, and they don't seem to have any correlation with the tiny change in this pull request. The tests pass fine locally (when run using `mvn clean install`) on both Java 7 and Java 8 on OSX.

pandacalculus:
The last green CI build seems to have occurred 6 days ago, and the NPEs were present in build #606 which occurred 6 days ago, triggered by @jimdanz

jimdanz:
Yup, this is our mistake -- we have a test that's currently flappy.  Thanks for the PR!

PR #139, opened 2015-02-18
added structure to support unified customer sources api

maxwells:
cc @canderson, @raycmorgan


jimdanz:
lg. Is there a reason that PaymentSource and DeletedStripeObject need to be interfaces, instead of classes that extend, respectively APIResource and StripeObject?

jimdanz:
Oh, just reread and apparently PaymentSource used to be a class, so I'm assuming there's a reason :)

maxwells:
there's no requirement -- I just preferred it for PaymentSource (having all the prop getters/setters and public interface for a class within its own file). I have no feelings one way or the other about DeletedStripeObject

jimdanz:
kk, sg.  :ship:

PR #143, opened 2015-03-12
Add STRIPE-ACCOUNT header support

danwang:
r? @bkrausz @jimdanz 

Add support for `Stripe-Account` header in `RequestOptions`.

This can currently only be used on methods with `RequestOptions` as the final parameter in their signature (none with `String apiKey`).

There's no tests because there's no way to test the existence of the headers without creating new accounts (if the real API is to be used, in the style of the existing tests) or ripping a lot of things apart (if mocks/captors are to be used). I tested using my own account + a test account in a super simple program:

```
  public static void main(String[] args) {
    Stripe.apiKey = "my test secret key";
    String connectedAccount = "the token of my connected account";

    try {
      ChargeCollection all = Charge.all(params, RequestOptions.builder().setApiKey(Stripe.apiKey).setStripeAccount(connectedAccount).build());
      System.out.println(all);
    } catch (StripeException e) {
      e.printStackTrace();
    }
  }
```

and verified that it returned charges from the connected account.

The issue of testing seems like too big to shave right now, and I'm pretty confident in the correctness of this.

### DIFF
--- a/src/main/java/com/stripe/net/APIResource.java
+++ b/src/main/java/com/stripe/net/APIResource.java
@@ -169,6 +169,9 @@ public abstract class APIResource extends StripeObject {
 		if (options.getIdempotencyKey() != null) {
 			headers.put("Idempotency-Key", options.getIdempotencyKey());
 		}
+		if (options.getStripeAccount() != null) {
+			headers.put("Stripe-Account", options.getStripeAccount());
+		}
 		return headers;
 	}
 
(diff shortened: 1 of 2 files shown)


bkrausz:
Looks like a couple of NPEs in tests: https://travis-ci.org/stripe/stripe-java/jobs/54169281

danwang:
@bkrausz not sure what happened, but I pushed the same diff with a different hash and the tests passed.

The tests are pretty flaky running locally.

bkrausz:
no prob, thanks for re-pushing. lgtm! :sheep:

PR #146, opened 2015-03-16
Added currency to Customer

ramirost:
I just added currency member to Customer object.

Please consider adding this to the next release to be able to detect this Exception before is produced: 

	com.stripe.exception.InvalidRequestException: Can't combine currencies on a single customer. This customer has had a subscription, coupon, or invoice item with currency eur


jimdanz:
thanks!

PR #148, opened 2015-03-16
Dependency inject network requests in APIResource

danwang:
r? @bkrausz 

This isn't really dependency injection, but it's close enough to fix testing so they become useful again.

This is a huge diff, so summary:
- Added a `StripeResponseGetter` interface that has a single `request` method.
- Moved `private static` functions dealing with network requests from `APIResource` to `LiveStripeResponseGetter < StripeResponseGetter`. This is mostly copy and paste, with a few exceptions where shared things stay on `APIResponse`.

TODO
- [x] clean up unnecessary imports

PR #149, opened 2015-03-17
Add account retrieval by ID and support for lists in params

danwang:
r? @jimdanz @bkrausz 

Add account retrieval by ID and add a super basic (stubbed!) unit test. This is an example of how #148 will work.

TODO:
- [x] add more tests for `Account`
- [x] ~~add nested map support~~ (already exists)
- [x] add support for lists
- [x] add support for legal entity
  - [x] verification
  - [x] address
  - [x] dob
  - [x] personal_address
  - [x] additional_owners
- [x] C
- [x] R
- [x] U

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -9,6 +9,7 @@ import com.stripe.net.APIResource;
 import com.stripe.net.RequestOptions;
 
 import java.util.List;
+import java.util.Map;
 
 public class Account extends APIResource {
 	String id;
@@ -22,6 +23,8 @@ public class Account extends APIResource {
 	String country;
 	String timezone;
 	String displayName;
+	Verification verification;
+	LegalEntity legalEntity;
 
 	public String getId() {
 		return id;
@@ -67,18 +70,49 @@ public class Account extends APIResource {
 		return displayName;
 	}
 
+	public Verification getVerification() {
+		return verification;
+	}
+
+	public LegalEntity getLegalEntity() {
+		return legalEntity;
+	}
+
+	public static Account create(Map<String, Object> params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return create(params, (RequestOptions) null);
+	}
+
+	public static Account create(Map<String, Object> params, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.POST, classURL(Account.class), params, Account.class, options);
+	}
+
 	public static Account retrieve()
 			throws AuthenticationException, InvalidRequestException,
 			APIConnectionException, CardException, APIException {
 		return retrieve((RequestOptions) null);
 	}
 
+	/**
+	 * In order to preserve backwards-compatibility, this method does two things.
+	 * If the parameter looks like an API key (starts with sk_), retrieve the
+	 * account resource with no ID parameter set. Otherwise, use the String
+	 * parameter as the account ID.
+	 */
 	@Deprecated
-	public static Account retrieve(String apiKey)
+	public static Account retrieve(String apiKeyOrAccountId)
 			throws AuthenticationException, InvalidRequestException,
 			APIConnectionException, CardException, APIException {
-		return retrieve(RequestOptions.builder().setApiKey(apiKey).build());
+		if (null == apiKeyOrAccountId || apiKeyOrAccountId.startsWith("sk_")) {
+			return retrieve(RequestOptions.builder().setApiKey(apiKeyOrAccountId).build());
+		} else {
+			return retrieve(apiKeyOrAccountId, (RequestOptions) null);
+		}
 	}
+
 	public static Account retrieve(RequestOptions options)
 			throws AuthenticationException, InvalidRequestException,
 			APIConnectionException, CardException, APIException {
@@ -88,5 +122,64 @@ public class Account extends APIResource {
 			null,
 			Account.class,
 			options);
+	}
+
+	public static Account retrieve(String id, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.GET, instanceURL(Account.class, id), null, Account.class, options);
+	}
+
+	public Account update(Map<String, Object> params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return update(params, (RequestOptions) null);
+	}
+
+	public Account update(Map<String, Object> params, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.POST, instanceURL(Account.class, this.id), params, Account.class, options);
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+
+		Account account = (Account) o;
+		return equals(id, account.id) &&
+			equals(chargesEnabled, account.chargesEnabled) &&
+			equals(detailsSubmitted, account.detailsSubmitted) &&
+			equals(transfersEnabled, account.transfersEnabled) &&
+			equals(currenciesSupported, account.currenciesSupported) &&
+			equals(email, account.email) &&
+			equals(statementDescriptor, account.statementDescriptor) &&
+			equals(defaultCurrency, account.defaultCurrency) &&
+			equals(country, account.country) &&
+			equals(timezone, account.timezone) &&
+			equals(displayName, account.displayName) &&
+			equals(verification, account.verification) &&
+			equals(legalEntity, account.legalEntity);
+	}
+
+	public static class Verification extends StripeObject {
+		List<String> fieldsNeeded;
+		Long dueBy;
+		Boolean contacted;
+
+		public List<String> getFieldsNeeded() {
+			return fieldsNeeded;
+		}
+		public Long getDueBy() {
+			return dueBy;
+		}
+		public Boolean getContacted() {
+			return contacted;
+		}
 	}
 }
--- a/src/main/java/com/stripe/model/LegalEntity.java
+++ b/src/main/java/com/stripe/model/LegalEntity.java
@@ -1,0 +1,176 @@
+package com.stripe.model;
+
+import com.stripe.exception.APIConnectionException;
+import com.stripe.exception.APIException;
+import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.List;
+import java.util.Map;
+
+public class LegalEntity extends StripeObject {
+	String type;
+	Address address;
+	String businessName;
+	DateOfBirth dob;
+	String firstName;
+	String lastName;
+	Address personalAddress;
+	Verification verification;
+	List<Owner> additionalOwners;
+
+	public String getType() {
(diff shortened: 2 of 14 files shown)


danwang:
Added `LegalEntity` resource and all its subresources. These were mostly copied off of the docs. Also added a deserialization test and a couple tests for serialization of nested (hash + list) params.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -235,17 +236,18 @@ private static String createQuery(Map<String, Object> params)
...
 	}
 
+
 	private static Map<String, String> flattenParams(Map<String, Object> params)
 			throws InvalidRequestException {
 		if (params == null) {
 			return new HashMap<String, String>();
 		}
-		Map<String, String> flatParams = new HashMap<String, String>();
+		Map<String, String> flatParams = new LinkedHashMap<String, String>();
```
danwang: This is used to preserve insertion order when iterating.

danwang:
Finished the `Account` methods. The model-specific tests don't actually test much besides which URL/method is used, but all the create/retrieve/update methods are almost exactly the same across all models.

src/main/java/com/stripe/model/LegalEntity.java
```diff
@@ -0,0 +1,164 @@
...
+		}
+		public Integer getMonth() {
+			return month;
+		}
+		public Integer getYear() {
+			return year;
+		}
+	}
+
+	public static class Verification extends APIResource {
```
bkrausz: There are 2 different Verification objects, the top level one on Accounts has fields_needed, due_by, and contacted. The one on LegalEntity and the Owners has status, document, and details.

src/main/java/com/stripe/model/LegalEntity.java
```diff
@@ -0,0 +1,164 @@
...
+	}
+	public Address getPersonalAddress() {
+		return personalAddress;
+	}
+	public String getPersonalIdNumber() {
+		return personalIdNumber;
+	}
+	public String getSsnLast4() {
+		return ssnLast4;
+	}
```
bkrausz: PersonalIdNumber and SSN last 4 are never readable.

bkrausz:
See comments on the fields, defer to @jimdanz on the structure.

src/main/java/com/stripe/model/Account.java
```diff
@@ -67,18 +69,45 @@ public String getDisplayName() {
...
+
 	public static Account retrieve()
 			throws AuthenticationException, InvalidRequestException,
 			APIConnectionException, CardException, APIException {
 		return retrieve((RequestOptions) null);
 	}
 
+	/**
+	 * In order to preserve backwards-compatibility, this method does two things.
+   * If the parameter looks like an API key (starts with sk_), retrieve the
```
jimdanz: whitespace patrol

src/main/java/com/stripe/model/LegalEntity.java
```diff
@@ -0,0 +1,164 @@
...
+		}
+		public String getDocument() {
+			return document;
+		}
+		public String getDetails() {
+			return details;
+		}
+	}
+
+	public static class Owner extends APIResource {
```
jimdanz: types that are just bundles of unpacked json, that don't ahve their own API methods, should extend `StripeObject` rather than `APIResource`

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
jimdanz: :+1:

jimdanz:
lg, exciting tests.  only comment is the `StripeObject` thing

src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -0,0 +1,116 @@
...
+		"    }" +
+		"  }," +
+		"  \"default_currency\": \"usd\"," +
+		"  \"country\": \"US\"," +
+		"  \"object\": \"account\"," +
+		"  \"business_name\": \"Stripe.com\"," +
+		"  \"business_url\": null," +
+		"  \"support_phone\": null," +
+		"  \"managed\": null" +
+		"}";
```
shalecraig: Can you put this blob with the other json files?

Something like `test/resources/com/stripe/model/account.json`.

src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+	@Test
+	public void testDeserialize() throws StripeException {
+		Account acc = APIResource.GSON.fromJson(accountHash, Account.class);
+
+		assertEquals("acct_1032D82eZvKYlo2C", acc.getId());
+		assertEquals(false, acc.getChargesEnabled());
+		assertEquals(false, acc.getDetailsSubmitted());
+		assertEquals(false, acc.getTransfersEnabled());
```
shalecraig: Should use assertFalse (or whatever it's called)

src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -0,0 +1,116 @@
...
+		List<String> currenciesSupported = new LinkedList<String>();
+		currenciesSupported.add("usd");
+		currenciesSupported.add("aud");
+		assertEquals(currenciesSupported, acc.getCurrenciesSupported());
+
+		assertEquals("company", acc.getLegalEntity().getType());
+		assertEquals("San Francisco", acc.getLegalEntity().getAddress().getCity());
+
+		assertEquals("site@stripe.com", acc.getEmail());
+		assertEquals(null, acc.getStatementDescriptor());
```
shalecraig: AssertNull

src/test/java/com/stripe/model/LegalEntityTest.java
```diff
@@ -0,0 +1,69 @@
...
+		assertEquals(null, le.getAddress().getLine2());
+		assertEquals("New York", le.getAddress().getCity());
+		assertEquals("NY", le.getAddress().getState());
+		assertEquals("10014", le.getAddress().getPostalCode());
+		assertEquals("US", le.getAddress().getCountry());
+		assertEquals("Monica", le.getFirstName());
+		assertEquals("Geller", le.getLastName());
+		assertEquals(new Integer(4), le.getDob().getDay());
+		assertEquals(new Integer(4), le.getDob().getMonth());
+		assertEquals(new Integer(1969), le.getDob().getYear());
```
shalecraig: I think autoboxing means that you don't need this, but it's fine.

src/test/java/com/stripe/model/LegalEntityTest.java
```diff
@@ -0,0 +1,69 @@
...
+		"		\"status\": \"verified\"," +
+		"		\"document\": null," +
+		"		\"details\": null" +
+		"	}" +
+		"}";
+
+	@Test
+	public void testDeserialize() throws StripeException {
+		LegalEntity le = APIResource.GSON.fromJson(legalEntityHash, LegalEntity.class);
+
```
shalecraig: Would it be better to just construct an expected version of the object, implement `.equals`, then `assertEquals`?

src/test/java/com/stripe/model/LiveStripeResponseGetterTest.java
```diff
@@ -0,0 +1,62 @@
...
+	@Test
+	public void testCreateQuery() throws StripeException, UnsupportedEncodingException {
+		Map<String, Object> params = new HashMap<String, Object>();
+		params.put("a", "b");
+		assertEquals("a=b", srg.createQuery(params));
+	}
+
+	@Test
+	public void testCreateQueryWithNestedParams() throws StripeException, UnsupportedEncodingException {
+		/* Use TreeMap because it preserves iteration order */
```
shalecraig: Incorrect comment

src/test/java/com/stripe/model/LiveStripeResponseGetterTest.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.*;
+
+public class LiveStripeResponseGetterTest {
+	LiveStripeResponseGetter srg = new LiveStripeResponseGetter();
```
shalecraig: Can you make this part of a `@Before` (or something)? Otherwise, this test is stateful.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -253,6 +255,14 @@ private static String createQuery(Map<String, Object> params)
 							nestedEntry.getValue());
 				}
 				flatParams.putAll(flattenParams(flatNestedMap));
+			} else if (value instanceof List<?>) {
+				Map<String, Object> flatNestedMap = new LinkedHashMap<String, Object>();
+				int index = 0;
+				for (Object element : (List<?>) value) {
```
shalecraig: This may be cleaner. It also doesn't allocate a new object for each iteration of the loop:

```java
Iterator<String> myIter = value.iterator();
for (int index=0; myIter.hasNext(); ++index) {
 flatNestedMap.put(String.format("%s[%s], key, index), myIter.next());
}
```

Both are correct, so feel free to #ignore if you don't want to.

src/main/java/com/stripe/model/LegalEntity.java
danwang: Fixed. Created `Account.Verification` and `LegalEntity.Verification`.

src/main/java/com/stripe/model/LegalEntity.java
danwang: Removed `personal_id_number`, `ssn_last_4`, `business_tax_id`, and `business_vat_id`.

src/main/java/com/stripe/model/LegalEntity.java
danwang: Fixed to extend `StripeObject`.

src/test/java/com/stripe/model/AccountTest.java
danwang: Done. Also moved `resource` from `DeserializerTest` into `BaseStripeTest` and made `DeserializerTest` extend `BaseStripeTest`.

src/test/java/com/stripe/model/LegalEntityTest.java
danwang: Done. Implemented `.equals` on all new objects.

src/test/java/com/stripe/model/LiveStripeResponseGetterTest.java
danwang: s/TreeMap/LinkedHashMap

src/test/java/com/stripe/model/LiveStripeResponseGetterTest.java
danwang: Added `srg` as an instance variable and initialize it in `@Before`.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
danwang: Changed to use an iterator.

How do you mean by the allocation of a new object? I thought `for (x : y)` in Java expanded into a loop using the iterator methods.

src/test/java/com/stripe/model/LegalEntityTest.java
danwang: I get the error

```
 reference to assertEquals is ambiguous, both method assertEquals(double,double) in org.junit.Assert and method assertEquals(java.lang.Object,java.lang.Object) in org.junit.Assert match
```

Any tips for getting rid of this? (moot point, as this is removed and replaced with one `Account#equals`)

danwang:
The CI failures are the same (flaky) tests I've seen on master. Still not sure why they're failing, but I'd rather move `StripeTest` over to new-style tests in a separate PR.

bkrausz:
sgtm, if you're happy with this then I am. :sheep:

PR #156, opened 2015-03-25
Colorized text in the readme

shalecraig:
cc @jimdanz


shalecraig:
Since the diff view doesn't show it colorized, you can see it [here](https://github.com/shalecraig/stripe-java/tree/patch-1#maven-users)

jimdanz:
:+1:

PR #162, opened 2015-04-23
Null check refunds

### LINKED ISSUE #161: Null check `refunds` in `getRefunds` on charges.
https://github.com/stripe/stripe-java/blob/54f695e42ea09bd681fce25fe5e1b9d7e9a026b0/src/main/java/com/stripe/model/Charge.java#L219

If `refunds` has not been set (or has not been deserialized), this will throw an NPE.

andrewpthorp:
This PR fixes #161 

/cc @berg @jimdanz


jimdanz:
We've fixed this on the API side, so that it should no longer be an issue (it should now be impossible to get a charge object with no refunds field.)

jimdanz:
I just realized that even with our fix, old event data will not get fixed up.  Merging this PR!

PR #166, opened 2015-05-01
Add concrete class for non-implemented payment sources to fallback on.

raycmorgan:
Also implement `alipay_account` payment source.

r? @jimdanz 
cc @maxwells

### DIFF
--- a/src/main/java/com/stripe/model/ConcretePaymentSource.java
+++ b/src/main/java/com/stripe/model/ConcretePaymentSource.java
@@ -1,0 +1,109 @@
+package com.stripe.model;
+
+import com.stripe.Stripe;
+import com.stripe.exception.APIConnectionException;
+import com.stripe.exception.APIException;
+import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * ConcretePaymentSource is a PaymentSource that is used as a catch-all for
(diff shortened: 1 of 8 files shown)


src/main/java/com/stripe/model/ConcretePaymentSource.java
```diff
@@ -0,0 +1,99 @@
...
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ConcretePaymentSource extends APIResource implements PaymentSource {
```
jimdanz: How about a comment here describing what this is / why it is?

src/test/java/com/stripe/StripeTest.java
```diff
@@ -7,50 +7,8 @@
...
-import com.stripe.model.Plan;
-import com.stripe.model.Recipient;
-import com.stripe.model.Refund;
-import com.stripe.model.Reversal;
-import com.stripe.model.ShippingDetails;
-import com.stripe.model.Subscription;
-import com.stripe.model.Token;
-import com.stripe.model.Transfer;
-import com.stripe.model.TransferReversalCollection;
+import com.stripe.model.*;
```
jimdanz: I've historically advocated against wildcard imports (https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.3.1-wildcard-imports) but perhaps it's OK in tests?  OOC did your IDE do this automatically or such?

raycmorgan: Ah, yeah that was my IDE. I can re-expand them.

jimdanz: IDE love that

PR #167, opened 2015-05-05
Missing field in Card object

ngutman:
Missing dynamicLast4 in Card object (Apple Pay integration)


raycmorgan:
Thanks for adding!

PR #171, opened 2015-05-14
add support for card metadata

dlambeth:


kyleconroy:
@dlambeth Thanks! This looks great.

Issue #176, opened 2015-05-29
NPE on "account.reject" event


agentcoops:
If a managed account under the control of a Connect user is currently flagged as fraudulent, the user in question receives a webhook event of type "account.reject" with the message "Fraudulent account. Please close and refund charges." However, the JSON sent in this case is currently malformed and missing the data.object.object field as specified in the documentation: https://stripe.com/docs/api#event_object. Because of this the Java client fails during deserialization of the JSON, throwing:
```
java.lang.NullPointerException 
at com.stripe.model.EventDataDeserializer.deserialize(EventDataDeserializer.java:105) 
at com.stripe.model.EventDataDeserializer.deserialize(EventDataDeserializer.java:17) 
at com.google.gson.TreeTypeAdapter.read(TreeTypeAdapter.java:58) 
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:93) 
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:172) 
at com.google.gson.Gson.fromJson(Gson.java:803) 
at com.google.gson.Gson.fromJson(Gson.java:768) 
at com.google.gson.Gson.fromJson(Gson.java:717) 
at com.google.gson.Gson.fromJson(Gson.java:689)
```

bkrausz:
See comment on pull request: https://github.com/stripe/stripe-java/pull/177#issuecomment-111293351

PR #177, opened 2015-05-29
Quick fix for account.reject event NPE

agentcoops:
This is a less than ideal solution for issue #176. If the "object" field is missing but "type" is present, the information necessary for deserialization can be extracted from the latter. This will work as a temporary fix for anyone currently suffering from this issue.


bkrausz:
Given that we no longer send account.rejects for normal accounts (this was a private webhook we accidentally starting using more broadly), I'm going to close out. Our contract is that every object will have an "object" property, any case where we don't do so is an API bug.

PR #178, opened 2015-05-30
External accounts

bkrausz:
r? @jimdanz
CC @maxwells @shalecraig @russelldavis 

No more ConceretePaymentMethod, I made ExternalAccount an abstract. The one downside is the need to make ExternalAccountTypeAdapterFactory. Not /too/ crazy complicated, but a bit hard to add tests for.


src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -38,26 +38,30 @@ public void testDeserialize() throws StripeException, IOException {
...
+        assertEquals(false, acc.getChargesEnabled());
+        assertEquals(false, acc.getDetailsSubmitted());
+        assertEquals(false, acc.getTransfersEnabled());
+
+		LinkedList<String> cs = new LinkedList<String>();
+		cs.add("usd");
+		cs.add("aud");
+        assertEquals(cs, acc.getCurrenciesSupported());
+
+        LegalEntity le = new LegalEntity();
```
jimdanz: looks like these functions mix tabs and 4-space indentation

src/main/java/com/stripe/model/MetadataStore.java
```diff
@@ -14,14 +14,10 @@
  */
 public interface MetadataStore<T> {
 	Map<String, String> getMetadata();
-	void setMetadata(Map<String, String> metadata);
 
 	MetadataStore<T> update(Map<String, Object> params) throws AuthenticationException, InvalidRequestException,
 		APIConnectionException, CardException, APIException;
 
-	@Deprecated
-	MetadataStore<T> update(Map<String, Object> params, String apiKey) throws AuthenticationException, InvalidRequestException,
```
jimdanz: would rather not nuke this method yet (save it for a major release) unless it's really causing you trouble

src/main/java/com/stripe/model/Account.java
```diff
@@ -196,40 +202,6 @@ public Account update(Map<String, Object> params, RequestOptions options)
 		return request(RequestMethod.POST, instanceURL(Account.class, this.id), params, Account.class, options);
 	}
 
-	@Override
-	public boolean equals(Object o) {
```
jimdanz: -1 to this.. our POJOs should have equals methods

jimdanz: nevrermind: didn't realize that we tend not to have equals methods.  Sounds like that's for fixing another day.

src/main/java/com/stripe/model/BankAccount.java
```diff
@@ -12,13 +19,25 @@
...
+            throws AuthenticationException, InvalidRequestException, APIConnectionException, CardException, APIException {
+        return update(params, (RequestOptions) null);
+    }
+
+    public BankAccount update(Map<String, Object> params, RequestOptions options)
+            throws AuthenticationException, InvalidRequestException, APIConnectionException, CardException, APIException {
+        return request(RequestMethod.POST, this.getInstanceURL(), params, BankAccount.class, options);
+    }
+
+    public DeletedBankAccount delete()
```
jimdanz: had to go back to java 101 and google for this http://stackoverflow.com/questions/14519797/overriding-a-method-with-different-return-types-in-java to remember that it works, but bullish bullish :)

src/test/java/com/stripe/StripeTest.java
```diff
@@ -2072,67 +2070,7 @@ public void testAlipayAccountDeleting() throws StripeException {
 		AlipayAccount alipayAccount = (AlipayAccount) cus.getSources().getData().get(0);
 		alipayAccount.delete();
 
-		PaymentSourceCollection sources = cus.getSources().all(new HashMap<String, Object>());
+		ExternalAccountCollection sources = cus.getSources().all(new HashMap<String, Object>());
 		assertEquals(0, sources.getData().size());
 	}
-
-	@Test
-	public void testAlipayAccountAsUnknownSourceCreation() throws StripeException {
```
jimdanz: It really would be great to have a test of gracefully handling an unknown external account type.
If there's no good way to hack a mapping anymore (I see why it wouldn't be straightforward to do that), maybe it's better to switch these tests over to using network stubbing?

jimdanz:
I'm very bullish on this approach in general, particularly the choice to have `ExternalAccount` be a common superclass that can have either a customer or an account.

Main concrete ASK is to resurrect a test of "able to deserialize an object that I haven't heard of."

@maxwells you're pretty close to a lot of the code that's churned here.  Would you like to take a look through pre-merge?

src/main/java/com/stripe/model/MetadataStore.java
bkrausz: Do we want to keep this consistent across classes? I didn't remove any uses of this, but I didn't add it to new classes, which is why I removed it from the interface.

src/test/java/com/stripe/model/AccountTest.java
bkrausz: Since I changed enough of the file I'll convert it to 4 spaces. Still need to do the giant conversion off of tabs at some point.

bkrausz:
> Main concrete ASK is to resurrect a test of "able to deserialize an object that I haven't heard of."

Yea, I tested manually by commenting out the alipay_account parsing, but agree that it's fragile enough that I'll figure out a way to add a test.

maxwells:
seems totally sane to me!

src/main/java/com/stripe/model/MetadataStore.java
jimdanz: ok, seems fine

jimdanz:
:ship:

PR #184, opened 2015-06-24
Add license information to the pom.xml

reiz:
Adding the MIT license to the pom.xml. That way the license is available via the package manager.


jimdanz:
Thanks for the PR!

PR #186, opened 2015-07-02
Add request IDs to StripeExceptions.

cupcait:


cupcait:
r? @kyleconroy 
cc @jimdanz

src/main/java/com/stripe/exception/StripeException.java
```diff
@@ -2,14 +2,32 @@
...
 
-	public StripeException(String message, Throwable e) {
+	public StripeException(String message, String requestId, Throwable e) {
 		super(message, e);
+		this.requestId = requestId;
 	}
 
 	private static final long serialVersionUID = 1L;
 
+    public String getRequestId() {
```
kyleconroy: Looks like the spacing got messed up here.

src/main/java/com/stripe/exception/StripeException.java
```diff
@@ -2,14 +2,32 @@
...
+		return requestId;
+    }
+
+    public String toString() {
+		String reqIdStr = "";
+		if (requestId != null) {
+			reqIdStr = "; request-id: " + requestId;
+        }
+
```
kyleconroy: nit: no need for this empty line

kyleconroy:
We should add a test that checks the request ID can be accessed from an error. `testChargeRetrieveNullId` seems like the best place to put this.

kyleconroy:
@jimdanz any opinion on adding http response headers to exceptions? Since we don't add the body or the status code, we decided just adding the request ID would be better.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -502,20 +510,20 @@ private static StripeResponse getMultipartStripeResponse(
...
 		case 404:
-			throw new InvalidRequestException(error.message, error.param, null);
+			throw new InvalidRequestException(error.message, requestId, error.param, null);
 		case 401:
 			throw new AuthenticationException(error.message);
 		case 402:
-			throw new CardException(error.message, error.code, error.param, error.decline_code, error.charge, null);
+			throw new CardException(error.message, requestId, error.code, error.param, error.decline_code, error.charge, null);
 		default:
 			throw new APIException(error.message, null);
```
jimdanz: Do we never return a request ID on 500s?  We definitely record the iar in some cases.

jimdanz:
@kyleconroy I'm +1 to just the request id!

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
kyleconroy: Yep, the requestId should be added in all these cases.

cupcait:
@kyleconroy @jimdanz updated! Can one of you take another look? Thanks!

src/test/java/com/stripe/StripeTest.java
```diff
@@ -344,7 +344,9 @@ public void testChargeRetrieveNullId() throws StripeException {
 			assertTrue(false);
 		}
 		catch (InvalidRequestException e) {
-			// Expected
+			String requestId = e.getRequestId();
+			assertTrue(requestId != null);
```
kyleconroy: Can we check that it also isn't an empty string?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -363,8 +364,16 @@ private static StripeResponse makeURLConnectionRequest(
 			}
 			int rCode = response.responseCode;
 			String rBody = response.responseBody;
+
+			String requestId = "";
+			Map<String, List<String>> headers = response.getResponseHeaders();
```
kyleconroy: I believe that headers can be null here.

kyleconroy:
Other than my two notes, looks good.

kyleconroy:
:+1:

PR #191, opened 2015-07-06
Adds http status code to InvalidRequestException

nikyoudale:
There is currently no way in the Java api bindings to tell the difference between a `404` and a `400` response. These are both identically mapped to `InvalidRequestException`.

Example use case:
Getting the upcoming invoice for a customer. There is a meaningful difference between a 'not found' vs a 'bad request' response.


kyleconroy:
Tracking here #222

PR #193, opened 2015-07-15
Add retrieve, update, close, and all methods to Dispute.

cupcait:
Also deprecate old `/v1/charges/:charge/dispute` endpoints.

r? @kyleconroy cc @jimdanz

### DIFF
--- a/src/test/java/com/stripe/StripeTest.java
+++ b/src/test/java/com/stripe/StripeTest.java
@@ -18,7 +18,6 @@ import com.stripe.model.BitcoinReceiver;
 import com.stripe.model.BitcoinTransaction;
 import com.stripe.model.Card;
 import com.stripe.model.Charge;
-import com.stripe.model.ChargeCollection;
 import com.stripe.model.ChargeRefundCollection;
 import com.stripe.model.Coupon;
 import com.stripe.model.Customer;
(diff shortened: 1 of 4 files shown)


src/test/java/com/stripe/StripeTest.java
```diff
@@ -536,6 +535,80 @@ public void testDisputedCharge() throws StripeException, InterruptedException {
...
+		Dispute dispute = disputedCharge.getDispute();
+		Dispute retrievedDispute = Dispute.retrieve(dispute.getId());
+		assertEquals(dispute.getId(), retrievedDispute.getId());
+	}
+
+	@Test
+	public void testDisputeList() throws StripeException, InterruptedException {
+		createDisputedCharge(100).getDispute();
+		createDisputedCharge(150).getDispute();
+		createDisputedCharge(300).getDispute();
```
cupcait: I'm not sure that these three charge creations are necessary - there seem to be existing disputes on this account already. For now I have these here just to be extra safe, but let me know if it should be peaceful to remove them.

cupcait:
Should I also add comments to deprecate `/charges/:charge/dispute` methods in this change?

kyleconroy:
@cupcait In this case, we can actually use the `@Deprecated` annotation.

src/test/java/com/stripe/StripeTest.java
kyleconroy: I think it's safe to remove these

kyleconroy:
:+1:

PR #196, opened 2015-07-20
In one test file, convert indentation to tabs

shalecraig:
99% of this file uses tabs, why not make it 100%.

R? @kyleconroy


kyleconroy:
:+1: once tests pass

PR #197, opened 2015-07-20
Deprecate transfer recipient, use destination instead

shalecraig:
Deprecate the `recipient` field in the `Transfer` class in favour of using the `destination` field instead.

R? @kyleconroy

### DIFF
--- a/src/main/java/com/stripe/model/Transfer.java
+++ b/src/main/java/com/stripe/model/Transfer.java
@@ -26,6 +26,7 @@ public class Transfer extends APIResource implements MetadataStore<Transfer> {
 	List<String> otherTransfers;
 	@Deprecated
 	String recipient;
+	String destination;
 	BankAccount account;
 	String balanceTransaction;
 	Map<String, String> metadata;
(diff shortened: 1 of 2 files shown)


kyleconroy:
Why did you need to remove the reversals test case?

shalecraig:
The test created a [recipient](https://github.com/stripe/stripe-java/pull/197/files#diff-be338c286171d2395d658e3e2168d399L121) with the [000123456789 (Transfer will succeed)](https://stripe.com/docs/testing) testing account number. Then, it tried to [create a reversal](https://github.com/stripe/stripe-java/pull/197/files#diff-be338c286171d2395d658e3e2168d399L1266) on the transfer.

In my change, I changed the default transfer params to have an account destination. The transfer to the stripe account succeeds immediately, but still may be reversed.

I've removed the test because it was essentially asserting on the stripe API behaviour, and only vaguely that the bindings worked correctly - if I do X then Y, Stripe throws an error.

A better test would be one where we create the transfer reversal, then assert that fields are filled out completed successfully. If you want, I can implement that.

kyleconroy:
No need to write an additional test, just wanted some context on why it was removed. Squash if you want, 4 commits (with good names) is also fine.

kyleconroy:
:+1:

PR #198, opened 2015-07-23
Add account to application fees

andrewpthorp:
per https://stripe.com/docs/upgrades#2013-12-03

/cc @shalecraig


src/test/java/com/stripe/model/DeserializerTest.java
```diff
@@ -105,6 +105,14 @@ public void deserializeAppFeeRefundList() throws IOException {
 	}
 
 	@Test
+	public void deserializeAppFee() throws IOException {
+		String json = resource("appfee_refund_list.json");
```
jimdanz: I can grab this on a followup PR but I'd love to rename this to "appfee_with_refund_list.json" because my first thought was "what?  a list of Application Fee Refunds?" when really it's one applicaiton fee object, that has a list of refunds.

andrewpthorp: Yup, I originally added my own fixture because I didn't realize what this
one was. Will fix.

jimdanz: :100:

Issue #199, opened 2015-07-27
LiveStripeResponseGetter::handleAPIError instantiates InvalidRequestException with arguments in the wrong order


EvanDooner:
Currently, when an `InvalidRequestException` is created in the `LiveStripeResponseGetter::handleAPIError(String, int, String)` method, the arguments are provided in the order: message, requestId, param, null (cause). However, the constructor for `InvalidRequestException` has the following order for its parameters: message, param, requestId, cause.

This is causing created `InvalidRequestException` instances to return the requestId when the getParam() method is called.

This bug was introduced in commit 105c664add8aa5457a5d04cd14451d5d69517cd9

kyleconroy:
This was fixed in July, forgot to close this issue.

PR #200, opened 2015-07-27
Correct order of constructor arguments in LiveStripeResponseGetter

EvanDooner:
The `handleAPIError(String, int, String)` method in the `LiveStripeResponseGetter` class provides the arguments to the `InvalidRequestException` in the wrong order.

Correct the order of the arguments to match the declared parameters of the `InvalidRequestException` constructor.

This fixes issue #199


kyleconroy:
Thanks for fixing. :D

PR #203, opened 2015-08-06
fixes #202. Add status setter to Charge

### LINKED ISSUE #202: No setter method for status attribute in Charge.java
Pretty much all the other fields in Charge.java have both getters and setters except for status, which only has a getter. 

Let's add a setStatus method to Charge, which would make it more consistent as well as useful for testing.

dlin303:
Pretty much all other attributes in Charge.java have both getters and setters except for status. Adding a setter to status will provide better consistency.


jrodbeta:
+1 for easy test mock object creation.

jimdanz:
:+1: thanks!

Issue #190, opened 2015-07-05
Charge does not expose "Transfer ID"


alexg313:
When a platform creates a charge on behalf of a connected account and provides the account ID of the connected account in the "destination" parameter, the full amount of the charge (minus the application fee) is immediately transferred to the connected account. 

When I login to my account (the platform), I can see the relevant Transfer ID in the Payment Details section of the payment. As far as I can tell, there is no way to programmatically retrieve this transfer Id.

hardtoe:
Pull request #207 resolves this issue if it is accepted.

kyleconroy:
PR has been merged

PR #207, opened 2015-08-25
added 'transfer' field to Charge object

hardtoe:
fixes issue #190 Charge does not expose "Transfer ID"


hardtoe:
will this be accepted or rejected?

jimdanz:
Sorry for the delay -- this looks good.  Thanks for the PR!

PR #125, opened 2014-12-11
Make StripeCollection implement java.util.Iterable

michaelklishin:
As discussed in #101.

Fixes #101.

PR #208, opened 2015-08-25
added support for iterating over stripe collections lazily-loading

hardtoe:
lazy loading iteration over StripeCollections per latest conversation in pull request #125 "Make StripeCollection implement java.util.Iterable"

This required a couple of interfaces to be added to normalize stripe objects and collections across the api:
1. HasId interface with getId() method was added to stripe objects that can be iterated over.
2. StripeCollectionInterface was added to normalize StripeCollection and StripeCollectionAPIResource classes which previously did not share a common base class.  StripeCollectionAPIResource unfortunately contains getURL() and getUrl() now.  I suggest that one of them should be deprecated in favor of the other.

### DIFF
--- a/src/main/java/com/stripe/model/StripeCollection.java
+++ b/src/main/java/com/stripe/model/StripeCollection.java
@@ -1,8 +1,12 @@
 package com.stripe.model;
 
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
-public abstract class StripeCollection<T> extends StripeObject {
+import com.stripe.net.RequestOptions;
+
+public abstract class StripeCollection<T extends HasId> extends StripeObject implements StripeCollectionInterface<T>, Iterable<T> {
 	List<T> data;
 	Integer totalCount;
 	Boolean hasMore;
@@ -41,5 +45,9 @@ public abstract class StripeCollection<T> extends StripeObject {
 	/** 3/2014: Legacy (from before newstyle pagination API) */
 	public void setCount(Integer count) {
 		this.count = count;
+	}
+	
+	public Iterator<T> iterator() {
+		return new StripeCollectionIterator<T>(this);
 	}
 }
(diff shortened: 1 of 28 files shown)


hardtoe:
looking for some feedback on this

src/main/java/com/stripe/model/StripeCollection.java
```diff
@@ -1,8 +1,12 @@
 package com.stripe.model;
 
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
-public abstract class StripeCollection<T> extends StripeObject {
+import com.stripe.net.RequestOptions;
```
kyleconroy: nit: Useless import.

src/main/java/com/stripe/model/StripeCollectionIterator.java
```diff
@@ -0,0 +1,80 @@
...
+	public boolean hasNext() {
+		return 
+			currentDataIterator.hasNext() ||
+			currentCollection.getHasMore();
+	}
+
+	@Override
+	public T next() {
+		if (currentDataIterator.hasNext()) {
+			final T next =
```
kyleconroy: no need for this to be `final`

src/main/java/com/stripe/model/StripeCollectionIterator.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	public void remove() {
+		throw new UnsupportedOperationException();
+	}
+
+	@SuppressWarnings("unchecked")
+	protected StripeCollectionInterface<T> list(
+		final Map<String, Object> params, 
+		final RequestOptions options
+	) throws Exception {
```
kyleconroy: This should throw a `StripeException`

kyleconroy:
@hardtoe Thanks for the PR! 

Overall, this looks great. We'd like there to be more tests, specifically ones with mocked out network requests simulating different paging scenarios. Once that's all finished, could you then squash your commits? 

If you don't have time to do all this, we can go ahead and write the tests. We'll just need to you fix the comments and squash.

brandur:
Closing this out in favor of #228 where we'll continue work and in a little more testing. Thanks for getting this started @hardtoe!

Issue #212, opened 2015-08-27
Tokens should be able to display linked bank accounts


matthewarkin:
The token class only has a card property it should also have a bank account property since a token can be either a card or bank account token.

mmcgrana:
Closed by #213. Thanks for flagging @matthewarkin!

PR #213, opened 2015-08-27
Add BankAccount to Token

mmcgrana:
cc @kyleconroy 

ref #212


mmcgrana:
@kyleconroy gave the :+1: IRL.

PR #214, opened 2015-08-30
Manually enable TLSv1.2 for the benefit of Java 6 and 7

ebroder:
Java 6 and 7 both support TLSv1.2, but don't enable it (or TLSv1.1) by default for TLS connections due to concerns at release-time around server compatibility. At this point, those are largely addressed, and TLSv1.0 is being deprecated by, among others, the PCI DSS within the next year.

We don't want to pin specifically to TLSv1.2, as that makes us not forward-compatible when TLSv1.3 (and later) is eventually released. However, the only way I could find to update the list of protocol versions was to call the `setEnabledProtocols` method on an `SSLSocket`.

And the only way I could find to do that was to provide my own `SSLSocketFactory` to `HttpsURLConnection`, which this PR does. The new factory accepts an "underlying" factory as an argument, and proxies all calls to it, enabling TLSv1.2 on the resulting sockets as appropriate.

r? @jimdanz
cc @kyleconroy @shalecraig  

I'm open to other ideas on how to make this happen (I considered messing around with the `https.protocols` property, but this seemed cleaner). I'm also not sure if it's a good idea to grab the `HttpsURLConnection` default `SSLSocketFactory` in the static initializer, but it seemed to work in my testing (it might not work on, e.g., AppEngine? I don't know how to test that).

I'm _also_ unclear why the tests are failing on openjdk6, but we should probably figure it out. (It seems to be coming from deep in the guts of the TLS handshake, which is more than a little terrifying)

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -7,6 +7,8 @@ import com.stripe.exception.AuthenticationException;
 import com.stripe.exception.CardException;
 import com.stripe.exception.InvalidRequestException;
 
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLSocketFactory;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -37,6 +39,8 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
 	 * environments.
 	 */
 	private static final String CUSTOM_URL_STREAM_HANDLER_PROPERTY_NAME = "com.stripe.net.customURLStreamHandler";
+
+	private static final SSLSocketFactory socketFactory = new StripeSSLSocketFactory(HttpsURLConnection.getDefaultSSLSocketFactory());
 
 	public <T> T request(
 			APIResource.RequestMethod method,
@@ -136,6 +140,9 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
 		conn.setUseCaches(false);
 		for (Map.Entry<String, String> header : getHeaders(options).entrySet()) {
 			conn.setRequestProperty(header.getKey(), header.getValue());
+		}
+		if (conn instanceof HttpsURLConnection) {
+			((HttpsURLConnection) conn).setSSLSocketFactory(socketFactory);
 		}
 
 		return conn;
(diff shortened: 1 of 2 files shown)


ebroder:
(For the record, I can't reproduce the java6 error locally, but my java6 is "1.6.0_36" instead of Travis's "1.6.0_34")

jimdanz:
Thanks for the full explanation here.  I"m OK with this if we can get the openjdk6 build failure sorted.

@kyleconroy do you want to review?

kyleconroy:
@jimdanz ACK, will figure out what's going on

ebroder:
Ok, I think I managed to get the openjdk6 tests passing. Some quick notes on what I learned:

The version of openjdk-6 that Travis is using is a security update to Ubuntu 12.04, but not the latest available update - it's about 4 or 5 updates behind what's current. You can fetch the packages for that specific version from Launchpad ([i386 and arch-independent](https://launchpad.net/~ubuntu-security-proposed/+archive/ubuntu/ppa/+build/6752557) and [amd64](https://launchpad.net/~ubuntu-security-proposed/+archive/ubuntu/ppa/+build/6752554)).

As it turns out, that build of openjdk doesn't seem to actually support TLSv1.2 at all (womp womp):

```scala
scala> System.getProperty("java.version")
res4: java.lang.String = 1.6.0_34

scala> import javax.net.ssl._
import javax.net.ssl._

scala> SSLContext.getDefault.getSupportedSSLParameters.getProtocols
res1: Array[java.lang.String] = Array(SSLv2Hello, SSLv3, TLSv1, TLSv1.1)
```

I'm a little confused why I was able to enable TLSv1.2 at all, but didn't feel like digging around too much. I added feature detection for which versions of TLS were supported, and that seems to fix the issue (though we, obviously, only negotiate up to TLSv1.1).

I also tested with the version of openjdk-6 that was released with 12.04 ([6b24-1.11.1-4ubuntu2](https://launchpad.net/ubuntu/+source/openjdk-6/6b24-1.11.1-4ubuntu2)). That version doesn't support anything newer than TLSv1.0. Hopefully that won't come up often, but I went ahead and did the same feature detection for TLSv1.1 just in case.

I'm still not sure whether this will work at all on AppEngine or Android (or any other platform that's not openjdk/oraclejdk-derived).

kyleconroy:
@ebroder thanks for fixing the openjdk tests. I'll deploy this branch to app engine and see if it works. Since we need to use URL fetch, I have no idea what we need to do.

ebroder:
@kyleconroy My suspicion (hope?) is that AppEngine won't be using `javax.net.ssl.HttpsURLConnection` but some other handler, in which case we should be fine if we just delay instantiating `StripeSSLSocketFactory` until the first time we need it.

kyleconroy:
I've confirmed that this works on App Engine (see http://localhost:8080/test). The canonical-api-line has `tls_version="TLSv1.2"`

jimdanz:
\o/

PR #217, opened 2015-09-17
Add `destination` to charge object

### LINKED ISSUE #187: Charge does not expose "destination" connected account
When a platform creates a charge on a connected account, the returned Charge should the "destination" parameter, which is the ID of the connected account the charge was created on/for. 

See docs here: https://stripe.com/docs/api#capture_charge

kyleconroy:
Fix #187

r? @cupcait


cupcait:
:+1: pending failing tests - it seems like they might not be related to this change in particular, but can you dig?

PR #223, opened 2015-09-18
Add HTTP status code to all exceptions.

### LINKED ISSUE #222: Adds http status code to InvalidRequestException
There is currently no way in the Java api bindings to tell the difference between a 404 and a 400 response. These are both identically mapped to InvalidRequestException.

Example use case:
Getting the upcoming invoice for a customer. There is a meaningful difference between a 'not found' vs a 'bad request' response.

kyleconroy:
If there wasn't an HTTP request, the status code will be 0.

Fix #222

### DIFF
--- a/src/main/java/com/stripe/exception/APIConnectionException.java
+++ b/src/main/java/com/stripe/exception/APIConnectionException.java
@@ -5,11 +5,11 @@ public class APIConnectionException extends StripeException {
 	private static final long serialVersionUID = 1L;
 
 	public APIConnectionException(String message) {
-		super(message, null);
+		super(message, null, 0);
 	}
 
 	public APIConnectionException(String message, Throwable e) {
-		super(message, null, e);
+		super(message, null, 0, e);
 	}
 
 }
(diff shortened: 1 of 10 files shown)


cupcait:
:+1:

jimdanz:
Can I register my vote for fixing the 404 vs 400 issue explicitly (ie, adding a `NotFoundException`) instead of this?  I'm a fan of mapping HTTP status codes to specific exception classes (but admittedly I'd missed the 404 vs 400 issue).

kyleconroy:
@jimdanz I'm fine with that solution, but I don't think there is a good reason to hide the HTTP status code from end users.

jimdanz:
I guess the only reason is potentially confusion.  "APIException is always a 500, so what's getStatusCode()?" but fine with it if you think peaceful.

kyleconroy:
It wouldn't always be a `500`. It's possible that they ran into an upstream gateway timeout and got back a 504. I do think it would be helpful to have a `NotFoundException`, but that would be a breaking change so we'd have to bump to 2.0.

PR #224, opened 2015-10-02
Adding Account.all(accountParams)

rgarbi:
The stripe API docs show a java method .all on the Account object. (here: https://stripe.com/docs/api?lang=java#list_accounts) I could not find it so I added it. I added a test getAllExternalAccounts() but it currently fails because the test account for this repo is not a platform. The specific error is ```You can only create new accounts if you've registered your platform, which you can do at https://dashboard.stripe.com/account/applications/settings``` The test works when I use my account.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -150,6 +150,18 @@ public class Account extends APIResource implements MetadataStore<Account> {
 		return request(RequestMethod.POST, classURL(Account.class), params, Account.class, options);
 	}
 
+	public static AccountCollection all(Map<String, Object> params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return all(params, (RequestOptions) null);
+	}
+
+	public static AccountCollection all(Map<String, Object> params, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.GET, classURL(Account.class), params, AccountCollection.class, options);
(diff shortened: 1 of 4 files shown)


kyleconroy:
Thanks for the PR! We'll need to add another `all` method that allows a user to specify `RequestOptions`.

```java
public static AccountCollection all(Map<String, Object> params)
	return all(params, (RequestOptions) null);
}

public static AccountCollection all(Map<String, Object> params, RequestOptions options)
	return request(RequestMethod.GET, classURL(Account.class), params, AccountCollection.class, options);
}
```

Could you also squash this into two commits? One for the Account changes, and one for the Long change? Thanks!

rgarbi:
You are welcome for the PR, happy to help out! The commits have been squashed and the additional method has been added.

src/main/java/com/stripe/model/Account.java
```diff
@@ -150,6 +150,18 @@ public static Account create(Map<String, Object> params, RequestOptions options)
 		return request(RequestMethod.POST, classURL(Account.class), params, Account.class, options);
 	}
 
+	public static AccountCollection all(Map<String, Object> params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.GET, classURL(Account.class), params, AccountCollection.class, (RequestOptions) null);
```
kyleconroy: This should call all

```java
return all(params, (RequestOptions) null);
```

rgarbi:
Ahh yes, good point. Fixed.

kyleconroy:
@rgarbi thanks!

PR #225, opened 2015-10-06
Add member variables for "client ip" and "email" in Token model

ayushsangani:
As specified in the API docs https://stripe.com/docs/api/java#retrieve_token while retrieving `Token` it should return `client_ip` which is missing from the Token model.

Also `curl request` and `Ruby API` returns `email` field in the response, it would be really helpful if Token has email field


kyleconroy:
Thanks @ayushsangani!

PR #226, opened 2015-10-07
Corrected typo in setTransfer declaration

olivierbellone:
Typo in Reversal.setTransfer() declaration prevented the method from working.


mmcgrana:
Good catch thanks @olivierbellone!

PR #228, opened 2015-10-13
Pagination

brandur:
Adds support for iteration over Stripe collections with lazy pagination that's activated as necessary.

Based on previous work by @hardtoe in #208. (Thanks for that!)

This pull is still a WIP. I'm going to take one more look at the interface and implement a basic test suite so that we have some better guarantees that everything works as we might expect.

### DIFF
--- a/src/main/java/com/stripe/model/StripeCollectionAPIResource.java
+++ b/src/main/java/com/stripe/model/StripeCollectionAPIResource.java
@@ -2,9 +2,10 @@ package com.stripe.model;
 
 import com.stripe.net.APIResource;
 
+import java.util.Iterator;
 import java.util.List;
 
-public abstract class StripeCollectionAPIResource<T> extends APIResource {
+public abstract class StripeCollectionAPIResource<T extends HasId> extends APIResource implements StripeCollectionInterface<T>, Iterable<T>  {
 	List<T> data;
 	Integer totalCount;
 	Boolean hasMore;
@@ -33,6 +34,9 @@ public abstract class StripeCollectionAPIResource<T> extends APIResource {
 	public String getURL() {
 		return url;
 	}
+	public String getUrl() {
+		return url;
+	}
 	public void setURL(String url) {
 		this.url = url;
 	}
@@ -43,5 +47,9 @@ public abstract class StripeCollectionAPIResource<T> extends APIResource {
 	/** 3/2014: Legacy (from before newstyle pagination API) */
 	public void setCount(Integer count) {
 		this.count = count;
+	}
+	
+	public Iterator<T> iterator() {
+		return new StripeCollectionIterator<T>(this);
 	}
 }
--- a/src/main/java/com/stripe/model/StripeCollectionAPIResource.java
+++ b/src/main/java/com/stripe/model/StripeCollectionAPIResource.java
@@ -1,14 +1,18 @@
 package com.stripe.model;
 
 import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
-public abstract class StripeCollectionAPIResource<T extends HasId> extends APIResource implements StripeCollectionInterface<T>, Iterable<T>  {
+public abstract class StripeCollectionAPIResource<T extends HasId> extends APIResource implements StripeCollectionInterface<T> {
 	List<T> data;
 	Integer totalCount;
 	Boolean hasMore;
+    private RequestOptions requestOptions;
+    private Map<String, Object> requestParams;
 	String url;
 	/** 3/2014: Legacy (from before newstyle pagination API) */
 	Integer count;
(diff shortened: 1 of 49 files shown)


brandur:
@kyleconroy @russelldavis I'm still getting some tests in, but the proposed interface here will look roughly like this:

``` java
foreach (Invoice invoice : Invoice.list(...).autoPagingIterable()) {
    ...
}
```

Or for a subcollection:

foreach (InvoiceItem item : invoice.getLines().autoPagingIterable()) {
    ...
}

The implementation as proposed by @hardtoe didn't include `#autoPagingIterable()` and instead had each collection type implement `Iterable<T>` directly, but I've modified it a bit so that it looks a little more like the Ruby and Python bindings.

For reference, the current mechanism looks like this:

``` java
List<Invoice> invoices = Invoice.all(...).getData();
foreach (Invoice invoice : invoices) {
    ...
}

// manual pagination here ...
```

The code is still not ready for review, but let me know if you have any early comments/objections on this interface. Thanks!

rickbert:
Great work guys!  A couple of minor questions: 
- do you also want HasId to expose the setter?
- although I realize id is String by convention Stripe-land, would it make sense to do the following
```
public interface HasId<T> { 
    public T getId();
    public void setId(T id);
}
```

kyleconroy:
cc @shalecraig who always has Java feels

kyleconroy:
Having to put `HasId` everywhere is quite annoying. What if we just made APIResource implement HasId and handle the case where an ID is null?

src/main/java/com/stripe/model/StripeCollectionAPIResource.java
```diff
@@ -33,6 +34,9 @@ public void setHasMore(Boolean hasMore) {
 	public String getURL() {
 		return url;
 	}
+	public String getUrl() {
```
kyleconroy: Merge artifact? This is here twice.

hardtoe:
I think that's the way to go. I wasn't very happy with HasId myself but I
wasn't sure what the consequences would be to change APIResource.

kyleconroy:
@brandur I'm :+1: to that interface. After looking at the code, we should have both StirpeObject and APIResource implement HasId and possibly pick use @rickbert's suggestion about updating the interface to be more generic.

russelldavis:
>  I'm :+1: to that interface

+1

rickbert:
Agreed, much better to impl it in StripeObject and APIResource rather than sprinkling it everywhere.  #nosprinkles

:+1:

src/main/java/com/stripe/model/StripeCollectionAPIResource.java
```diff
@@ -33,6 +34,9 @@ public void setHasMore(Boolean hasMore) {
...
+		return url;
+	}
```
shalecraig: Can we deprecate one of `getURL` or `getUrl` them, and have the deprecated one call the non-deprecated one?

src/main/java/com/stripe/model/StripeCollectionIterator.java
```diff
@@ -0,0 +1,80 @@
...
+	@SuppressWarnings("rawtypes")
+	private final Class<? extends StripeCollectionInterface> collectionType;
+	
+	private StripeCollectionInterface<T> currentCollection;
+	private Iterator<T> currentDataIterator;
+	
+	private String lastId;
+	
+	StripeCollectionIterator(final StripeCollectionInterface<T> stripeCollection) {
+        this.url = Stripe.getApiBase() + stripeCollection.getUrl();
```
shalecraig: weird indenting.

brandur:
Ack all comments here! Thanks for taking a look everybody, and I'll respond to each note specifically hopefully a little later today.

While trying to get request parameters and options plumbed through to subsequent pages I ended up on an epic journey of refactoring which I'd been hoping not to get into. It's delayed things a little bit, but I'm getting closer.

src/main/java/com/stripe/model/StripeCollectionAPIResource.java
brandur: Yep! Done in 46af413.

brandur:
Hey everyone, please consider this ready for final review. Unfortunately, I had to make some changes to how lists are invoked which ended up in another incredibly monstrous diff with all the refactoring involved. Sorry about that :$

Please see responses to various questions/comments below:

> do you also want HasId to expose the setter?

@rickbert I think I'm going to leave this off until necessary. `HasId` is largely for pagination right now, and I'll try to wait for a need to justify the expansion of this interface.

> although I realize id is String by convention Stripe-land, would it make sense to do the following

@rickbert Good call, but I think that it's pretty unlikely that we'll ever see non-string identifiers at any point in the future. If this does become necessary, we can always deprecate the current interface and moved us onto a generic version.

> Having to put HasId everywhere is quite annoying. What if we just made APIResource implement HasId and handle the case where an ID is null?

@kyleconroy @hardtoe @rickbert I started off down this path, but ended up reverting it in the end. I think it's kind of nice that you get a compile-time check right now that will prevent you from creating a collection unless you can guarantee that your type can produce an identifier.

The rote repetition is certainly a consideration, but in order to make any changes or additions to this package right now you already have to manually manipulate so many lines of code (see _any_ model class to see what I'm talking about) that I don't think one extra is going to be all that big of a deal.

I've fixed the various problems that @shalecraig called out as well. He also pointed out offline that the recursion in the paging iterator wasn't really necessary so I've removed that as well.

One last note: the tabbing is a freaking disaster right now. This is largely my fault, but I tried to fix it en masse and realized that the project has so many pre-existing tabbing problems that I end up modifying even files that I didn't touch and make the diff even bigger. To get around that, I'll do one more fix-up pass on master after I commit this patch, so please ignore any whitespace problems for now :$

Thanks again everyone!

kyleconroy:
You're right about the spacing issues. I think a one-time pass to fix up all files is a fantastic idea.

rickbert:
Excellent, sounds like you've got a good tradeoff.  Appreciate you following up, and I completely agree that your considerations/compromises w.r.t. setId(...) and genericizing it are reasonable and pragmatic. 

Nice work!

src/main/java/com/stripe/net/APIResource.java
```diff
@@ -130,4 +131,27 @@ public static String urlEncode(String str) throws UnsupportedEncodingException {
...
+
+    /**
+     * Similar to #request, but specific for use with collection types that
+     * come from the API (i.e. lists of resources).
+     *
+     * Collections need a little extra work because we need to plumb request
+     * options and params through so that we can iterate to the next page if
+     * necessary.
+     */
+    protected static <T extends StripeCollectionInterface> T requestCollection(
```
jimdanz: Strong :+1: to this abstraction, feels just right.

brandur: @jimdanz Glad like you like it! It took a fair bit of refactoring, but agreed, the result isn't the worst :)

brandur:
Thanks again for the feedback on this project everyone! I'm pulling this in.

Issue #229, opened 2015-10-13
Lists are not serialized correctly


olivierbellone:
When [creating a product](https://stripe.com/docs/api/java#create_product) with the following code:

```java
Map<String, Object> productParams = new HashMap<String, Object>();
productParams.put("name", "T-shirt");
productParams.put("description", "Comfortable gray cotton t-shirts");
List<String> attributes = Arrays.asList("size", "gender");
productParams.put("attributes", attributes);
Product product = Product.create(productParams);
```

the request body will be encoded like this:
```
{
  description: "Comfortable gray cotton t-shirts",
  name: "T-shirt",
  attributes:
  {
    0: "size",
    1: "gender"
  }
}
```

Because the `attributes` list is not correctly serialized, the product creation then fails with:
```
error:
{
  type: "invalid_request_error",
  message: "Invalid array",
  param: "attributes"
}
```

`attributes` should be serialized as:
```
attributes: [
  "size",
  "gender"
]
```

jloisel:
Happens on Order.all too, trying to pass ids to the request and it fails with invalid array.

jloisel:
Pretty sure it affects all the requests which require to serialize lists.

jloisel:
Can be reproduced using curl:

> curl -XGET https://api.stripe.com/v1/orders?ids%5B0%5D=<OrderId>&status=paid -u <Test Key>

brandur:
Wow, was just looking into this a bit and it seems that [this was done on purpose](https://github.com/stripe/stripe-java/blob/brandur-pagination/src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java#L55,L66). As you can see in that test, it knowingly changes a list into an object instead of projecting it into an array.

I'm not sure why (maybe the server is supposed to support this?), but will try to work my way down the list of top contributors to ask when I get to the office.

/cc @kyleconroy In case he has any additional context.

kyleconroy:
@brandur Indeed. This is how `/v1/accounts` works. We should check in on the other bindings to make sure that order creation is working properly in the other bindings. We should probably put this special case in the Account class and make the general case do the correct thing.

brandur:
[...]

PR #231, opened 2015-10-13
Fix encoding of Lists being sent in via API requests

brandur:
As described in #229, we were previously translating lists to a special type of indexed object when sending them into the API via a request.  This patch corrects that behavior, adds some significant clean-up to the encoding utilities, and adds some additional testing so that we'll have better guarantees that everything works.

@kyleconroy As far as I can tell, there isn't anything special-cased around `LegalEntity`'s `additionalOwners` accessor and/or parameters, meaning that you'd probably want to modify these the "usual" way (I think):

``` java
Account account = Account.retrieve("acct_16ozEuCdGbJFKhCu");
LegalEntity entity = account.getLegalEntity();
List<Owner> additionalOwners = entity.getAdditionalOwners();
Owner owner = additionalOwner[3];

//
// set the information we'd like to see
//

Map<String, String> ownerParams = new HashMap<String, String>();
ownerParams["first_name"] = "Stripe";

//
// and then start nesting parameters like crazy ...
//

Map<String, Object> additionalOwnerParams = new HashMap<String, Object>();
additionalOwnerParams["3"] = ownerParams;

Map<String, Object> legalEntityParams = new HashMap<String, Object>();
legalEntityParams["additional_owners"] = additionalOwnerParams;

params["legal_entity"] = legalEntityParams;

owner.update(params);
```

This is obviously terrible, but I haven't seen any evidence yet that having the currently broken List-handing logic would make it any better. Any thoughts?

### DIFF
--- a/src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
+++ b/src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
@@ -68,27 +68,27 @@ public class LiveStripeResponseGetterTest {
 		assertEquals(encode("nested[]=A&nested[]=B&nested[]=C&a=b&c=d"), srg.createQuery(params));
 	}
 
-    // It's really arrays of hashes where things in
-    // application/x-www-urlencoded-form start to get *really* hairy, so let's
-    // have a special test to cover thsi type of case.
+	// It's really arrays of hashes where things in
+	// application/x-www-urlencoded-form start to get *really* hairy, so let's
+	// have a special test to cover thsi type of case.
 	@Test
 	public void testCreateQueryWithComplexParams() throws StripeException, UnsupportedEncodingException {
-        List<String> deepNestedList = new LinkedList<String>();
+		List<String> deepNestedList = new LinkedList<String>();
 		deepNestedList.add("A");
 		deepNestedList.add("B");
 
-        Map<String, String> deepNestedMap1 = new LinkedHashMap<String, String>();
+		Map<String, String> deepNestedMap1 = new LinkedHashMap<String, String>();
 		deepNestedMap1.put("C", "C-1");
 		deepNestedMap1.put("D", "D-1");
 
-        Map<String, String> deepNestedMap2 = new LinkedHashMap<String, String>();
+		Map<String, String> deepNestedMap2 = new LinkedHashMap<String, String>();
 		deepNestedMap2.put("C", "C-2");
 		deepNestedMap2.put("D", "D-2");
 
 		List<Object> nested = new LinkedList<Object>();
-        nested.add(deepNestedList);
-        nested.add(deepNestedMap1);
-        nested.add(deepNestedMap2);
+		nested.add(deepNestedList);
+		nested.add(deepNestedMap1);
+		nested.add(deepNestedMap2);
 
 		/* Use LinkedHashMap because it preserves iteration order */
 		Map<String, Object> params = new LinkedHashMap<String, Object>();
--- a/src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
+++ b/src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
@@ -52,17 +52,49 @@ public class LiveStripeResponseGetterTest {
 	}
 
 	@Test
-	public void testCreateQueryWithList() throws StripeException, UnsupportedEncodingException {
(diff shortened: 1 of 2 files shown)


kyleconroy:
@brandur looks like the whitespace is off. `stripe-java` uses tabs.

brandur:
@kyleconroy Thanks! Fixed in 7b314b7.

src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
```diff
@@ -52,17 +52,49 @@ public void testCreateQueryWithNestedParams() throws StripeException, Unsupporte
...
+		List<Object> nested = new LinkedList<Object>();
+		nested.add(deepNestedList);
+		nested.add(deepNestedMap1);
+		nested.add(deepNestedMap2);
+
+		/* Use LinkedHashMap because it preserves iteration order */
+		Map<String, Object> params = new LinkedHashMap<String, Object>();
+		params.put("nested", nested);
+
+		assertEquals(encode("nested[][]=A&nested[][]=B&nested[][C]=C-1&nested[][D]=D-1&nested[][C]=C-2&nested[][D]=D-2"), srg.createQuery(params));
```
kyleconroy: Nesting a list inside a list isn't supported. You can only nest hashes in a list.

```ruby
{
  "nested"=> [
    nil,
    nil, {
      "C"=>"C-1",
      "D"=>"D-1"
    }, {
      "C"=>"C-2",
      "D"=>"D-2"
    }
  ]
}
```

kyleconroy:
This looks good to me. However, this needs to be a version bump, as the following will change

```java
Map<String, String> ownerParams = new HashMap<String, String>();
ownerParams.put("first_name", "Stripe");

List<Object> additionalOwners = new LinkedList<Object>();
additionalOwners.add(ownerParams);

Map<String, Object> legalEntityParams = new HashMap<String, Object>();
legalEntityParams.put("additional_owners", additionalOwners);

params.put("legal_entity", legalEntityParams);

// Current behavior
assertEquals(encode("legal_entity[additional_owners][0][first_name]=Stripe"), srg.createQuery(params));

// New behavior
assertEquals(encode("legal_entity[additional_owners][][first_name]=Stripe"), srg.createQuery(params));
```

kyleconroy:
@brandur I added two more tests (I wrote them to make sure I understood what was going on).

I think we're good to go.

jloisel:
:+1: Well done! Hope this get released soon so we can resume our Stripe integration.

src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
brandur: Fair point. Removed that case and renamed the test to mention "array of hashes" specifically in c7b6486. Thanks!

brandur:
> This looks good to me. However, this needs to be a version bump, as the following will change

Fair enough, I'll bump the major version. Thanks for taking a look!

In general, it would great if we could assemble a much better story around updating values in some of those deeply nested resources. What's in there right now is pretty janky no matter which way you do it.

brandur:
Holding off for a bit longer on this release because it seems that there is a chance that we'll introduce a problem elsewhere. See also stripe/stripe-ruby#340.

jloisel:
I think you should fix the server, it should accept both ids[0]=1&ids[1] and ids[]=1&ids[]=2.

brandur:
@jloisel The trouble with the integer-index model (i.e. `ids[0]=1&ids[1]`) is that there's no way for the server to disambiguate between a replacement of the entire array and a replacement of each element in the array.

From your example above, if we start with a server side value of:

    ids = [1, 2, 3]

And we send `ids[]=1&ids[]=2`. We can unambiguously treat this as an array replacement, so we're left with:

    ids = [1, 2]

But if the server receives `ids[0]=1&ids[1]`, then we must treat that as a replacement of elements in order to prevent unintended loss of data. Therefore we're left with the same server-side value that we started with:

    ids = [1, 2, 3]

We'll eventually try to fully support the `ids[0]=1&ids[1]` syntax for replacement, but unfortunately as is, the implementation was not correct (hence this pull).

jloisel:
I wasn't aware that your rest api was stateful, keeping previous state in memory. I thought the rest api would be stateless. Or maybe you are referring to a common server side implementation which is both used by the stateless rest api and by stateful components.

PR #232, opened 2015-10-20
Fix non-deterministic test

brandur:
Something has changed recently (probably in the server-side
implementation on which these bindings rely) where a specific sorting
of order items coming back in an order is no longer guaranteed. These
tests were written so that in a list of (1) sku, (2) shipping, and (3)
tax, it was always assumed that the sku would appear first. This
assumption no longer holds up.

Patch now searches order items for sku and compares against that instead
of assuming that it's in 0th position.


brandur:
And more generally, I'm having a pretty hard time getting Travis to pass. Lots of random failures due to request timeouts.

kyleconroy:
Stripe Java and PHP are the only libraries left that make live API calls. I
think for the long term health of these projects, we should probably stop
making a majority of these requests.

PR #234, opened 2015-10-23
Encode empty lists as blank strings

brandur:
It's currently impossible to empty a list via the Java bindings due to the fact
that an empty list gets "skipped" by the parameter encoder. This problem is
described more completely in [T7312][T7312].

This patch corrects that behavior with a contribution from @olivierbellone and
adds regression testing.

For reference, [here's the server-side code that demonstrates how the list 
unsetting actually works for `additional_owners`][pay-server].

[pay-server]: https://github.com/stripe-internal/pay-server/blob/6f5c81a9e85f1a1ff028cdeade688f7794d95d61/api/lib/account_common.rb#L1141,L1142
[T7312]: https://phab.stripe.com/T7312


kyleconroy:
:+1:

PR #235, opened 2015-11-13
Add osgi support

AntoineDcode:
Use maven-bundle-plugin to add OSGI spécific instructions into the manifest

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,20 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-gpg-plugin</artifactId>
+                <version>1.5</version>
+                <executions>
+                    <execution>
+                        <id>sign-artifacts</id>
+                        <phase>verify</phase>
+                        <goals>
+                            <goal>sign</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.1</version>
                 <configuration>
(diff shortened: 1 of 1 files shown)


brandur:
@AntoineDcode Would you mind giving us a few more details around the background and motivation for this one? Thanks!

Also, it looks like there's a build failure on OpenJDK 6. Do you know if it's supported?

AntoineDcode:
Hi,

We are planning to integrate the stripe-java sdk in an application which is deployed inside a apache-karaf container which use OSGi under the hood .
To allow OSGi to wire correctly dependencies, the jar manifest must have some specific instructions to describes the needed packages to import and/or import.

As the maven-bundle-plugin only add OSGi specific instructions, the jar will still be compatible with any other application.

Regarding the OpenJDK 6 failure, it has been caused by using a wrong version of maven-bundle plugin.

brandur:
@AntoineDcode Hey! We don't have a whole lot of experience with OSGi, but [this answer on Stack Overflow](http://stackoverflow.com/questions/1340483/osgi-handling-3rd-party-jars-required-by-a-bundle) recommends not bundling up every JAR dependency. Instead, they suggest the use of a [bnd tool](http://www.aqute.biz/Bnd/Bnd) to build the bundles up for you. Could you investigate the feasibility of this approach?

AntoineDcode:
Hi @brandur,

The [maven-bundle-plugin](http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html) use bnd under the hood (and mentioned in this [reply](http://stackoverflow.com/a/22840650)). So the dependencies are not bundling into the result jar.

brandur:
> The maven-bundle-plugin use bnd under the hood (and mentioned in this reply). So the dependencies are not bundling into the result jar.

Hey Antoine! Sorry, I'm not sure I really understand, can you elaborate a bit? As far as I can tell, it looks like that SO answer mentioned that dependencies _are_ bundled and talks about the possibility of suppressing them if so desired.

brandur:
I'm going to close this one for now, but let us know if there's some additional discussion that we should be having here. Thanks!

PR #239, opened 2015-11-27
Fix PagingIterator to work after 2 iterations

the-fine:
Fixes #237

Issue #237, opened 2015-11-26
Issue with autoPagingIterable() and RequestOptions


olivierbellone:
(Using stripe-java 1.38 and gson 2.2.4.)

This code:
```java
RequestOptions options = RequestOptions.builder().setApiKey("sk_test_...").build();

Map<String, Object> planParams = new HashMap<String, Object>();
planParams.put("limit", 2);

Iterable<Plan> itPlans = Plan.list(planParams, options).autoPagingIterable();

for (Plan plan : itPlans) {
    System.out.println(plan.getId());
}
```

will list the first 4 plans, then fail with the following exception:
```
java.lang.RuntimeException: Unable to lazy-load stripe objects
	at com.stripe.model.PagingIterator.next(PagingIterator.java:60)
	at com.stripe.model.PagingIterator.next(PagingIterator.java:12)
	at StripeExample.main(StripeExample.java:40)
Caused by: com.stripe.exception.AuthenticationException: No API key provided. (HINT: set your API key using 'Stripe.apiKey = <API-KEY>'. You can generate API keys from the Stripe web interface. See https://stripe.com/api for details or email support@stripe.com if you have questions.
	at com.stripe.net.LiveStripeResponseGetter._request(LiveStripeResponseGetter.java:409)
	at com.stripe.net.LiveStripeResponseGetter.request(LiveStripeResponseGetter.java:65)
	at com.stripe.net.APIResource.request(APIResource.java:131)
	at com.stripe.model.PagingIterator.list(PagingIterator.java:83)
	at com.stripe.model.PagingIterator.next(PagingIterator.java:55)
	... 2 more
```

Strangely enough, it works flawlessly when using:
```java
planParams.put("limit", 1);
```
instead.

There is also no issue if the API key is set at the global level:
```java
Stripe.apiKey = "sk_test_..."
```

PR #240, opened 2015-12-01
Elaborate on paging iterator test to demonstrate failure

brandur:
This extends the paging iterator test from two pages to three in order
to get some test coverage on the sections where we pull a page from a
collection that is *not* the original collection. Not having this piece
is what allowed the bug in #237 to surface despite having some testing
in place.

This should be rebased and brought in after the fix for #237 in #239 has
been brought in.


brandur:
@russelldavis Mind taking a review pass on this? Thanks!

russelldavis:
I'm not very familiar with stripe-java, but this lgtm.

brandur:
@russelldavis Thanks!

And no worries about stripe-java familiarity; I just pulled in #239 and rebased against master, so hopefully the red to green transition in CI will speak for itself.

PR #242, opened 2015-12-01
Add external account verify

wbronitsky:
r? @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/ExternalAccount.java
+++ b/src/main/java/com/stripe/model/ExternalAccount.java
@@ -54,6 +54,22 @@ public class ExternalAccount extends APIResource implements HasId {
 		}
 	}
 
+	public ExternalAccount verify(Map<String, Object> params) throws 
+			AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return verify(params, (RequestOptions) null);
+	}
+
+	public ExternalAccount verify(Map<String, Object> params, RequestOptions options) throws
+  		AuthenticationException, InvalidRequestException,
+  		APIConnectionException, CardException, APIException {
+		if (this.getCustomer() != null) {
+			return request(RequestMethod.POST, String.format("%s/verify", this.getInstanceURL()), params, ExternalAccount.class, options);
+		} else {
+			throw new InvalidRequestException("Only customer bank accounts can be verified in this manner.", null, null, null, null);
+		}
+	}
+
 	public ExternalAccount update(Map<String, Object> params) throws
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/model/ExternalAccount.java
```diff
@@ -77,4 +91,8 @@ public DeletedExternalAccount delete(RequestOptions options) throws
 			APIConnectionException, CardException, APIException {
 		return request(RequestMethod.DELETE, this.getInstanceURL(), null, DeletedExternalAccount.class, options);
 	}
+
+	private String getValidateURL() {
```
brandur: Kind of a nit: but putting the URL formatting in a new method like this seems a little unconventional relative to the rest of the library. What do you think about putting the `String.Format` [in-line as seen here](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Invoice.java#L387,L388)?

brandur:
+1 other than the minor comment above and that it seems like your editor may have had a whitespace failure here (check the diff in GitHub if it's not obvious locally). Looks good to come in after those are fixed up. Thanks!

wbronitsky:
@brandur thanks for the review! the test and the spacing should be fixed

src/main/java/com/stripe/model/ExternalAccount.java
```diff
@@ -54,6 +54,22 @@ public String getInstanceURL() {
 		}
 	}
 
+	public ExternalAccount verify(Map<String, Object> params) throws 
+			AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return verify(params, (RequestOptions) null);
+  }
```
brandur: Looks like the indent on this line could still be fixed.

src/test/java/com/stripe/model/ExternalAccountTest.java
```diff
@@ -59,4 +59,26 @@ public void testUnknownExternalAccountDeletion() throws StripeException, IOExcep
 			null, APIResource.RequestType.NORMAL, RequestOptions.getDefault()
 		);
 	}
+
+	@Test
+  public void testVerify() throws StripeException, IOException {
```
brandur: And lastly, mind just converting the leading whitespace in this test to tabs as well? Thanks!

wbronitsky:
@brandur sorry for firing blindly, I don't Java so well. Let me know if I need to change more

brandur:
@wbronitsky NP. Thanks for fixing it up!

brandur:
Released as 1.39.0.

PR #250, opened 2016-01-07
Added missing types to the name-to-classes map

### LINKED ISSUE #249: Some object types not properly type-casted when extracting an event's object
When accessing an event's object, e.g.:
```java
StripeObject stripeObject = event.getData().getObject();
```

the resulting `stripeObject` will be automatically type-casted to the appropriate type in some cases (for instance, `charge` objects are properly type-casted to `Charge`) but not in others (for instance, `application_fee` objects are type-casted to `StripeRawJsonObject`).

PR with a fix incoming.

olivierbellone:
Fixes #249

### DIFF
--- a/src/main/java/com/stripe/model/EventDataDeserializer.java
+++ b/src/main/java/com/stripe/model/EventDataDeserializer.java
@@ -19,26 +19,38 @@ public class EventDataDeserializer implements JsonDeserializer<EventData> {
 	@SuppressWarnings("rawtypes")
 	static final Map<String, Class> objectMap = new HashMap<String, Class>();
 	static {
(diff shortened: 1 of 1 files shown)


olivierbellone:
r? @brandur 

I added all the known object types, including some that currently do not appear in any events, but this way the deserializing method is future-proof in case these types appear in events some day.

I also reorganized the types in the order they currently appear in the API documentation as that was the most convenient way to make sure I wasn't missing any. Sorry for the unnecessary line adds/removals.

brandur:
Great catch Olivier! Thanks for looking into this.

Looks like the failing build is due to a bad expiration year. That should be fixed in master now.

PR #255, opened 2016-01-09
Patch 2

sterfried:
Updated several objects to include missing fields. Added corresponding Hash objects for relevant fields.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -16,6 +16,7 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 	Boolean chargesEnabled;
 	Boolean detailsSubmitted;
 	Boolean transfersEnabled;
+	Boolean debitNegativeBalances;
 	List<String> currenciesSupported;
 	String email;
 	String statementDescriptor;
@@ -30,10 +31,15 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 	String businessName;
 	String businessUrl;
 	String businessLogo;
+	String businessPrimaryColor;
 	String supportPhone;
 	String supportUrl;
 	String supportEmail;
+	String productDescription;
 	Boolean managed;
+	AccountDeclineChargeOn declineChargeOn;
+	AccountTosAcceptance tosAcceptance;
+	AccountTransferSchedule transferSchedule;
 	ExternalAccountCollection externalAccounts;
 
 	public String getId() {
@@ -50,6 +56,10 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 
 	public Boolean getTransfersEnabled() {
 		return transfersEnabled;
+	}
+
+	public Boolean getDebitNegativeBalances() {
+		return debitNegativeBalances;
 	}
 
 	public List<String> getCurrenciesSupported() {
@@ -113,6 +123,11 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 		return businessLogo;
 	}
 
+	public String getBusinessPrimaryColor()
+	{
+		return businessPrimaryColor;
+	}
+
 	public String getSupportPhone()
 	{
 		return supportPhone;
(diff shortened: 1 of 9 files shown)


sterfried:
Looking through the current codebase, it appears that hash objects in the api docs are represented as their own objects in the Java library. I attempted to follow this implementation. If objects need to be renamed or I was mistaken in my approach, feel free to let me know the preferred methodology and I will correct it. Thanks!

brandur:
@sterfried Awesome work, thank-you! And apologies that these bindings are in such rough shape; we appreciate your efforts to get some fixes in.

Cross-referencing with the API reference, everything in here looks like it checks out. One thing though: it looks like standard convention for models is to include a setter for any property (even if it's not necessarily something that the user sets themselves). Would you mind adding setters for the new properties here? Thanks again!

sterfried:
Absolutely, will do so shortly.

sterfried:
Prior to creating the setters for the newly-added fields, I took a look at the current state of LegalEntity.java, I notice that there are no setters for any of the current fields. I'm not sure if I misunderstood your request or if that class is entirely lacking those methods as well.

I'm happy to add setters for each field in that class and any class I've touched (as well as going through other classes and ensuring they also have setter methods [Account.java does not have any]) but wanted to make sure I properly understood your request before doing so.

brandur:
@sterfried Good question! I see where you're coming from now.

And yes, totally right that it seems like some classes like `Account` and `LegalEntity` (looks like a few others too like `InvoiceItem`) don't have setters. That said, going through random models it kind of looks like to me that the majority of them seem to have getter/setter pairs and I suspect that this was the original intent of the library and it may be the couple models that we're talking about that were outliers that were added later without. I'd probably lean toward putting the setters in as well ... what do you think?

sterfried:
Personally, I'm not sure of the utility of having setters for "unsettable" fields are but if that's what the preferred coding style is then I'm happy to follow it and add them to the fields I'm adding, as well as possibly do a quick audit on the other classes and ensure they have them as well. Will do so this evening unless I hear otherwise! :)

certainmagic:
Having setters is really useful for test code. It allows us to create
objects programmatically instead of having to write and parse json. Fwiw,
I'd vote for having lots of setters. (In fact, if you could add one
for ExternalAccount.id, so we can set the id of a card cleanly, that'd be
great.)

thanks,
ab

brandur:
> Personally, I'm not sure of the utility of having setters for "unsettable" fields are but if that's what the preferred coding style is then I'm happy to follow it and add them to the fields I'm adding, as well as possibly do a quick audit on the other classes and ensure they have them as well. Will do so this evening unless I hear otherwise! :)

Ah, yes, I tend to totally agree, but I consider this case to be a bit of an exception because there's no way to set the properties currently short of reflection. If we had a constructor or some other mechanism that allowed them to be assigned values it would be a different story, but like @certainmagic alluded to, as is there's no conventional way to build an object for something like a test (again, short of reflection through JSON deserialization or something). Thoughts?

And don't worry about the adding setters for any properties that we're included in this pull request. We'll get those fixed up separately!

sterfried:
Ah, I definitely see the reasoning behind it now re: test code. Thank you for the clarification, learn something new daily! 

I'll add them later this evening for the additions in this pull request.

sterfried:
It appears the CI checks failed but I don't believe them to be due to the changes that I've committed (appears to be a networking error). I think I saw this happen on an earlier commit as well but it turned green at some point. Is there a way to manually trigger a CI check? Or is this something resulting from my commits?

brandur:
@sterfried Yeah, unfortunately the test suite here is super flaky because it relies on the main Stripe API. For future reference, you can restart it by clicking on "Details" on the right of the failed entry, then click on the refresh icon (looks like an arrow looping in a circle on the right-hand side of the screen) once you're in Travis. I just restarted this one.

sterfried:
Duly noted!

brandur:
Awesome. Thanks again for tackling this @sterfried! Much appreciated.

sterfried:
No problem. Happy to continue contributing, thanks for the assistance in doing so @brandur.

brandur:
NP!

Released as 1.41.0.

PR #266, opened 2016-02-07
Added support for CountrySpec

remistr:
r? @brandur 

I added a few more tests than the previous implementations to ensure sub-maps worked. Also I didn't add `All()` methods as they are deprecated on other resources. We'd need to update the docs or I can add those for `CountrySpec` too.


brandur:
Nice job @remistr!

The only comment that I'd add is similar to the one I put on the Go bindings: you may want to consider additional data types in some places instead of nested `Map`/`List`. Looking at [Charge again](https://github.com/stripe/stripe-java/blob/3f7630cba038edebd58c05352ef9c90dd62684f8/src/main/java/com/stripe/model/Charge.java#L15,L46), it does seem like types are preferred in some places. Feel free to disregard if this is too much of a pain though. Thanks!

brandur:
@remistr Looks great! And I really like the name `VerificationFieldDetails` to help us solve yesterday's problem. Thanks!

PR #269, opened 2016-03-07
Allow params/options to be passed when paging through subcollections

### LINKED ISSUE #264: InvoiceLineItemCollection provided by getLines doesn't get same auth request options with autoPagingIterable
I'm using this library in a multithreaded environment against various keys granted via oauth by Stripe users for my app, so I provide the appropriate key in request options for each call I make. However, it seems like some of the models' methods don't copy these credentials on additional requests that are made.

e.g., I call Invoice.list passing my request options, and then on each invoice I call getLines and use the autoPagingIterable to access all invoice line items. However, for any invoice with enough line items such that an additional request is needed I get an exception like this:

```
Caused by: com.stripe.exception.AuthenticationException: No API key provided. (HINT: set your API key using 'Stripe.apiKey = <API-KEY>'. You can generate API keys from the Stripe web interface. See https://stripe.com/api for details or email support@stripe.com if you have questions.
```

e: I'm using 1.41.0

brandur:
Currently when a list is fetched from the API params and options can be
given to it as long as it's for a top-level type. If a resource includes
a subcollection (e.g. line items under invoices), the current page can
be iterated, but future pages that are fetched do not inherit the params
and options that were given to the parent.

This change suggests a breaking change to any subcollection getters
(e.g. getLines) so that we can return proxy objects which will allow use
a little like the following:

``` java
items = invoice.getLines().autoPagingIterable(params, options)
```

It would also be nice to do the same for top-level resources so that
this:

``` java
invoices = Invoice.list(params, options).autoPagingIterable()
```

Might become this:

``` java
invoices = Invoice.list().autoPagingIterable(params, options)
```

For consistency's sake. This turns out be a little more of a yakshave in
practice though because we need to have the proxy class remember a
collection type as well as the resource's type.

Note also that I've only added a proxy onto a single subcollection
getter, but in fact we'd need to put these only all of them before this
pull could be finalized.

Fixes #264.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -187,9 +187,9 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 		this.transferSchedule = transferSchedule;
 	}
 
-	public ExternalAccountCollection getExternalAccounts()
+	public PagingProxy<ExternalAccount> getExternalAccounts()
 	{
-		return externalAccounts;
+		return new PagingProxy<ExternalAccount>(externalAccounts);
 	}
 
 	public static Account create(Map<String, Object> params)
--- a/src/main/java/com/stripe/model/ApplicationFee.java
+++ b/src/main/java/com/stripe/model/ApplicationFee.java
@@ -113,14 +113,14 @@ public class ApplicationFee extends APIResource implements HasId {
 		this.charge = charge;
 	}
 
-	public FeeRefundCollection getRefunds() {
+	public PagingProxy<FeeRefund> getRefunds() {
 		// API versions 2014-07-26 and earlier render charge refunds as an array
 		// instead of an object, meaning there is no sublist URL.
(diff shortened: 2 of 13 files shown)


brandur:
Also created a patch in 1a326e5 that shows the same concept applied to the top-level lists so that you can do:

``` java
invoices = Invoice.list().autoPagingIterable(params, options)
```

The problem there though is that we get an iterator over `Invoice` instead of `InvoiceCollection`. I'm honestly not sure how big of a problem this is.

brandur:
It looks like that could be an issue in some cases like `ChargeRefundCollection`. It may not be worth trying to bring in for the time being.

brandur:
Ugh, so I updated the rest of this only to realize that it wasn't going to work. It gets us pagination, but notably drops the ability to do non-list operations like this:

``` java
item = invoice.getLines().create(...);
```

I'm going to close this and try something else.

PR #270, opened 2016-03-08
Expose attaching bank account to customer API

papillon97:
Currently we only expose the API to attach a card to a customer, we should do the same for bank accounts.

@jimdanz @garrettf

### DIFF
--- a/src/main/java/com/stripe/model/Customer.java
+++ b/src/main/java/com/stripe/model/Customer.java
@@ -220,10 +220,38 @@ public class Customer extends APIResource implements MetadataStore<Customer>, Ha
 		return createCard(params, (RequestOptions) null);
 	}
 
+	public BankAccount createBankAccount(String token) throws AuthenticationException,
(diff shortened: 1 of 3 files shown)


jimdanz:
lgtm but cc @shalecraig re: the standardization test failure "junit.framework.AssertionFailedError: Methods on APIResources like Customer.createBankAccount should take a final parameter as a RequestOptions parameter."

@shalecraig it looks like `createBankAccount` faithfully follows the style of createCard, which is explicitly whitelisted in the StandardizationTest:
```java
				// Skip the `public static Card createCard(String id) {...` helper method on Customer.
				if (String.class.equals(finalParamType) && parameters.size() == 1 && "createCard".equals(method.getName())) {
					continue;
				}
```

Perhaps we need to repeat that whitelisting for the createBankAccount case?

papillon97:
@jimdanz Whitelisted the createBankAccount case

shalecraig:
Whitelisting sounds peaceful.

PR #279, opened 2016-04-08
Make all references to URLs upper case "URL" instead of "Url"

### LINKED ISSUE #278: Conflicting getter definitions for property "url":
I got this, using 
`java.lang.IllegalArgumentException: Conflicting getter definitions for property "url": com.stripe.model.StripeCollectionAPIResource#getUrl(0 params) vs com.stripe.model.StripeCollectionAPIResource#getURL(0 params)`

the problem in your class `StripeCollectionAPIResource` is:
image

I'm using version 1.48.0 with Maven. This is a Google Cloud Project running with a local jetty, but I don't think it matters.

brandur:
We also remove a deprecated method from a class that had both.

Fixes #278.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -29,11 +29,11 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 	Keys keys;
 	Map<String, String> metadata;
 	String businessName;
-	String businessUrl;
+	String businessURL;
 	String businessLogo;
 	String businessPrimaryColor;
 	String supportPhone;
-	String supportUrl;
+	String supportURL;
 	String supportEmail;
 	String productDescription;
 	Boolean managed;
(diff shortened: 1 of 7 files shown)


brandur:
As seen in #278, the current implementation which contains a class with both cases of "URL" is causing issues in at least one case.

This change is a little annoying because it will require a major version bump, but I kind of think since any fix to #278 would require one anyway, we should just go all the way here and make *all* references to URLs case-consistent.

Thoughts/opinions?

/cc @stripe/api-libraries

lsoares:
great. what is the expected next launch with this fix?

brandur:
@lsoares Depends a bit on review time and whether anyone has grievances about my idea of a major version bump to get this out :$ I'm going to leave it sitting a bit longer in case something comes in (probably until near the end of the work day in Pacific time here, but it should be good to go out after that.

@andrewnelder Hey! Since you've already been looking at this a bit, would you mind giving the patch an official review pass? (Should be pretty minimal in this case.)

andrewnelder:
Tests pass.  No broken references.  Ship it!  +1

brandur:
@andrewnelder Thank-you! Pulling this in.

brandur:
Released as version 2.0.0.

PR #280, opened 2016-04-12
Add some missing attributes

olivierbellone:
This PR adds a bunch of missing attributes to several objects, and deprecates some attributes that have been removed in recent API versions.

To facilitate maintainability, I have reordered the attributes (and their accessors) in the same order as in the API reference, i.e. `id` first, then alphabetic order.

r? @brandur 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/ApplicationFee.java
+++ b/src/main/java/com/stripe/model/ApplicationFee.java
@@ -11,19 +11,22 @@ import com.stripe.net.RequestOptions;
 import java.util.Map;
 
 public class ApplicationFee extends APIResource implements HasId {
+	String id;
+	String account;
 	Integer amount;
+	Integer amountRefunded;
+	String application;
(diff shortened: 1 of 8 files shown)


brandur:
@olivierbellone Amazing! Looks good to me.

> To facilitate maintainability, I have reordered the attributes (and their accessors) in the same order as in the API reference, i.e. id first, then alphabetic order.

This has been a long time coming! One minor comment: in the future, it might be better to split up the field addition/deprecation and alphabetization into separate PRs to make it a little easier on reviewers. The latter produces quite a lot of churn in the code, and makes it so that I don't really have a hope of checking for the former for correctness.

But anyway, thanks for tackling this!

olivierbellone:
I should definitely have done that! Sorry :( I'll keep it in mind for future updates!

brandur:
Released in 2.1.0!

PR #281, opened 2016-04-12
Add `ChargeOutcome` to `Charge`

brandur:
Adds a type for a charge's outcome and puts it into the `Charge` model.

/cc @brahn

### DIFF
--- a/src/main/java/com/stripe/model/Charge.java
+++ b/src/main/java/com/stripe/model/Charge.java
@@ -32,6 +32,7 @@ public class Charge extends APIResource implements MetadataStore<Charge>, HasId 
 	String invoice;
 	Boolean livemode;
 	Map<String, String> metadata;
+	ChargeOutcome outcome;
 	String order;
 	Boolean paid;
 	String receiptEmail;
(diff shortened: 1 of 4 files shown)


brahn:
sweet!  A few n00b questions:
* Am I correct in assuming that the bindings will gracefully handle the situation where:
  * `outcome` is missing entirely?
  * `outcome` is present but nil?
  * One of the keys in the outcome is missing?  (I see that the test treats the case of present but null)

* Given that this object is read-only, are the `setFoo` methods necessary?

Otherwise :+1: though if there's a standard review process for bindings I suspect my lack of java background disqualifies me :-)

brandur:
> Am I correct in assuming that the bindings will gracefully handle the situation where:

Yeah, GSON is pretty forgiving when it comes to how things get parsed. All those situations should do the right thing.

> Given that this object is read-only, are the setFoo methods necessary?

Hey, we just assign setters for everything by convention. Although they may not be useful in that it's effectively a readonly object coming back from the API, they _can be_ useful for users who are trying to check Stripe API behavior from the test suite. There they can synthesize an artificial charge and use the setters to put whatever they want in its outcome without having to resort to writing out JSON or otherwise using reflection.

brandur:
Released as 2.2.0.

PR #282, opened 2016-04-13
add all, create, retrieve methods to Subscriptions, use v1/subs url

jxu577:
cc @kyleconroy @cupcait

### DIFF
--- a/src/main/java/com/stripe/model/Subscription.java
+++ b/src/main/java/com/stripe/model/Subscription.java
@@ -30,6 +30,25 @@ public class Subscription extends APIResource implements MetadataStore<Subscript
 	Double taxPercent;
 	Map<String, String> metadata;
 
+	public static Subscription create(Map<String, Object> params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return create(params, (RequestOptions) null);
+	}
+
+	@Deprecated
+	public static Subscription create(Map<String, Object> params, String apiKey)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return create(params, RequestOptions.builder().setApiKey(apiKey).build());
+	}
+
+	public static Subscription create(Map<String, Object> params, RequestOptions options)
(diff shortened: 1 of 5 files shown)


cupcait:
cc @stripe/api-libraries

src/main/java/com/stripe/model/SubscriptionCollection.java
```diff
@@ -0,0 +1,4 @@
+package com.stripe.model;
+
+public class SubscriptionCollection extends StripeCollection<Customer> {
```
kyleconroy: I believe this should be `StripeCollection<Subscription>`

src/main/java/com/stripe/model/Subscription.java
```diff
@@ -63,7 +82,63 @@ public Subscription cancel(Map<String, Object> params, String apiKey) throws Aut
...
+
+	@Deprecated
+	public static SubscriptionCollection all(Map<String, Object> params,
+			String apiKey) throws AuthenticationException,
+			InvalidRequestException, APIConnectionException, CardException,
+			APIException {
+		return list(params, RequestOptions.builder().setApiKey(apiKey).build());
+	}
+
+	@Deprecated
```
cupcait: Do we actually need to add new deprecated methods here?

kyleconroy: That's a good point. Let's not add them.

src/main/java/com/stripe/model/SubscriptionCollection.java
kyleconroy: I'm also a bit concerned that the tests still passed.

kyleconroy: Any idea why the tests didn't fail?

jxu577: hm I believe the Java tests just verify that the correct call to the API is made but does not actually execute the request and render the response

kyleconroy:
We'll need to add `created` to the Subscription.java. We'll also need to rebase on master. Other than that, looks good to go.

PR #283, opened 2016-04-18
Adds source_types accessor in balance objects

olivierbellone:
r? @brandur 
cc @stripe/api-libraries 

This PR adds an accessor to the `source_types` attribute in the `Money` class used by `Balance` objects.

It also adds a basic test to ensure that this new attribute is deserialized correctly.

### DIFF
--- a/src/main/java/com/stripe/model/Money.java
+++ b/src/main/java/com/stripe/model/Money.java
@@ -1,14 +1,86 @@
 package com.stripe.model;
 
-public class Money {
+public class Money extends StripeObject {
 	Long amount;
 	String currency;
+	SourceTypes sourceTypes;
 
 	public Long getAmount() {
 		return amount;
 	}
 
+	public void setAmount(Long amount) {
+		this.amount = amount;
+	}
+
 	public String getCurrency() {
 		return currency;
+	}
+
+	public void setCurrency(String currency) {
+		this.currency = currency;
+	}
+
+	public SourceTypes getSourceTypes() {
+		return sourceTypes;
+	}
+
+	public void setSourceTypes(SourceTypes sourceTypes) {
+		this.sourceTypes = sourceTypes;
+	}
+
+	public static class SourceTypes extends StripeObject {
+		Long alipayAccount;
+		Long bankAccount;
+		Long bitcoinReceiver;
+		Long card;
+
+		public Long getAlipayAccount() {
+			return alipayAccount;
+		}
+
+		public void setAlipayAccount(Long alipayAccount) {
+			this.alipayAccount = alipayAccount;
+		}
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/model/Money.java
```diff
@@ -1,8 +1,9 @@
 package com.stripe.model;
 
-public class Money {
+public class Money extends StripeObject {
 	Long amount;
 	String currency;
+  SourceTypes sourceTypes;
```
brandur: So it's definitely a little unusual, but the Java code here uses tabs instead of spaces. Maybe just do a global find and replace on all the files modified in your diff.

src/main/java/com/stripe/model/Money.java
```diff
@@ -11,4 +12,44 @@ public Long getAmount() {
...
+    return sourceTypes;
+  }
+
+  public static class SourceTypes extends StripeObject {
+    Long alipayAccount;
+    Long bankAccount;
+    Long bitcoinReceiver;
+    Long card;
+
+    public Long getAlipayAccount() {
```
brandur: Just in the interest of convention, would you mind adding setters for these as well? (e.g. `setAlipayAccount`) They won't be used in any production paths, but could be useful for testing and such.

src/test/java/com/stripe/model/BalanceTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.stripe.BaseStripeTest;
+import com.stripe.exception.StripeException;
+import com.stripe.model.Balance;
+import com.stripe.model.Money;
+import com.stripe.net.APIResource;
+import com.stripe.net.LiveStripeResponseGetter;
+import com.stripe.net.RequestOptions.RequestOptionsBuilder;
+import com.stripe.net.RequestOptions;
+import org.junit.After;
```
brandur: Mind just re-organizing these imports so that the `org.junit` stuff lives with the other import of the same kind below?

brandur:
@olivierbellone Thanks for tackling this! I left some minor comments, but it looks good to me!

olivierbellone:
@brandur Here you go!

brandur:
@olivierbellone Hey! Just to clarify the leading whitespace thing: I meant that in general for the entire project, tabs are used, so it might be a good idea to use them in your new files as well. So find and replace `s/  /\t/g` rather than vice versa. Sound reasonable?

olivierbellone:
@brandur: Ah, gotcha, sorry! Both `.java` files should use tab indents now.

brandur:
Thanks Olivier!

brandur:
Released as 2.3.0.

PR #284, opened 2016-04-19
Reorder attributes and accessors by alphabetic order

olivierbellone:
r? @brandur 
cc @stripe/api-libraries 

This is a big PR but it should have no impact. I just moved existing code around to order attributes and their accessors by alphabetic order. This will make it much easier to add and remove attributes as the API evolves.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -13,49 +13,72 @@ import java.util.Map;
 
 public class Account extends APIResource implements HasId, MetadataStore<Account> {
 	String id;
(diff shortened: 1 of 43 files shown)


brandur:
Awesome! Thanks a lot @olivierbellone :)

For the record, this is a tad risky because it's unlikely that I'd ever be able to see a mistake on review, but IMO also a well worthwhile improvement, so let's do it.

olivierbellone:
Yep. I tried to be as careful as possible, and resisted the temptation to do any other changes no matter how minor (e.g. adding setters for some attributes that only have getters).

PR #291, opened 2016-05-11
Fixes the case where previous_attributes is null

### LINKED ISSUE #290: "Not a JSON Object: null" when parsing customer.source.updated event
I've set up an Apache CXF rest service to receive Stripe webhooks. The CXF service just gives me the json String which I then pass to the Stripe Java SDK to unmarshall the json into an Event object. 

This works fine if I attempt to do this with a charge.succeeded event. However, when attempting to unmarshall a customer.source.update event it fails with the following error: "java.lang.IllegalStateException: Not a JSON Object: null".

I am testing this with the Stripe Dashboard's webhooking testing feature. This is the json being produced by the Stripe Dashboard when I tell it to send a test customer.source.update event:

```json
{
  "created": 1326853478,
  "livemode": false,
  "id": "evt_00000000000000",
  "type": "customer.source.updated",
  "object": "event",
  "request": null,
  "pending_webhooks": 1,
  "api_version": "2016-03-07",
  "data": {
    "object": {
      "id": "card_00000000000000",
      "object": "card",
      "address_city": "Austin",
      "address_country": "USA",
      "address_line1": "4500 Ave D",
      "address_line1_check": "pass",
      "address_line2": null,
      "address_state": "TX",
      "address_zip": "78751",
      "address_zip_check": "pass",
      "brand": "Visa",
      "country": "US",
      "customer": "cus_00000000000000",
      "cvc_check": null,
      "dynamic_last4": null,
      "exp_month": 12,
      "exp_year": 2016,
      "funding": "credit",
      "last4": "4242",
      "metadata": {
      },
      "name": "Testy McTester",
      "tokenization_method": null
    },
    "previous_attributes": null
  }
}
```
 
I went through it in a debugger and it appears to be choking on the 'previous_attributes' element.

This is the Stripe SDK code that I'm calling. I've verified that the responseJsonString variable is populated with the JSON string:

```java
Event event = Event.GSON.fromJson(responseJsonString, Event.class);
```

Any ideas?

olivierbellone:
Fixes #290.

In practice, an event's `previous_attributes` should never be set to `null` -- it should either be entirely absent from the JSON payload, or have a non-null value if present.

However, some of the dummy events sent by the "Send test webhook" button do have `previous_attributes=null`. Ideally, we should also fix this, but this PR does make the event deserializer a little more robust.

r? @kyleconroy 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/EventDataDeserializer.java
+++ b/src/main/java/com/stripe/model/EventDataDeserializer.java
@@ -110,11 +110,13 @@ public class EventDataDeserializer implements JsonDeserializer<EventData> {
 			String key = entry.getKey();
 			JsonElement element = entry.getValue();
 			if("previous_attributes".equals(key)) {
-				Map<String, Object> previousAttributes = new HashMap<String, Object>();
-				if (element.getAsJsonObject() != null) {
+				if (element.isJsonNull()) {
+					eventData.setPreviousAttributes(null);
+				} else if (element.isJsonObject()) {
+					Map<String, Object> previousAttributes = new HashMap<String, Object>();
 					populateMapFromJSONObject(previousAttributes, element.getAsJsonObject());
+					eventData.setPreviousAttributes(previousAttributes);
 				}
-				eventData.setPreviousAttributes(previousAttributes);
 			} else if ("object".equals(key)) {
 				String type = element.getAsJsonObject().get("object").getAsString();
 				Class<StripeObject> cl = objectMap.get(type);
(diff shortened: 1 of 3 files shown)


brandur:
Thanks for digging into this Olivier! :)

kyleconroy:
A simple test case would be nice, but not required. LGTM

olivierbellone:
@kyleconroy Added a test as you suggested.

kyleconroy:
👍

PR #294, opened 2016-05-20
Remove StripeAPIResourceCollection

### LINKED ISSUE #292: "Unable to lazy-load stripe objects" for TransferTransactionCollection.autoPagingIterable()
I may be doing this incorrectly, but when I use this flow:

	TransferCollection transfers = Transfer.list(transfer_params, request_options);
	Iterable<Transfer> transfer_i = transfers.autoPagingIterable();
	Iterator<Transfer> transfer_it = transfer_i.iterator();
	while (transfer_it.hasNext()){
		Transfer transfer = transfer_it.next();
		TransferTransactionCollection transactions = transfer.transactions(new HashMap<String, Object>(), request_options);
		Iterable<TransferTransaction> transactions_i = transactions.autoPagingIterable();
		Iterator<TransferTransaction> transactions_it = transactions_i.iterator();
		while (transactions_it.hasNext()){
			TransferTransaction transaction = transactions_it.next();
		}
	}

it works for what appears to be the first "page" but then throws an error: 

		at com.stripe.model.PagingIterator.next(PagingIterator.java:60)
		at com.stripe.model.PagingIterator.next(PagingIterator.java:12)
	Caused by: com.stripe.exception.InvalidRequestException: No such transfer: tr_****************; request-id: req_****************
		at com.stripe.net.LiveStripeResponseGetter.handleAPIError(LiveStripeResponseGetter.java:591)
		at com.stripe.net.LiveStripeResponseGetter._request(LiveStripeResponseGetter.java:443)
		at com.stripe.net.LiveStripeResponseGetter.request(LiveStripeResponseGetter.java:65)
		at com.stripe.net.APIResource.request(APIResource.java:133)
		at com.stripe.net.APIResource.requestCollection(APIResource.java:150)
		at com.stripe.model.PagingIterator.list(PagingIterator.java:83)
		at com.stripe.model.PagingIterator.next(PagingIterator.java:55)
		
I noticed this issue _may_ have been addressed for a different collection for Issue #264 .

I attempted to transpose the format described in https://github.com/stripe/stripe-java/issues/264#issuecomment-199552752 but the TransferTransactionCollection.autoPagingIterable() doesn't accept the same parameters and it threw the same exception in that format (for loop, but with parameters).

I was using v. 1.4.1.1 but updated to 2.5.0 hoping it would be addressed, but no difference.

brandur:
The current implementation has two collection implementations:

* `StripeAPIResourceCollection`: Extends `APIResource`; used when a
  collection needs access to either `request` or `requestCollection`.
* `StripeAPICollection`: Used in all other situations.

The two collection implementations are functionally identical except for
their parent class. This is somewhat error prone because as seen in #292,
it's easy to add something to one and forget about the other.

This pull eliminates `StripeAPIResourceCollection` and moves all
collections over to just extend `StripeCollection`, solving the
duplication problem and reducing code complexity.

The downside is that I've moved visibility of `request` and
`requestCollection` on `APIResource` to public so that they can be
called into from collection classes. It would be better if these had
something like `internal` visibility (a la C#), but Java doesn't really
support anything like that. IMO, it's still a reasonable trade-off and
we should still consider those `request*` methods "effectively" private
to the package and not part of the public API.

Fixes #292.

/cc @olivierbellone Thoughts on this one? Thanks!


olivierbellone:
This looks good!

Re. visibility, would omitting the modifier altogether (i.e. making the methods package-private) work?

brandur:
> Re. visibility, would omitting the modifier altogether (i.e. making the methods package-private) work? 

Good question! So the problem with that one is that the `APIResource` implementation is in `com.stripe.net` while we need to make calls to those request methods in `com.stripe.model`. It may be possible to shuffle some of these types around a bit to share a package, but that would probably be a breaking change so probably best to leave as is :/

olivierbellone:
Ah, I knew I was missing something. I agree that this change is not worth breaking the current namespace structure. Let's :ship: this!

brandur:
@olivierbellone Cool! Thanks for taking a look.

PR #295, opened 2016-05-25
Set up order returns.

rasmusrygaard:
r? @stripe/api-libraries 
cc @jimdanz 

Need to bump the API version to have tests pass since the default `returnOrder` behavior is to automatically refund the charges which was introduced in https://stripe.com/docs/upgrades#2016-02-23

### DIFF
--- a/src/main/java/com/stripe/model/Order.java
+++ b/src/main/java/com/stripe/model/Order.java
@@ -261,4 +261,17 @@ public class Order extends APIResource implements HasId, MetadataStore<Order> {
 		return request(RequestMethod.POST, String.format("%s/pay",
 				instanceURL(Order.class, this.getId())), params, Order.class, options);
 	}
+
+	public OrderReturn returnOrder(Map<String, Object> params) throws AuthenticationException,
+			InvalidRequestException, APIConnectionException, CardException,
+			APIException {
+		return this.returnOrder(params, (RequestOptions) null);
+	}
+
+	public OrderReturn returnOrder(Map<String, Object> params, RequestOptions options) throws AuthenticationException,
+			InvalidRequestException, APIConnectionException, CardException,
+			APIException {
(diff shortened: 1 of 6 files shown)


src/test/java/com/stripe/StripeTest.java
```diff
@@ -2441,6 +2442,9 @@ public void testOrderCreateReadUpdatePay() throws StripeException {
 
 		Order paid = updated.pay(ImmutableMap.<String,Object>of("source", defaultSourceParams));
 		assertEquals("paid", paid.getStatus());
+
+        OrderReturn returned = paid.returnOrder(null);
+        assertEquals(paid.getId(), returned.getOrder());
```
brandur: Mind fixing up the indentation here? Thanks!

src/main/java/com/stripe/model/Order.java
```diff
@@ -261,4 +261,17 @@ public Order pay(Map<String, Object> params, RequestOptions options) throws Auth
...
+			InvalidRequestException, APIConnectionException, CardException,
+			APIException {
+		return this.returnOrder(params, (RequestOptions) null);
+	}
+
+	public OrderReturn returnOrder(Map<String, Object> params, RequestOptions options) throws AuthenticationException,
+			InvalidRequestException, APIConnectionException, CardException,
+			APIException {
+		return request(RequestMethod.POST, String.format("%s/returns",
+				instanceURL(Order.class, this.getId())), params, OrderReturn.class, options);
```
brandur: And same on this line; one extra tab should do the trick.

brandur:
Left a couple minor comments, but this looks good. Thanks!

rasmusrygaard:
@brandur Thanks! Fixed up the indentation

brandur:
Thanks! Released as 2.6.0.

PR #296, opened 2016-06-01
Fix API tests

olivierbellone:
r? @brandur 
cc @stripe/api-libraries 

This PR fixes a couple things in the API tests:

- the Relay tests are run on a different Stripe account. Previously, they replaced the API key globally. This makes the tests fail when run locally (presumably because Travis uses a single thread while when run locally, the tests are parallelized). The Relay tests now specify the key for each API call.

- the `getAllExternalAccounts()` test was run on the Relay account, and expected an `InvalidRequestError` exception, presumably because the Relay account isn't configured as a Connect platform (?). It now runs on the main account, and doesn't expect any exception.

- the country specs test were incorrectly indented

### DIFF
--- a/src/test/java/com/stripe/StripeTest.java
+++ b/src/test/java/com/stripe/StripeTest.java
@@ -2467,8 +2467,8 @@ public class StripeTest {
 		Assert.assertNotNull(accountCollection);
 	}
 
-    @Test
-    public void testCountrySpecRetrieve() throws StripeException {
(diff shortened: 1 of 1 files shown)


brandur:
Oh man, so many questionable things are going on in this codebase :$

Great finds here @olivierbellone! Thanks a lot for taking the time to get all of these fixed up.

LGTM.

PR #302, opened 2016-06-19
Use Project Lombok

olivierbellone:
cc @stripe/api-libraries 

This is a proposal to make use of [Project Lombok](https://github.com/rzwitserloot/lombok) and particularly its [`@Getter` / `@Setter`](https://projectlombok.org/features/GetterSetter.html) and [`@EqualsAndHashCode`](https://projectlombok.org/features/EqualsAndHashCode.html) annotations to reduce the amount of boilerplate code.

This PR removes about 6,000 lines of code, while adding some functionality as some of the classes currently lacks some setters or `equals()` / `hashCode()` methods.

This is obviously a non-trivial change, and I don't suggest it be merged as is, but I think it's an interesting improvement worth exploring.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,7 @@
     "recommendations": [
         "vscjava.vscode-java-pack",
         "shengchen.vscode-checkstyle",
+        "GabrielBB.vscode-lombok",
         "naco-siren.gradle-language"
     ]
 }
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ buildscript {
 plugins {
     id 'java'
     id 'maven'
+    id 'io.franzbecker.gradle-lombok' version '1.11'
     id 'checkstyle'
     id 'net.saliman.cobertura' version '2.5.2'
     id 'com.github.kt3k.coveralls' version '2.8.2'
@@ -54,6 +55,11 @@ jar {
 
         instruction 'Export-Package', '*'
     }
+}
+
+lombok {
+    version = "1.16.22"
+    sha256 = ""
 }
 
 apply from: 'deploy.gradle'
--- a/lombok.config
+++ b/lombok.config
@@ -1,0 +1,1 @@
+lombok.getter.noIsPrefix = true
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -13,6 +13,13 @@ import com.stripe.net.RequestOptions;
 import java.util.List;
 import java.util.Map;
 
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
 public class Account extends APIResource implements HasId, MetadataStore<Account> {
   String id;
   String object;
@@ -33,7 +40,6 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
   ExternalAccountCollection externalAccounts;
   Keys keys;
   LegalEntity legalEntity;
-  Boolean managed;
   Map<String, String> metadata;
   Boolean payoutsEnabled;
   AccountPayoutSchedule payoutSchedule;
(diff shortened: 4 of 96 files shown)


olivierbellone:
Fair point. FWIW, I ran some basic tests with Eclipse and it did seem to handle it well (e.g. autocompletion worked for the generated methods), but I don't know about other IDEs.

I've just found out about the [delombok](https://projectlombok.org/features/delombok.html) tool and updated the PR to use it. This adds a compilation step where Lombok outputs the sources with the generated methods (in the `target/generated-sources/delombok` directory), which I guess should ensure that the methods are visible to IDEs.

This also seems to cause the build to fail due to missing Javadoc on the non-generated methods -- I'm not quite sure what's the interaction between Lombok and Javadoc here, but it's probably worth adding the missing Javadoc anyway.

brandur:
@olivierbellone Crazy effort on the pull request here! Thank-you!

I'm a big fan of how this makes changes and additions to models a lot easier, but even more importantly, _a lot safer_ in that it reduces the probability of typos or forgotten updates to `equals`/`hashCode`.

That said, I definitely share concerns about IDEs. If seems like delombok should handle this problem pretty well, but TBH, the mechanism of copying one source directory into another directory of transformed source seems at first take to be a process that might be somewhat brittle, and prone to error when put into practice. Unless we can get some first hand accounts on how lombok has worked out other projects though, only time would tell on that one; it's quite possible of course that it all fits together very well.

Either way, I think that _something_ like this is the future of where we have to go with the bindings. The only question is whether we want to put in Lombok now, or try to get our own code generator scripts running in the future. These aren't necessarily mutually exclusive ideas either.

olivierbellone:
@ebroder @brandur: Thank you both for the feedback!

My goal with this PR was simply to bring this possible improvement to the attention of the bindings team. I totally understand (and share) the concerns.

Feel free to close the PR :)

brandur:
@olivierbellone Anytime! Given that a lot of work has gone into this PR, I'm disinclined to close it without a more comprehensive feedback gathering process. I've opened an internal discussion thread to try and solicit a few more opinions.

lsoares:
BTW.. I use a lot NetBeans IDE for Java dev. and in fact it supports Lombok natively.

brandur:
@lsoares Cool, thanks for weighing in! We could use all the feedback we can get on this one, so this is a useful data point.

chrylis:
As a user, definitely in favor. Lombok support is becoming more standard, and the Stripe data classes have... *lots* of boilerplate.

brandur-stripe:
This push just came up in our chat channel. Too bad it never got merged, but I'm still very positive on using something like this to tear up our `equals`/`hashCode`/`toString` boilerplate at the very least, and move to something like properties if editors/IDEs seem to be relatively happy.

Evan also brought up this one as an alternative:
https://github.com/google/auto/tree/master/value

evan-stripe:
FTR, I liked AutoValue because it seems to integrate better with IntelliJ out of the box - you have to run a build to trigger generation of the classes, but once that's done you don't need a plugin to materialize the methods for autocomplete or anything

ob-stripe:
Yeah, I've been rebasing the branch onto master every few months to keep it up-to-date, and did it again this morning after seeing the conversation on Slack.

I just skimmed through AutoValue's documentation -- it certainly looks interesting too!

mrmcduff-stripe:
Have we considered just using Kotlin for data classes? It's one line of
code per class, and you get equals/hashCode/toString for free, and you get
100% interop with Java for free.

kjc-stripe:
@evan-stripe Just re-read the autovalue documentation. I didn't realize that it's a compile-time only dependency. I think that makes it far more interesting.

@mrmcduff-stripe I think mixing in Kotlin here would be a mistake. I think it would be strange for some files to be Kotlin and some files to be Java

mrmcduff-stripe:
Why would it be so strange? Kotlin was designed to be mixed in with Java,
and users can call getters, setters, constructors, equals, hashcode, and
everything else just like they would ordinarily.

To clarify the savings: a sample Java data class

public class VideoGame {

    private String name;
    private String publisher;
    private int reviewScore;

    public VideoGame(String name, String publisher, int reviewScore) {
        this.name = name;
        this.publisher = publisher;
        this.reviewScore = reviewScore;
    }

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

    public String getPublisher() {
        return publisher;
    }

    public void setPublisher(String publisher) {
        this.publisher = publisher;
    }

    public int getReviewScore() {
        return reviewScore;
    }

    public void setReviewScore(int reviewScore) {
        this.reviewScore = reviewScore;
    }

    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;

        VideoGame that = (VideoGame) o;

        if (reviewScore != that.reviewScore)
            return false;
        if (name != null ? !name.equals(that.name) :
                that.name != null) {
            return false;
        }
        return publisher != null ?
                publisher.equals(that.publisher) :
                that.publisher == null;

    }

    @Override
    public int hashCode() {
        int result = name != null ? name.hashCode() : 0;
        result = 31 * result + (publisher != null ?
                publisher.hashCode() : 0);
        result = 31 * result + reviewScore;
        return result;
    }

    @Override
    public String toString() {
        return "VideoGame{" +
                "name='" + name + '\'' +
                ", publisher='" + publisher + '\'' +
                ", reviewScore=" + reviewScore +
                '}';
    }
}




And its equivalent in Kotlin:

data class VideoGame(val name: String, val publisher: String, var
reviewScore: Int)


To anyone using a VideoGame class, they'd never know the difference.

kjc-stripe:
What dependencies would you need to make that work?

ob-stripe:
Ola everyone. I'd like to reopen the case for using Lombok in the library.

Google's AutoValue is not a good fit for us, because it generates immutable classes and our models are mutable (they probably shouldn't be, but that's a different story).

Kotlin might work, but when I tried my hand at it a few months back, I ran into issues when interfacing Kotlin code with Java code (specifically when extending `APIResource` and implementing the `MetadataStore` / `HasId` / other interfaces). Admittedly, I didn't spend too much time on this and it might be easy to solve, but I'm not sure the overhead of adding a different language in the library is worth it.

To address some concerns:
- I verified (using [this tool](https://lvc.github.io/japi-compliance-checker/)) that this PR does not change or remove any existing method signatures. It adds 345 new methods (missing getters / setters / `equals` / `hashCode`).
- Lombok is supported by most Java IDEs: https://projectlombok.org/setup/overview. Some do require plugins, but this wouldn't impact library _users_, only developers who work on the library itself. We can add a section in the README about it.

cc @brandur-stripe @evan-stripe @kjc-stripe -- wdyt?

brandur-stripe:
@ob-stripe Very cool idea to use the API compliance checking tool — I had no idea anything like this existed.

My preference would be to default to forward progress, and you've done quite a bit of work here to address everyone's concerns. Ship it.

ob-stripe:
Alright, pulling this in! \o/

PR #303, opened 2016-07-12
Added support for 3D Secure

olivierbellone:
r? @brandur 
cc @stripe/api-libraries @bct

Support for the new server-side 3D Secure API (currently in private beta).

### DIFF
--- a/src/main/java/com/stripe/model/ThreeDSecure.java
+++ b/src/main/java/com/stripe/model/ThreeDSecure.java
@@ -1,0 +1,114 @@
+package com.stripe.model;
+
+import com.google.gson.annotations.SerializedName;
+
+import com.stripe.Stripe;
+import com.stripe.exception.APIConnectionException;
+import com.stripe.exception.APIException;
+import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Map;
+
+public class ThreeDSecure extends APIResource implements HasId {
+	String id;
+	Integer amount;
+	Boolean authenticated;
+	Card card;
+	Long created;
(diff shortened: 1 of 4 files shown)


src/main/java/com/stripe/model/ThreeDSecure.java
```diff
@@ -0,0 +1,114 @@
...
+
+public class ThreeDSecure extends APIResource implements HasId {
+  String id;
+  Integer amount;
+  Boolean authenticated;
+  Card card;
+  Long created;
+  String currency;
+  Boolean livemode;
+  @SerializedName("redirect_url")
```
brandur: Good catch :)

src/main/java/com/stripe/model/ThreeDSecure.java
```diff
@@ -0,0 +1,114 @@
...
+import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Map;
+
+public class ThreeDSecure extends APIResource implements HasId {
+  String id;
```
brandur: Sorry for the nit, but I think the rest of the library uses tabs for indentation instead of the two spaces convention. Would you mind changing over this file and its corresponding test file?

brandur:
M-M-M-M-MONSTER contribution!

Thanks @olivierbellone! I left a minor comment on this one, but looks generally good.

olivierbellone:
@brandur I really need to configure my editor to use tabs for Java files...

Anyway, here you go :)

brandur:
LGTM!

Travis doesn't seem to be in good shape right now (the build here is still starting), but I think we got a pretty good idea about it given that the last build succeeded, so I'm going to pull this in anyway.

brandur:
Released as 2.7.0.

PR #304, opened 2016-08-01
Add Source Support

andrewpthorp:
Add support for Sources to Java.

r? @brandur @spolu @wbronitsky


src/main/java/com/stripe/model/Source.java
```diff
@@ -0,0 +1,258 @@
...
+		return verification;
+	}
+
+	public void setVerification(SourceVerificationFlow verification) {
+		this.verification = verification;
+	}
+
+	// Type-specific getters/setters
+
+	public Map<String, String> getIdeal() {
```
spolu: Do we want to unify those against one and a same attribute?

spolu: getTypeMap() maybe?

andrewpthorp: Agree, that is a good idea! I added `getTypeData()` - PTAL!

spolu: Let's remove the specific one, no?

brandur:
@andrewpthorp If I understand correctly, we introduce a generic source model here, which then has a set of getters that allow access to specific source implementation data as buckets of untyped (string, string) maps. I still don't think I have a great grasp of all the concepts introduced here, but this strikes me as a little undesirable as far as Java conventions tend to go.

Would you mind jotting down a few notes around your design vision here (and possibly address some possible alternative implementations if there are any)?

andrewpthorp:
@brandur So, with APM world, we have different "Sources" (an iDeal source, for instance, differs from a SEPA source). Each source returns a type specific property from our API. So, an `iDeal` source will have an `ideal` hash in the response, while a `sepa` source will have a `sepa` hash in the response (but not an `ideal` hash, or any other type hash).

The params that can be returned for each hash are defined by the actual spec[0]. I think the only other thing we could do here would be to make the type response something more specific (i.e. a `sepaData`, `idealData`, etc), but I don't know if that would be more java-appropriate. 

Thoughts?

[0] https://github.com/stripe-internal/pay-server/blob/301ce036afde40175098e86c9d8d36214a4c88d0/config/sources.yaml#L179-L184

src/main/java/com/stripe/model/Source.java
andrewpthorp: /shrug

I'm really not actually sold on the generic method here. The merchant will need to know what they are dealing with anyway (so they would know what is actually in the HashMap).

Do we want to just have `sepaData`, `idealData`, etc as actual classes that are type-checked? I didn't do this originally because I thought it would be nice to be more generic (so we can add / remove props without cutting a Java release), but it sounds like @brandur doesn't like this idea?

andrewpthorp:
@brandur (from slack) -- We are going to stick with `Map<String, String>` for now. When any given type makes it to public beta, we will add it to the "beta checklist" to type-check the data we get load.

andrewpthorp:
PTAL @brandur

spolu:
(this looks so great. Thanks @andrewpthorp!)

src/main/java/com/stripe/model/SourceDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+	private void populateMapFromJSONObject(Map<String, String> objMap, JsonObject jsonObject) {
+		for(Map.Entry<String, JsonElement> entry: jsonObject.entrySet()) {
+			String key = entry.getKey();
+			String value = entry.getValue().getAsString();
+			objMap.put(key, value);
+		}
+	}
+
+	public Source deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+		throws JsonParseException {
```
brandur: Mind giving this line another indent? (Kind of a per project convention here.)

src/main/java/com/stripe/model/Source.java
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	public void setTypeData(Map<String, String> typeData) {
+		this.typeData = typeData;
+	}
+
+	// APIResource methods
+
+	public String getSourceInstanceURL()
+		throws InvalidRequestException {
```
brandur: And same with the various `throws` in this file.

src/main/java/com/stripe/model/Source.java
```diff
@@ -0,0 +1,216 @@
...
+		if (this.getCustomer() != null) {
+			return String.format("%s/%s/sources/%s", classURL(Customer.class), this.getCustomer(), this.getId());
+		} else {
+			return instanceURL(Source.class, this.getId());
+		}
+	}
+
```
brandur: Mind just tightening up this whitespace?

src/main/java/com/stripe/model/Source.java
```diff
@@ -0,0 +1,216 @@
...
+	public static Source retrieve(String id, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.GET, instanceURL(Source.class, id), null, Source.class, options);
+	}
+
+	@Override
+	public Source verify(Map<String, Object> params) throws
```
brandur: And sorry for the nitpickiness, but mind just moving all `throws` to appear on the next line for consistency?

src/main/java/com/stripe/model/SourceDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.gson.JsonPrimitive;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+import java.util.HashMap;
+
+public class SourceDeserializer implements JsonDeserializer<Source> {
+
+	private void populateMapFromJSONObject(Map<String, String> objMap, JsonObject jsonObject) {
+		for(Map.Entry<String, JsonElement> entry: jsonObject.entrySet()) {
```
brandur: Spacing between `for` and parenthesis here.

brandur:
@andrewpthorp Hey! I left a few minor cosmetic nitpicks above, but this looks good to me! (I'll defer to @spolu to have caught any inconsistencies in the layout of the domain logic here though because I don't know what I'm talking about.)

As discussed in chat, it'll be nice to build some more strongly typed classes for the different source types after this is closer to GA, but this seems okay to iterate on for now.

LGTM.

spolu:
The domain logic LGTM

andrewpthorp:
@brandur I believe these are all done -- mind helping me cut a release?

brandur:
> @brandur I believe these are all done -- mind helping me cut a release?

Thanks! And not at all!

brandur:
Released as 2.8.0.

PR #309, opened 2016-09-01
Changed accountBalance's type to Long

### LINKED ISSUE #308: Wrong customer_balance type
https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Customer.java#L16 
it is specified as Integer, however, it is possible to set a number that exceeds int max value through the dashboard or via API.

olivierbellone:
Fixes #308.

Changed `accountBalance`'s attribute in the `Customer` class from `Integer` to `Long` to avoid overflow issues.

r? @brandur
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Customer.java
+++ b/src/main/java/com/stripe/model/Customer.java
@@ -13,7 +13,7 @@ import java.util.Map;
 
 public class Customer extends APIResource implements MetadataStore<Customer>, HasId {
 	String id;
-	Integer accountBalance;
+	Long accountBalance;
 	String businessVatId;
 	Long created;
 	String currency;
@@ -48,11 +48,11 @@ public class Customer extends APIResource implements MetadataStore<Customer>, Ha
 		this.id = id;
 	}
 
-	public Integer getAccountBalance() {
+	public Long getAccountBalance() {
 		return accountBalance;
 	}
 
-	public void setAccountBalance(Integer accountBalance) {
+	public void setAccountBalance(Long accountBalance) {
 		this.accountBalance = accountBalance;
 	}
 
--- a/src/main/java/com/stripe/model/AlipayAccount.java
+++ b/src/main/java/com/stripe/model/AlipayAccount.java
@@ -16,7 +16,7 @@ public class AlipayAccount extends ExternalAccount {
 	Long created;
 	String fingerprint;
 	Boolean livemode;
-	Integer paymentAmount;
+	Long paymentAmount;
 	String paymentCurrency;
 	Boolean reusable;
 	Boolean used;
@@ -47,11 +47,11 @@ public class AlipayAccount extends ExternalAccount {
 		this.livemode = livemode;
 	}
 
-	public Integer getPaymentAmount() {
+	public Long getPaymentAmount() {
 		return paymentAmount;
 	}
 
-	public void setPaymentAmount(Integer paymentAmount) {
+	public void setPaymentAmount(Long paymentAmount) {
 		this.paymentAmount = paymentAmount;
 	}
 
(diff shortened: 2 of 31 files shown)


brandur:
@olivierbellone Thanks for the fix!

And argh! Obviously this seems needed, but because these are using the "class" scalar types, I think that this is a breaking change. `int` can be implicitly coerced to `long`, but the same isn't true for `Integer` and `Long`.

``` java
class Main {
	public static void main(String[] arg) {
		Integer x = 2;
		Long y = x;
		System.out.println(y);
	}
}
```

```
$ javac Main.java
Main.java:4: error: incompatible types: Integer cannot be converted to Long
                Long y = x;
                         ^
1 error
```

I'm not too adverse to a breaking change if necessary though. Is this pretty important?

src/main/java/com/stripe/model/Customer.java
```diff
@@ -13,7 +13,7 @@
 
 public class Customer extends APIResource implements MetadataStore<Customer>, HasId {
 	String id;
-	Integer accountBalance;
+	Long accountBalance;
```
dukehoops: (lurk) why aren't these variables private?

brandur: @dukehoops By default these fields get a modifier called "package-private" which means that they can only be used internally to the package. It's true that private might be a little better, but at least we aren't leaking an API.

olivierbellone:
Argh indeed :(

In practice, I don't think it's that critical, but it's pretty easy (in test mode ;) ) to cause an account's balance to go above 2^31-1.

brandur:
@olivierbellone Fair enough! Alright, we should definitely decide what to do with this then.

If we do cut a major release, do you think it might be worth making any other tweaks to "long" while we're at it? "Amount" on charge, refund, balance transaction, plan, etc. seems like a possible candidate, although this is obviously a much bigger change.

brandur:
@olivierbellone Friendly bump on this one :) The originally reported bug just got a bump. Let us know if you're short on time and want us to just take this over too.

olivierbellone:
@brandur I'm a bit short on time today and will be OOO tomorrow, so please feel free to take over.

I do think that if we're going to release this as a breaking change, we might as well update all the other fields that need it too. Unless I'm mistaken, the maximum value for a charge is 9,999,999,999 which is above 2^31 - 1.

remistr:
@olivierbellone the maximum charge is 99999999 in cents or $999,999.99 in USD

brandur:
> @olivierbellone the maximum charge is 99999999 in cents or $999,999.99 in USD

Haha, wait. So that means that the maximum possible charge fits inside of a signed integer. Should we just not make this change then?

brandur:
Never mind my last comment. I was missing some context here but have just gather some more from @olivierbellone. I'm going to take a shot at converting some other obvious fields over to 64 bit integers, and then we can get this merged.

brandur:
Alright, I just pushed a change here that wholesale changes anything representing money from an integer to a long so that we're consistent across everything.

This is quite a bit of churn, but the good news is that I don't think that it will be much work for integrations to upgrade. The main methods for retrieving a scalar value out of an `Integer` object are:

* `doubleValue`
* `intValue`
* `longValue`

And the good news is that the `Long` object has an identical set of methods. With any luck, customers won't even need to change any code.

r? @dpetrovics Would you mind taking a review pass on this one?

@remistr Any concerns?

remistr:
No issue from me!

dpetrovics:
Thanks for taking this on, @brandur !

LGTM

brandur:
Travis seems to be having a hard time finding us workers, but I'm able to successfully execute the test suite locally. I'm going to pull this in and cut a release.

brandur:
Released as 3.0.0.

PR #311, opened 2016-09-02
Added support for description, iin, and issuer attributes on the Card model.

andrewnelder:
* Modified Card model to include: `description`, `iin`, and `issuer` attributes.
* Added version bump to v2.8.1


src/main/java/com/stripe/model/Card.java
```diff
@@ -34,6 +34,12 @@
 	String status;
 	String tokenizationMethod;
 
+	// Please note that these field are for internal use only and are not typically returned
+	// as part of standard API requests.
+	String description;
```
zackstripe: Way outside of the scope of this PR but as we're following the getters / setter idiom we should declare these as private.

src/main/java/com/stripe/model/Card.java
```diff
@@ -221,6 +227,18 @@ public void setTokenizationMethod(String tokenizationMethod) {
 		this.tokenizationMethod = tokenizationMethod;
 	}
 
+	public String getDescription() { return description; }
```
zackstripe: Nit. Format is inconsistent with other getters/setters.

zackstripe:
One nit, otherwise LGTM.

src/main/java/com/stripe/model/Card.java
remistr: @zackstripe I think it's due to the way we built the lib. See brandur's comment here: https://github.com/stripe/stripe-java/pull/309#discussion_r77237022

brandur: +1. An explicit `private` would be better, but we these get `package-private` by default and at least won't leak out of the package.

src/main/java/com/stripe/model/Card.java
brandur: Committed a small patch in ad217ff to give these getter/setters a little more breathing room.

brandur:
@andrewnelder Thanks!

BTW, for future reference, no need to bump the version everywhere in the project. We have scripts that'll do it do that for us and it's usually better to use them because they're not likely to miss anything by accident. I'll removed the version bumps here for now.

brandur:
Released as 2.9.0.

PR #314, opened 2016-09-08
Introduce delete variant without arguments

### LINKED ISSUE #312: Account `delete()`-method should not require arguments
According to the documentation, the Account `delete()`-method requires no arguments; however, our implementation requires that you pass at least a `null`-value.

[Stripe API Docs - Delete Account (Java)](https://stripe.com/docs/api/java#delete_account)

[Account.java @ L297 - L307](https://github.com/stripe/stripe-java/blob/728d17f1a26c6b0ac13e83490af9922a006612bc/src/main/java/com/stripe/model/Account.java#L297-L307)

brandur:
Many other models have delete variants that don't require any
parameters, and we indeed reference a currently non-existent
parameterless variant for account's delete in our API documentation.

This patch adds this missing method.

Fixes #312.

r? @andrewnelder Would you mind taking a look at this one? Thanks!

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -294,6 +294,12 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 		return request(RequestMethod.POST, instanceURL(Account.class, this.id), params, Account.class, options);
 	}
 
+	public DeletedAccount delete()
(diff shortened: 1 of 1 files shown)


ob-stripe:
Closing this one in favor of #342, since that one adds another signature (`delete(RequestOptions options)`) and a test.

PR #316, opened 2016-09-15
Java bindings for ApplePayDomain

vijay-singh:
Bindings for ApplePayDomain

### DIFF
--- a/src/main/java/com/stripe/model/ApplePayDomain.java
+++ b/src/main/java/com/stripe/model/ApplePayDomain.java
@@ -55,28 +55,16 @@ public class ApplePayDomain extends APIResource implements HasId {
 		return create(params, (RequestOptions) null);
 	}
 
-	public static ApplePayDomain retrieve(String id) throws AuthenticationException,
-			InvalidRequestException, APIConnectionException, CardException,
-			APIException {
-		return retrieve(id, (RequestOptions) null);
-	}
-
-	public DeletedApplePayDomain delete() throws AuthenticationException,
-			InvalidRequestException, APIConnectionException, CardException,
-			APIException {
-		return delete((RequestOptions) null);
-	}
-
-	public static ApplePayDomainCollection list(Map<String, Object> params)
-			throws AuthenticationException, InvalidRequestException,
-			APIConnectionException, CardException, APIException {
-		return list(params, (RequestOptions) null);
-	}
-
 	public static ApplePayDomain create(Map<String, Object> params, RequestOptions options)
 			throws AuthenticationException, InvalidRequestException,
 			APIConnectionException, CardException, APIException {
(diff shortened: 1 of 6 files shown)


src/main/java/com/stripe/model/DeletedApplePayDomain.java
```diff
@@ -0,0 +1,23 @@
+package com.stripe.model;
+
+
```
brandur: Super nit, but mind just tightening up the whitespace here? Thanks!

src/main/java/com/stripe/model/ApplePayDomain.java
```diff
@@ -0,0 +1,110 @@
...
+
+	public void setDomainName(String domainName) {
+		this.domainName = domainName;
+	}
+
+	public static ApplePayDomain create(Map<String, Object> params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return create(params, (RequestOptions) null);
+	}
```
brandur: Mind grouping these methods together by name? (i.e. Both versions of `create` are colocated in the file.) I know it's kind of a mess elsewhere, but we should be aiming to organize these a little better.

src/main/java/com/stripe/model/ApplePayDomain.java
```diff
@@ -0,0 +1,110 @@
...
+		return request(RequestMethod.DELETE, getInstanceURL(id), null, DeletedApplePayDomain.class, options);
+	}
+
+	public static ApplePayDomainCollection list(Map<String, Object> params, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return requestCollection(getClassURL(), params, ApplePayDomainCollection.class, options);
+	}
+
+	public static String getClassURL() {
```
brandur: I notice that most of the other models with this method don't use `static`. Would you mind making it an instance method for consistency?

Also, can you see if you can change it to private visibility? I don't think there's any reason to leak this interface out of the package.

src/main/java/com/stripe/model/DeletedApplePayDomain.java
vijay-singh: done!

src/main/java/com/stripe/model/ApplePayDomain.java
vijay-singh: done!

src/main/java/com/stripe/model/ApplePayDomain.java
vijay-singh: changed to private visibility

brandur:
Thanks! This looks good to me.

brandur:
Released as 2.10.0.

PR #317, opened 2016-09-16
Add support for multiplan subscriptions

jxu577:
This PR adds bindings support for subscriptions with multiple items:

- Adds SubscriptionItem object and tests

- Add tests to verify passing in items arrays into invoice creation, subscription update, and subscription create


jxu577:
r? @kyleconroy @stripe/api-libraries 

Currently tests for subscription item retrieval and deletion in `StripeTest` are failing because the API requires that you pass in subscriptions as a parameter. Once we remove that restriction those tests should pass.

kyleconroy:
👍

PR #318, opened 2016-09-16
Use example.com as test domain

vijay-singh:
Instead of jackshack.website :)


brandur:
Looks like the added randomness was enough to get that test failure fixed. Thanks!

PR #322, opened 2016-10-05
Transfer Transactions should use requestCollection

dpetrovics:
`Transfer.transactions` was using `APIResource.request` instead of `APIResource.requestCollection`, which meant that the returned `TransferTransactionCollection` didn't have `requestOptions` or `requestParams`. (https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/net/APIResource.java#L163)

These are needed if we want to make subsequent calls to get the next page of data in the collection.

r? @brandur


brandur:
Oh geeze, thanks a lot for tracking this one down @dpetrovics!

LGTM.

Now that you've fixed the problem, want to to try your hand at a release to get this out? We'd probably want to merge it in and bump the patch number.

Issue #320, opened 2016-09-23
StripeTest.java should be split into smaller test files


kyleconroy:
The current file is really larger. We should test each resource in a different file.

dpetrovics:
closed by #323

PR #323, opened 2016-10-12
Splitup StripeTest into smaller classes

dpetrovics:
This one is a bit cumbersome to review, but really just entails a restructuring of our functional tests. Addresses #320.

StripeTest was essentially an enormously long list of tests that actually hit the Stripe API. I haven't added or removed any of them, just did some refactoring to create a `BaseStripeFunctionalTest` class with some default params, and then group tests by class within the `com.stripe.functional` package. I kept the existing unit tests under `com.stripe.model`.

TODO: 
- [x] Update the Readme for the mvn test command to suggest including the entire path for a given test, since there are unit + functional tests with the same class name.
- [x] Running tests locally, `DisputeTest`, `RelayTest`, and `SourceTest` have failures. Confirm + fix.

r? @brandur


brandur:
Oh man, thanks a lot for taking the bullet for the team on this @dpetrovics ;) This must have been one serious grind, but having this stuff in separate files is so much better.

It looks from the TODOs that you may be aware of this, but I notice that there's a number of failures in CI that look like they're probably results of the refactor:
https://travis-ci.org/stripe/stripe-java/jobs/166926265

I wouldn't be surprised if the tests actually started to get order-dependent at some point. It may be that some of these failures are the result of not being able to depend on side effects from previously run tests anymore.

dpetrovics:
OK, all tests should be passing now!

PTAL @brandur

brandur:
LGTM.

Thanks for shaving this yak!

PR #324, opened 2016-10-14
Support 403 status codes

jflinter:
r? @brandur

(also, sorry for the commit weirdness - I accidentally pushed to master, and then immediately pushed another commit reverting the change.)


src/main/java/com/stripe/exception/PermissionsException.java
```diff
@@ -0,0 +1,12 @@
...
+public class PermissionsException extends AuthenticationException {
+
+
+  public PermissionsException(String message, String requestId, Integer statusCode) {
+    super(message, requestId, statusCode);
+  }
+
+  private static final long serialVersionUID = 1L;
+
+}
```
brandur: Mind tightening up the whitespace in here (i.e. kill the excess newlines)? Also, man we need a CI check on this, but I think the rest of the project is using tabs instead of two spaces for indentation.

jflinter:
re-r? @brandur

PR #325, opened 2016-10-18
Added support for outcome[risk_level]

remistr:
r? @brandur

### DIFF
--- a/src/test/resources/com/stripe/model/charge_outcome.json
+++ b/src/test/resources/com/stripe/model/charge_outcome.json
@@ -1,6 +1,7 @@
 {
     "network_status": "approved_by_network",
     "reason": null,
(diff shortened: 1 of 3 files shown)


src/test/resources/com/stripe/model/charge_outcome.json
```diff
@@ -1,5 +1,6 @@
 {
     "network_status": "approved_by_network",
+    "risk_level": "normal",
```
brandur: Sorry, miniscule nit, but given that this list is already alphabetized, can you move `risk_level` down a stop?

remistr: @brandur can't believe I read it 3 times to ensure it was the right order! Fixed!

PR #326, opened 2016-10-23
changing the readme

viveksinghggits:
java version should be 1.6 or later


brandur:
Hi @viveksinghggits, thanks for the patch here! I'm going to pull it in, but I should point out that the current wording of "Java 1.6 and later" is actually also technically correct. Even if "or" is probably more common to use for this sort of phrase around version requirements, the words "and" and "or" are effectively interchangeable here.

PR #327, opened 2016-10-25
charge: add Review type

mduvall:
## What's in here

Similar to https://github.com/stripe/stripe-go/pull/339, however this approach uses the already established `String` IDs without expansion on resources.

r? @brandur


brandur:
@mduvall Thanks! We'll work off this branch.

cc @dpetrovics

src/main/java/com/stripe/model/ExpandableField.java
```diff
@@ -0,0 +1,29 @@
+package com.stripe.model;
+
+//TODO: Should this extend StripeObject?
```
brandur: Hm, any particular reason that it should? I can't see that as being obviously desirable.

brandur:
@dpetrovics Love the additions so far! Amazing work :)

As discussed before, I'll hold off on full review until you get a change to tighten everything up to your own satisfaction. One thing that jumped out at me though: could we move tests for the new expandable deserializer into their own test class instead of tacking them onto "charge"? I find this just makes it easier and gives you a tighter feedback loop if you're trying to debug a problem with the general deserializer specifically.

Issue #268, opened 2016-02-26
Expand support


kinfkong:
I think we should support the expand properties in the objects.

brandur:
Oof, you're absolutely right on this one. Unfortunately, I think we're going to need to commit to a fairly serious re-write of the code as it stands to get something like that in, but I'm going to tag this issue as "future" and leave it open so that we can keep an eye on it.

appmagnetics:
Expand one property works if specified by param passed to APIResource.request 

		Map<String, Object> params = new HashMap<String, Object>();
		params.put("expand[]", "balance_transaction");

Can you confirm that over this interface you cannot use multiple expands like this 

		params.put("expand[]", "balance_transaction");
		params.put("expand[]", "customer");

Since the Map does not allow multiple identical keys, right?

brandur:
@appmagnetics The library will automatically unpack any `List` given as a parameter value into the required Rack-style array encoding. So you should be able to do this:

``` java
		List<String> expands = new LinkedList<String>();
		expands.add("balance_transaction");
		expands.add("customer");

		Map<String, Object> params = new LinkedHashMap<String, Object>();
		params.put("expand", expands);
```

appmagnetics:
Wow! it worked. I appreciate a fast answer from your side!

brandur:
> Wow! it worked. I appreciate a fast answer from your side!

Great! And no worries! That's what we're here for.

simon-marino:
+1 on this, see https://github.com/stripe/stripe-android/issues/81

sowdri:
Needed expand support for `destination` field on `Payout` object.

ob-stripe:
Hi @sowdri. We're currently working on adding support for all expandable attributes. You can see the PR here: https://github.com/stripe/stripe-java/pull/381

The PR should be merged and a new version released verys soon.

ob-stripe:
As of [v5.3.0](https://github.com/stripe/stripe-java/blob/v5.3.0/CHANGELOG#L1-L3), the Java library supports expanding all expandable attributes! 🎆

PR #328, opened 2016-10-28
Expandable Object Support,  Implement on Charge

dpetrovics:
This PR adds support for Expandable Objects and implements them for all expandable fields in the `Charge` class. We do this by creating a generic class `ExpandableField<T>`, that contains a String id and an object of type T. The primary advantage of this approach is that we're able to register just one new type adapter to deserialize any expanded object. This adapter, called `ExpandableFieldDeserializer` will return an `ExpandableField` that either:
- is null if the json value is null
- only has a String id and null `expandedObject` if the json value is a string JsonPrimitive
- has both a String id and a properly typed `expandedObject` for expanded JsonObjects.

Creating a generic deserializer was a bit awkward because of Java's type erasure. We have to perform some slightly magical runtime reflection in order to extract the correct class of the object to be expanded (if the json is an expanded object) so that it's properly deserialized.

A couple other important design decisions:
- We maintain backwards compatibility for all previous getters and setters for expandable fields. (These just get and set String IDs as before).
- There is a new getter and setter for each expandable field, with the same name as before plus "Object" as the suffix. So, for example, for the `applicationFee` property on `Charge` we still support `getApplicationFee` and `setApplicationFee`, but now also have a `getApplicationFeeObject` and `setApplicationFeeObject`.
- Because of the above bullet point, the `ExpandableField` generic class is pretty well encapsulated within the package - users should just use the previous accessors for String props and the new accessors for expanded props.

To use expansion in the api, just append to your params like so: `params.put("expand[]", "customer")`. The default is still to not expand, as you'd expect.

Testing:
- Tests for deserializing a simple Charge with no expansions.
- Tests for deserializing an (unrealistically) fully expanded Charge (sorry about the large .json file).
- Getter/Setter tests for all expandable properties.
-  A handful of functional tests for expanding properties on a Charge.
- `ExpandableFieldDeserializer` tests for null, String, and Object json values.

Addresses #268.
r? @brandur


src/test/java/com/stripe/model/ExpandableFieldDeserializerTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class ExpandableFieldDeserializerTest extends BaseStripeTest {
+
+	private static Gson gson = APIResource.GSON;
+
+	private class TestObject {
```
brandur: Perfect — using a simplified test type instead of one of our more complex actual ones is exactly the right thing to do.

src/main/java/com/stripe/model/Charge.java
```diff
@@ -79,19 +78,53 @@ public void setAmountRefunded(Long amountRefunded) {
...
+	public ApplicationFee getApplicationFeeObject() {
+		if (this.applicationFee==null) {
+			return null;
+		}
+		return this.applicationFee.getExpanded();
 	}
 
-	public void setApplicationFee(String applicationFee) {
-		this.applicationFee = applicationFee;
+	public void setApplicationFeeObject(ApplicationFee c) {
```
brandur: I just wanted to call out this naming scheme (i.e. suffix `Object`) to anyone else who may be reviewing. I think that this is the right thing to given that we call these "expandable objects" in our API reference, but there may be other alternatives in case there's some disagreement.

Notably, this will be our naming scheme going forward for anything with an expansion. New fields like charge review will have a `getReviewObject` method, but not a `getReview`.

src/main/java/com/stripe/model/Charge.java
```diff
@@ -79,19 +78,53 @@ public void setAmountRefunded(Long amountRefunded) {
 	}
 
 	public String getApplicationFee() {
```
brandur: Should we annotate these with deprecated? I could see the argument either way: leaving them in may be convenient for some while taking them out may reduce the cognitive complexity of the library's API.

src/main/java/com/stripe/model/ExpandableField.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+
```
brandur: Sorry for the nit, but mind tightening up the whitespace here? Thanks!

src/main/java/com/stripe/model/ExpandableField.java
```diff
@@ -0,0 +1,35 @@
+package com.stripe.model;
+
+public class ExpandableField<T> {
```
brandur: Could you add a basic comment describing what this class is for? Possibly with a link back to:

https://stripe.com/docs/api#expanding_objects

src/main/java/com/stripe/model/ExpandableField.java
```diff
@@ -0,0 +1,35 @@
...
+    private String id;
+    private T expandedObject;
+
+    public ExpandableField (String id, T expandedObject) {
```
brandur: Whitespace between `Field` and the opening parenthesis.

src/main/java/com/stripe/model/ExpandableFieldDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+                .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+                .create();
+
+        if (json.isJsonNull()) {
+            return null;
+        }
+
+        ExpandableField expandableField;
+
+        // Check if json is a String ID:
```
brandur: Could you make this comment a little more elaborate? e.g. "Check if the JSON data is a string ID. If it is, then that's an indication that this field has not been expanded."

src/main/java/com/stripe/model/ExpandableFieldDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+            JsonPrimitive jsonPrimitive = json.getAsJsonPrimitive();
+            if (jsonPrimitive.isString()) {
+                expandableField = new ExpandableField(jsonPrimitive.getAsString(), null);
+                return expandableField;
+            }
+            else {
+                throw new JsonParseException("ExpandableField is a non-string primitive type.");
+            }
+        }
+        // Check if json is an expanded Object:
```
brandur: And similarly here. e.g. "Check if the JSON data is an object. If it is, it's an indication that the field has been expanded in the response."

src/main/java/com/stripe/model/ExpandableFieldDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+            JsonObject fieldAsJsonObject = json.getAsJsonObject();
+            String id = fieldAsJsonObject.getAsJsonPrimitive("id").getAsString();
+            // We need to get the type inside the generic ExpandableField to make sure fromJson correctly serializes
+            // the JsonObject:
+            Type clazz = ((ParameterizedType) typeOfT).getActualTypeArguments()[0];
+            expandableField = new ExpandableField(id, gson.fromJson(json, clazz));
+            return expandableField;
+        }
+
+        // If json is neither a String nor an Object, error:
```
brandur: And maybe expand on this one with a "We expect all expandable objects to fit the known string-or-object design. If one doesn't, then something may have changed in the API and this code may need to be updated."

src/test/java/com/stripe/model/ChargeTest.java
```diff
@@ -0,0 +1,151 @@
...
+        expandedCharge = APIResource.GSON.fromJson(expandedJson, Charge.class);
+    }
+
+    @Test
+    public void testDeserialize() throws IOException {
+        assertEquals("ch_198c152iMoKf4jyWWYvyqjTX", basicCharge.getId());
+        assertEquals(1477429167, (long) basicCharge.getCreated());
+    }
+
+
```
brandur: Extra line here.

src/test/java/com/stripe/model/ChargeTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertEquals(transfer.getId(), "tr_199IPs2eZvKYlo2C9hMneg1R");
+        assertEquals((long) transfer.getAmount(), 400);
+        //Setters:
+        expandedCharge.setTransfer("newidhere");
+        assertEquals(expandedCharge.getTransfer(), "newidhere");
+        Transfer newTransfer = new Transfer();
+        expandedCharge.setTransferObject(newTransfer);
+        assertEquals(expandedCharge.getTransferObject(), newTransfer);
+    }
+}
```
brandur: Thanks for writing all these tests!

I suspect that thoroughly vetting just the core expansion code is going to be mostly adequate for our purposes. Adding a test for every possible expansion is really just testing the API rather than our own code and I'm not totally convinced that's necessary. Given that they're already written, I'd leave them, but I don't think we should require a new test case for every possible expansion for future patches.

brandur:
cc @ob-stripe as well given that you expressed some interest in the project.

src/main/java/com/stripe/model/Charge.java
remi-stripe: @brandur Since expanding is optional, new fields should have both right? You might still want to get the review id stored in the `review` property instead of expanding it and would ask for `getReview()` in that case no?

src/test/java/com/stripe/model/ChargeTest.java
remi-stripe: I wonder if we do want some custom tests for fields that can be multiple types depending on the situation such as `destination` on a transfer for example? This is out of scope for this PR I think since there are no such fields in the Charge object though.

src/main/java/com/stripe/model/Charge.java
brandur: > @brandur Since expanding is optional, new fields should have both right? You might still want to get the review id stored in the review property instead of expanding it and would ask for getReview() in that case no?

@remi-stripe So I think a little like the Go bindings, you still get an expansion object without an expansion, so you'd go `charge.getReviewObject().getId()`.

dpetrovics-stripe: (The downside to this approach is two getters / setters, but the advantage is more clarity as to what this field actually is)

src/test/java/com/stripe/model/ChargeTest.java
dpetrovics-stripe: Thanks for bringing up the multiple type scenario! As per Slack convo - I agree that it'd be nice to have an implementation where we don't have to redeploy every time a new type is added to the expansion list. Unfortunately I don't think that's really possible without:

a) Keeping some list of string "type" to corresponding Java class, so the deserializer knows, and having it dispatch at run time. Not totally sure if this is possible.
b) Wrapping these properties in a really generic StripeObject, thats quite unclear.

I think whichever approach we use, we can still build on this PR to achieve that functionality.

dpetrovics-stripe: Ah - and just wanted to note that these are model tests, not functional tests - so they're not actually hitting the API. They ingest that large json file and make sure each subfield is properly serialized into the correct object. They also check getters/setters. I agree its a bit verbose - just wanted to be sure that I didn't have a typo somewhere in the model.

dpetrovics-stripe:
Thanks for the comments! I believe I've addressed all of them. 

ptal @brandur

dpetrovics-stripe:
Unfortunately I don't see an obvious way of refactoring that duplication. In order to use the `getID` method, we need to know the parameterized class of the `ExpandableField` - so we'd probably need to make a helper for every class that could be expanded. That might make sense, but I'd be inclined to put that off until we convert more classes to see if it's worth it

brandur:
> Unfortunately I don't see an obvious way of refactoring that duplication. In order to use the getID method, we need to know the parameterized class of the ExpandableField - so we'd probably need to make a helper for every class that could be expanded. That might make sense, but I'd be inclined to put that off until we convert more classes to see if it's worth it

It would be fairly idiomatic in Java to define a non-generic interface for the class so that you can use it without the type parameter. e.g.

``` java
public interface IExpandableField {
    public boolean isExpanded();
    public String getId();
    public void setId(String id);
}

public class ExpandableField<T> implements HasId {
	...
}
```

And then do something like this with a helper on `APIResource`:

``` java
	private void setObjectOrId<T>(String newID, IExpandableField currentObject) {
		if (currentObject == null || (currentObject.isExpanded() && currentObject.GetId() != newID)) {
			return new ExpandableField<T>(newID, null);
		}

		currentObject.SetId(newID);
		return currentObject;
	}

	public void setApplicationFee(String applicationFeeID) {
		setObjectOrId<ApplicationFee>(applicationFeeID, this.applicationFee)
	}
```

(And you may not even need the interface.)

src/test/java/com/stripe/functional/ChargeTest.java
```diff
@@ -25,6 +25,17 @@ public void testChargeCreate() throws StripeException {
     }
 
     @Test
+    public void testChargeCreateExpandBT() throws StripeException {
```
brandur: It looks like this test is named with "BT" while another new one down below uses "BalanceTransaction". Can we just use the full name for consistency? I think it's fine if locals still contain "BT".

brandur:
@dpetrovics Looks good man! I'd say merge ready with just a few other minor tweaks.

Just a couple minor nits around whitespace:

1. Can you make sure that all files are prefixed with tabs instead of four spaces. It seems to be the convention of this project.
2. Can you make sure that you're spacing out your comparison operations (e.g. `this.balanceTransaction == null` instead of `this.balanceTransaction==null`).

ptal @dpetrovics

dpetrovics-stripe:
OK, I ended up being able to abstract out the String setter logic. It turns out that we already have a `HasId` interface, and that by specifying a bound to the type parameter here: https://github.com/stripe/stripe-java/pull/328/files#diff-27d820fcdb9725d8ca505fa0c28d0e74R7 and adding a cast here: https://github.com/stripe/stripe-java/pull/328/files#diff-05f1a464e038a3e3a35204533f0e69beR40 we're good to go. 

I added a `setExpandableFieldID` to `APIResource` which encapsulates the setter loggic and just returns an `ExpandableField<T>` .

dpetrovics-stripe:
Whitespace related changes should all be done now as well.

ptal @brandur

brandur-stripe:
LGTM!

PR #332, opened 2016-11-06
Minor Fixes

ronbak:
Hi,

this merge request fixes some minor issues.
It was created with help of [Feram][repo].

If it helped to improve this repository,
please consider to [star Feram's repo][repo] to show some love! 😊

PS:
The [Feram team][feramhq] is currently working on making this technology
available for everyone. If you want to be one of the first to use it,
please sign up up for the newsletter on [feram.co].

[repo]: https://github.com/feramhq/feram
[feramhq]: https://github.com/feramhq
[feram.co]: http://feram.co

### DIFF
--- a/src/main/java/com/stripe/model/StatusTransitions.java
+++ b/src/main/java/com/stripe/model/StatusTransitions.java
@@ -4,7 +4,7 @@ import com.stripe.net.APIResource;
 
 public class StatusTransitions extends APIResource {
   Long canceled;
-  Long fulfiled;
+  Long fulfilled;
   Long paid;
   Long returned;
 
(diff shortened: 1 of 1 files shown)


ob-stripe:
Unfortunately, it looks like the field is misspelled in the API itself: https://stripe.com/docs/api#order_object-status_transitions-fulfiled.

Thanks for the PR, but we'll have to refuse this one.

ronbak:
But this should definitely be fixed!
A typo in an API sounds like a real pain in the ass.
Which one is the upstream repository?

ob-stripe:
Our language bindings are open-source, but our API itself isn't.

In any case, fixing the typo would be difficult at this point, as some of our users have working integrations that use the misspelled field, and breaking their integrations simply isn't an option. We could fix the name with a new API version, but then that would mean that our Java library would have to support _both_ spellings, introducing more complexity for little value.

ronbak:
>  We could fix the name with a new API version, but then that would mean that our Java library would have to support both spellings

Sounds like the right thing to do … even though it's painful ;-)

brandur:
Hah, see HTTP `Referer` (sic).

PR #337, opened 2017-01-02
Handle Review in Event

strpbrk:
This simple change allows me to fetch and process "review.closed" events.


dpetrovics-stripe:
Thanks for adding this, @strpbrk ! I believe this should also work for "review.opened" events: https://stripe.com/docs/api#event_types-review.opened

LGTM

PR #338, opened 2017-01-03
Added support for network_reason_code to the Dispute object

remi-stripe:
This property is not part of the public API and for internal use only.

r? @brandur-stripe

### DIFF
--- a/src/main/java/com/stripe/model/Dispute.java
+++ b/src/main/java/com/stripe/model/Dispute.java
@@ -25,6 +25,7 @@ public class Dispute extends APIResource implements HasId {
 	Map<String, String> metadata;
 	String reason;
 	String status;
+	String networkReasonCode; // Not part of the public API.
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/stripe/model/Dispute.java
```diff
@@ -138,6 +139,14 @@ public void setStatus(String status) {
 		this.status = status;
 	}
 
+	public String getNetworkReasonCode() {
```
brandur: Mind also commenting this method? I think it will be slightly better because it's more likely to show up in a user's IDE if they happen to be using one (which, in Java, there's a reasonable chance).

Otherwise, LGTM. Thanks!

remi-stripe:
@brandur-stripe PTAL

brandur-stripe:
LGTM.

brandur-stripe:
Released as 3.6.0.

PR #339, opened 2017-01-06
Adds toJson() method

ob-stripe:
r? @brandur-stripe 

This PR adds a `toJson()` method to `StripeObject` which returns the JSON serialization of the object, without the extra fluff included in `toString()`.

It also adds the `object` property to all first-order API resource models, which was previously missing (it could be deduced from the model class, but it was missing from the JSON output).

Serializing objects is useful for interacting with our iOS SDK, which expects some specific endpoints (cf. [here](https://stripe.com/docs/mobile/ios/standard#prepare-your-api)).

The test is pretty 😞 , but the JSON serialization does not guarantee key order (making it hard to compare it to the original JSON dump used to create the instance), and our models don't have `equals` operators that only compare property values.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -10,6 +10,7 @@ import com.stripe.net.RequestOptions;
 
 public class Account extends APIResource implements HasId, MetadataStore<Account> {
 	String id;
+	String object;
 	String businessLogo;
 	String businessName;
 	String businessPrimaryColor;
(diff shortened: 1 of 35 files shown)


brandur:
@ob-stripe Thanks for the changes and background information!

> The test is pretty 😞 , but the JSON serialization does not guarantee key order (making it hard to compare it to the original JSON dump used to create the instance), and our models don't have equals operators that only compare property values.

Honestly, the new test looks fine to me. I probably would have only bothered asserting that any one property was equivalent to the pre-serialized version.

LGTM.

brandur:
Released as 3.7.0.

PR #340, opened 2017-01-20
Remove subscriptions from customer

jack-stripe:
I also made sure that Customer is expandable on Subscription.

### DIFF
--- a/src/main/java/com/stripe/model/Customer.java
+++ b/src/main/java/com/stripe/model/Customer.java
@@ -165,10 +165,17 @@ public class Customer extends APIResource implements MetadataStore<Customer>, Ha
 		this.sources = sources;
 	}
 
+    /**
+     * @deprecated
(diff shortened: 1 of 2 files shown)


brandur:
Hey Jack, thanks a lot for the initiative in putting these together!

Just a couple quick notes/questions:

1. Can you make sure that your whitespace is matching what's in the project already (I think it uses tab characters largely).
2. API version `2017-01-18` isn't actually a real thing is it? https://stripe.com/docs/upgrades
3. I'm not sure that we can change the type of the customer field on subscription without a breaking change :/ That's not the end of the world, but we might want to instead just pool some of these up and release them all in one go.

jack-stripe:
@brandur-stripe - similarly to the equivalent golang PR, I'll hold off on doing anything here until the API change is out.

brandur:
Cool! As with https://github.com/stripe/stripe-go/pull/360, I'm going to close this for now, but feel free to reopen whenever.

PR #342, opened 2017-02-01
Removes unused params from Account.delete() method

ob-stripe:
r? @brandur-stripe

This PR fixes the `Account.delete()` method to not accept any parameters, since they're unneeded (the API doesn't accept any parameters for this request anyway, and the Java example in the [API reference](https://stripe.com/docs/api/java#delete_account) doesn't pass any either).

It also adds a test for the method.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -303,6 +303,18 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 		return request(RequestMethod.POST, instanceURL(Account.class, this.id), params, Account.class, options);
 	}
 
+	public DeletedAccount delete()
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return delete((RequestOptions) null);
+	}
+
+	public DeletedAccount delete(RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return delete(null, (RequestOptions) null);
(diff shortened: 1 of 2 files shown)


ob-stripe:
It just occurred to me that this is, technically, a breaking change. Should I modify the PR to add new method signatures but maintain the previous ones?

remi-stripe:
@ob-stripe yes you should had 2 new methods that call the old ones with `null` as the params instead

src/main/java/com/stripe/model/Account.java
```diff
@@ -303,6 +303,18 @@ public Account update(Map<String, Object> params, RequestOptions options)
...
+	public DeletedAccount delete()
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return delete((RequestOptions) null);
+	}
+
+	public DeletedAccount delete(RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.DELETE, instanceURL(Account.class, this.id), null, DeletedAccount.class, options);
```
remi-stripe: IMO they should both call the other methods. They shouldn't make the request themselves so that there's only one path to a call to `request`

ob-stripe:
Oops, just saw #314 🤦‍♂️

brandur:
Oh man, I'd totally forgotten about #314. Thanks for taking care of this!

LGTM.

brandur:
Released as 3.9.0.

PR #343, opened 2017-02-09
Removed sourced_transfers from balance transaction

jnovak-stripe:
Removed sourced_transfers from balance transaction resource because that field has been deprecated in new versions of the API.

### DIFF
--- a/src/main/java/com/stripe/model/BalanceTransaction.java
+++ b/src/main/java/com/stripe/model/BalanceTransaction.java
@@ -24,9 +24,11 @@ public class BalanceTransaction extends APIResource implements HasId {
 	List<Fee> feeDetails;
 	Integer net;
 	String source;
-	TransferCollection sourcedTransfers;
 	String status;
 	String type;
+
+	@Deprecated
+	TransferCollection sourcedTransfers;
 
 	public String getId() {
 		return id;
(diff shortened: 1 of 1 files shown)


brandur:
Thanks! As you're probably already aware, this is a breaking change and will necessitate a major version bump (which isn't the end of the world, but just FYI). Is there any other helpful information that I should include in the change log like a recommendation replacement?

LGTM.

ob-stripe:
OOC, is there a specific reason why we're not just flagging the attribute and accessors as deprecated with a link to the API changelog? This is how we handle most (all?) other deprecated attributes, e.g.:
- https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Customer.java#L33-L34
- https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Customer.java#L176-L183

jnovak-stripe:
@brandur We're relying on transfer_group to determine linkings. So if you wanted exactly the same functionality, you'd need to do one request to get the bt and charge (then get the transfer_group from it). Then a 2nd API request to query transfers with ?transfer_group=xxx.

Do you have a preference for deprecating vs removing? I removed simply based on ignorance, not a grand plan. Happy to modify PR to use deprecated attribute. Thoughts?

ob-stripe:
The pro of deprecating vs. removing is that users can keep using the latest version of the library regardless of their account's API version. If we remove deprecated attributes, then users with older API versions will be stuck with older library versions, which will be problematic if we introduce new features (since you can use new features even if you're on an older API version).

brandur:
> @brandur We're relying on transfer_group to determine linkings. So if you wanted exactly the same functionality, you'd need to do one request to get the bt and charge (then get the transfer_group from it). Then a 2nd API request to query transfers with ?transfer_group=xxx.

Thanks!

And +1 @ob-stripe. Deprecation is the right way to go here. I should have suggested that.

PR #344, opened 2017-02-09
Deprecate sourced transfers

jnovak-stripe:
Removed sourced_transfers from balance transaction resource because that field has been deprecated in new versions of the API.

Having some Travis trouble so trying a new PR. Original PR is https://github.com/stripe/stripe-java/pull/343


jnovak-stripe:
I can't explain why this passes but the other PR didn't. My hypothesis is that I did a git rebase that confused Travis. Closing the other PR.

r? @brandur

src/main/java/com/stripe/model/BalanceTransaction.java
```diff
@@ -116,6 +118,7 @@ public void setSource(String source) {
 		this.source = source;
 	}
```
ob-stripe: Could you add a Javadoc comment here? Something like this:

```java
/**
 * @deprecated
 * Recent API versions no longer return this field (https://stripe.com/docs/upgrades#2017-01-27).
 * Prefer listing all transfers with the `transfer_group` parameter: https://stripe.com/docs/api/java#list_transfers-transfer_group.
 */
```

Feel free to reword this as you think is appropriate.

jnovak-stripe: added

brandur-stripe:
Makes sense. Thank you!

LGTM.

brandur-stripe:
Released as 3.10.0.

PR #345, opened 2017-02-14
updated link to test files

gauravmokhasi:
there is no longer a single StripeTest.java. individual test files are instead present at the updated location provided


brandur-stripe:
Thanks!

PR #346, opened 2017-02-15
Add invoicing parameters to Subscription and Invoice

jacqueline-stripe:
This PR adds and tests new invoicing related parameters:
Subscription model: billing, days_until_due
Invoice model: billing, due_date, number

I don't believe any changes need to be made for what parameters are accepted in function calls.

### DIFF
--- a/src/main/java/com/stripe/model/Invoice.java
+++ b/src/main/java/com/stripe/model/Invoice.java
@@ -17,6 +17,7 @@ public class Invoice extends APIResource implements MetadataStore<Invoice>, HasI
 	Long applicationFee;
 	Integer attemptCount;
 	Boolean attempted;
+	String billing;
 	ExpandableField<Charge> charge;
 	Boolean closed;
 	Long created;
@@ -25,12 +26,14 @@ public class Invoice extends APIResource implements MetadataStore<Invoice>, HasI
 	Long date;
 	String description;
 	Discount discount;
+	Long dueDate;
 	Long endingBalance;
 	Boolean forgiven;
 	InvoiceLineItemCollection lines;
 	Boolean livemode;
 	Map<String, String> metadata;
 	Long nextPaymentAttempt;
+	String number;
 	Boolean paid;
 	Long periodEnd;
 	Long periodStart;
@@ -91,6 +94,14 @@ public class Invoice extends APIResource implements MetadataStore<Invoice>, HasI
 
 	public void setAttempted(Boolean attempted) {
 		this.attempted = attempted;
+	}
+
+	public String getBilling() {
+		return billing;
+	}
+
+	public void setBilling(String billing) {
+		this.billing = billing;
 	}
 
 	public String getCharge() {
@@ -171,6 +182,14 @@ public class Invoice extends APIResource implements MetadataStore<Invoice>, HasI
 		this.discount = discount;
 	}
 
+	public Long getDueDate() {
+		return dueDate;
+	}
+
+	public void setDueDate(Long dueDate) {
+		this.dueDate = dueDate;
+	}
+
 	public Long getEndingBalance() {
 		return endingBalance;
 	}
@@ -213,6 +232,14 @@ public class Invoice extends APIResource implements MetadataStore<Invoice>, HasI
 
 	public void setNextPaymentAttempt(Long nextPaymentAttempt) {
 		this.nextPaymentAttempt = nextPaymentAttempt;
+	}
+
+	public String getNumber() {
+		return number;
+	}
+
+	public void setNumber(String number) {
+		this.number = number;
 	}
 
 	public Boolean getPaid() {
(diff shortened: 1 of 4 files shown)


brandur-stripe:
@jacqueline-stripe Thanks! This looks generally good to me.

I notice that all these fields are still marked as `dashboard_only` and gated internally. I assume all of these are coming off?

> TODO: investigate tests are correct

Can you see if you can run them successfully and let me know?

jacqueline-stripe:
@brandur-stripe the gates will come off when we launch (Launching the bindings and the feature is a bit of a chicken-and-egg problem). cc @stripe/api-libraries 

r? @pantera-stripe

brandur-stripe:
> @brandur-stripe the gates will come off when we launch (Launching the bindings and the feature is a bit of a chicken-and-egg problem)

Cool! Sounds good.

src/main/java/com/stripe/model/Invoice.java
```diff
@@ -25,12 +26,14 @@
 	Long date;
 	String description;
 	Discount discount;
+	Long dueDate;
```
pantera-stripe: The API docs also list `days_until_due` as a create parameter. Is there a reason we're only exposing  the due date parameter?

src/test/java/com/stripe/functional/InvoiceTest.java
```diff
@@ -256,4 +258,22 @@ public void testInvoiceMetadata() throws StripeException {
...
+        params.put("billing", "send_invoice");
+        params.put("due_date", dueDate);
+        Invoice invoice = Invoice.create(params);
+        assertEquals(invoice.getBilling(), "send_invoice");
+        assertEquals(invoice.getDueDate(), dueDate);
+
+        Map<String,Object> updateParams = new HashMap<String,Object>();
+        updateParams.put("paid", true);
+        Invoice updatedInvoice = invoice.update(updateParams);
+        assertEquals(updatedInvoice.getPaid(), true);
```
pantera-stripe: You can just use `assertTrue(updatedInvoice.getPaid())` here.

src/test/java/com/stripe/functional/InvoiceTest.java
```diff
@@ -256,4 +258,22 @@ public void testInvoiceMetadata() throws StripeException {
...
+    public void testSendInvoice() throws StripeException {
+        InvoiceItem invItem = InvoiceItem.create(getInvoiceItemParams());
+        Map<String,Object> params = new HashMap<String,Object>();
+        Long dueDate = (System.currentTimeMillis() / 1000) + 600;
+        params.put("customer", invItem.getCustomer());
+        params.put("billing", "send_invoice");
+        params.put("due_date", dueDate);
+        Invoice invoice = Invoice.create(params);
+        assertEquals(invoice.getBilling(), "send_invoice");
+        assertEquals(invoice.getDueDate(), dueDate);
```
pantera-stripe: Assertion order reversed on both of these assertions. Should be `assertEquals(expect, actual)` http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(java.lang.Object, java.lang.Object)

src/test/java/com/stripe/functional/SubscriptionTest.java
```diff
@@ -192,4 +192,26 @@ public void testSubscriptionMetadata() throws StripeException {
...
+        Map<String, Object> subscriptionParams = new HashMap<String, Object>();
+        subscriptionParams.put("plan", plan.getId());
+        subscriptionParams.put("billing", "send_invoice");
+        subscriptionParams.put("days_until_due", 30);
+        subscriptionParams.put("customer", customer.getId());
+        Subscription sub = Subscription.create(subscriptionParams);
+        assertEquals(sub.getPlan().getId(), plan.getId());
+        assertEquals(sub.getBilling(), "send_invoice");
+        assertEquals(sub.getDaysUntilDue(), (Integer) 30);
+
```
pantera-stripe: Assertion orders reversed again.

src/test/java/com/stripe/functional/SubscriptionTest.java
```diff
@@ -192,4 +192,26 @@ public void testSubscriptionMetadata() throws StripeException {
...
+        Map<String, Object> updateParams = new HashMap<String, Object>();
+        updateParams.put("days_until_due", 10);
+        Subscription subUpdated = sub.update(updateParams);
+        assertEquals(subUpdated.getDaysUntilDue(), (Integer) 10);
```
pantera-stripe: Same.

pantera-stripe:
ptal @jacqueline-stripe

src/main/java/com/stripe/model/Invoice.java
jacqueline-stripe: I believe the Java bindings only require you to add the fields on the model, not the parameters accepted to API calls. `days_until_due` is a parameter that can be passed on create, but is not a field on the invoice model.

pantera-stripe:
I'm Pantera, and I approve this PR.

jacqueline-stripe:
@brandur-stripe feel free to release as early as 2/22 (We will not be launching to public until March, but similar to Multiplan Subscriptions, I believe it should be ok to release the bindings for people to test out).

brandur-stripe:
Released as 3.11.0.

PR #349, opened 2017-02-21
Alphabetize resource list in deserializer

brandur-stripe:
Simply alphabetizes the resource list in the deserializer. This is
somewhat pedantic as far as changes go, but it's not amazing right now
in that resources essentially get added to arbitrary positions in the
list.

r? @ob-stripe


ob-stripe:
lgtm. I think it was my idea to use the order in which the objects are described in the API reference, but in retrospect, it was silly and alphabetical order is much cleaner.

brandur-stripe:
Thanks OB!

PR #353, opened 2017-03-10
Make charge dispute expandable

brandur-stripe:
Charge disputes are no longer expanded by default in newer API versions.
Here we make that an expandable field, and fix functional tests that
depend on it to leverage that.

This is a breaking change.

r? @ob-stripe


ob-stripe:
Too bad this is a breaking change :/

Code lgtm, but it looks like Travis builds are stuck for now.

brandur-stripe:
Thanks! Travis finally came through. Pulling this in.

PR #356, opened 2017-03-24
Outcomes on charges are now collapsed by default

### LINKED ISSUE #355: `rule`-attribute of ChargeOutcome is broken
API Version [2017-02-14](https://stripe.com/docs/upgrades#2017-02-14) changed the type of the '[rule](https://github.com/stripe/stripe-java/blob/dd9529f48388b26d08b4377335711b2ad0cfd875/src/main/java/com/stripe/model/ChargeOutcome.java#L9)'-attribute of `ChargeOutcome.java`.  Needs to either be replaced with a String-type or support expansion.  `ChargeOutcomeRule.java` may also need to be removed; depending on the outcome.

brandur-stripe:
This makes outcomes on charges an expandable field. We leave the old
`getOutcome` and `setOutcome` in place to try and maintain backwards
compatibility.

To compensate for the reserved names, I introduce `getOutcomeID` and
`setOutcomeID` instead, which aren't really a convention right now (but
maybe should be?).

We should also consider just breaking the old version of the field.
That's up for discussion.

Fixes #355.

### DIFF
--- a/src/main/java/com/stripe/model/Charge.java
+++ b/src/main/java/com/stripe/model/Charge.java
@@ -33,7 +33,7 @@ public class Charge extends APIResource implements MetadataStore<Charge>, HasId 
 	ExpandableField<Invoice> invoice;
 	Boolean livemode;
 	Map<String, String> metadata;
-	ChargeOutcome outcome;
+	ExpandableField<ChargeOutcome> outcome;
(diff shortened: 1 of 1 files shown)


brandur-stripe:
I actually misunderstood the change while writing this, and this implementation is wrong. I'll send a new PR.

PR #357, opened 2017-03-26
Added support for the new Payout object

remi-stripe:
The Transfer object used to represent all movements of funds in Stripe. It split in two concepts:
- Transfer: this describes the movement of funds between Stripe accounts and is specific to Stripe Connect.
- Payout: this describes the movement of funds from a Stripe account to a bank account, debit card or any future payout method.

This change is behind an API version so old API versions would still use the Transfer object for everything while new API version would see the split.

This applies beyond the new object as some properties/methods are removed from Transfer and other properties are renamed on other objects.

### DIFF
--- a/src/main/java/com/stripe/model/EventDataDeserializer.java
+++ b/src/main/java/com/stripe/model/EventDataDeserializer.java
@@ -44,6 +44,7 @@ public class EventDataDeserializer implements JsonDeserializer<EventData> {
 		objectMap.put("order", Order.class);
 		objectMap.put("order_item", OrderItem.class);
 		objectMap.put("order_return", OrderReturn.class);
+		objectMap.put("payout", Payout.class);
 		objectMap.put("plan", Plan.class);
 		objectMap.put("product", Product.class);
 		objectMap.put("refund", Refund.class);
--- a/src/main/java/com/stripe/model/Payout.java
+++ b/src/main/java/com/stripe/model/Payout.java
@@ -1,0 +1,237 @@
+package com.stripe.model;
+
+import com.stripe.exception.APIConnectionException;
+import com.stripe.exception.APIException;
+import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.List;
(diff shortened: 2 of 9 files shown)


remi-stripe:
@brandur-stripe I have some open questions on this one too, hope it's okay!

1. Since we're adding a new class `Payout` do we still want to support old methods of passing the API key as a parameter or no API key at all? I've assumed so but wanted to ask.
2. A lot of properties/methods are removed from `Transfer` after that API version. Should we mark the get/set as deprecated for those objects?
3. Some properties are renamed such as `transferSchedule` which becomes `payoutSchedule`. This was associated with the class `AccountTransferSchedule` in the library. Do we need to duplicate it?

brandur-stripe:
@remi-stripe 

> Since we're adding a new class Payout do we still want to support old methods of passing the API key as a parameter or no API key at all? I've assumed so but wanted to ask.

Nah, lets kill those things. `RequestOptions` are the way to go now.

> A lot of properties/methods are removed from Transfer after that API version. Should we mark the get/set as deprecated for those objects?

> Some properties are renamed such as transferSchedule which becomes payoutSchedule. This was associated with the class AccountTransferSchedule in the library. Do we need to duplicate it?

brandur-stripe:
@remi-stripe Minus the few things discussed above, this looks great! Looking at your test results, I think your tests aren't working quite right. There are some instructions on how to verify them locally in the README if it helps!

remi-stripe:
@brandur-stripe  Will try and fix tomorrow. I think part of the tests don't work because they rely on the new API version being released. I did test those locally and they worked as far as I can remember

remi-stripe:
@brandur-stripe When I run the test suite locally I  get 
```
Results :

Tests run: 307, Failures: 0, Errors: 0, Skipped: 0
```

I added the `@Deprecated` flag to the getter/setter on `Transfer` and removed the method from Payout that pass the API key explicitly.

PTAL

brandur-stripe:
Looks great. Thanks Remi.

brandur-stripe:
Remi, I rebased your branch, added c4ce2f5 and force pushed to see if we can get the test suite running on the newly released API version. If you want to make further modifications you'll need to do something like git fetch origin <branch> && git reset --hard origin/<branch>.

brandur-stripe:
Released as 4.2.0.

PR #358, opened 2017-03-30
Make `rule` under charge outcome expandable

### LINKED ISSUE #355: `rule`-attribute of ChargeOutcome is broken

brandur-stripe:
Adds accessors for an expandable `rule` under charge outcome now that
this field is now collapsed by default. We also start assigning
`ChargeOutcomeRule` and ID with `HasId`.

I changed the tests around a little bit to make them a little more
consistent with how we test expansions elsewhere in the suite.

Fixes #355.

r? @anelder-stripe

### DIFF
--- a/src/main/java/com/stripe/model/ChargeOutcome.java
+++ b/src/main/java/com/stripe/model/ChargeOutcome.java
@@ -1,12 +1,14 @@
 package com.stripe.model;
 
-public class ChargeOutcome extends StripeObject {
+import com.stripe.net.APIResource;
+
+public class ChargeOutcome extends APIResource {
 	protected String networkStatus;
 	protected String reason;
 	protected String riskLevel;
+	protected ExpandableField<ChargeOutcomeRule> rule;
 	protected String sellerMessage;
 	protected String type;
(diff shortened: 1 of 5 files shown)


ob-stripe:
lgtm!

Do you think we should deprecate `getX` and favor `getXId` / `getXObject` for all expandable fields?

brandur-stripe:
Thanks @ob-stripe!

> Do you think we should deprecate getX and favor getXId / getXObject for all expandable fields?

I'd personally say "yes" just because we'll be able to roll out changes slowly and keep everything forward compatible. It also kind of just makes sense anyway.

PR #360, opened 2017-04-05
Peg the test suite's API version

brandur-stripe:
Pegs the test suite's API version so that we can upgrade it
independently of the account's implicit version. It's also to check how
much breaks when we move to a relatively modern version of the API.

See also https://github.com/stripe/stripe-php/pull/337 and https://github.com/stripe/stripe-python/pull/298.

### DIFF
--- a/src/test/java/com/stripe/BaseStripeFunctionalTest.java
+++ b/src/test/java/com/stripe/BaseStripeFunctionalTest.java
@@ -72,8 +72,13 @@ public class BaseStripeFunctionalTest {
     @BeforeClass
     public static void setUp() {
         Stripe.apiKey = "tGN0bIwXnHdwOa85VABjPdSn8nWY7G7I"; // stripe public
+
+        // Peg the API version so that it can be varied independently of the
+        // one set on the test account.
+        Stripe.apiVersion = "2017-02-14";
+
         // test key
-        supportedRequestOptions = RequestOptions.builder().setStripeVersion("2015-02-16").build();
+        supportedRequestOptions = RequestOptions.builder().build();
 
         defaultCardParams.put("number", "4242424242424242");
         defaultCardParams.put("exp_month", 12);
--- a/src/test/java/com/stripe/BaseStripeFunctionalTest.java
+++ b/src/test/java/com/stripe/BaseStripeFunctionalTest.java
@@ -78,7 +78,7 @@ public class BaseStripeFunctionalTest {
         Stripe.apiVersion = "2017-02-14";
 
         // test key
-        supportedRequestOptions = RequestOptions.builder().build();
+        supportedRequestOptions = RequestOptions.builder().setStripeVersion(Stripe.apiVersion).build();
 
         defaultCardParams.put("number", "4242424242424242");
         defaultCardParams.put("exp_month", 12);
--- a/src/test/java/com/stripe/functional/ChargeTest.java
+++ b/src/test/java/com/stripe/functional/ChargeTest.java
@@ -1,6 +1,7 @@
 package com.stripe.functional;
 
 import com.google.common.collect.ImmutableMap;
+import static org.hamcrest.CoreMatchers.instanceOf;
 import com.stripe.Stripe;
 import com.stripe.exception.CardException;
 import com.stripe.exception.InvalidRequestException;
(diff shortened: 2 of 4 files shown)


src/test/java/com/stripe/functional/DisputeTest.java
```diff
@@ -205,22 +205,4 @@ public void testSubmitEvidence() throws StripeException, InterruptedException {
...
-        Dispute initialDispute = disputedCharge.getDisputeObject();
-        assertNull(initialDispute.getEvidence());
-        assertNull(initialDispute.getEvidenceSubObject());
-        Map<String, Object> disputeParams = ImmutableMap.<String, Object>of("evidence", myEvidence);
-
-        Dispute updatedDispute = disputedCharge.updateDispute(disputeParams);
-        assertNotNull(updatedDispute);
-        assertEquals(myEvidence, updatedDispute.getEvidence());
-        assertNull(updatedDispute.getEvidenceSubObject());
-    }
```
brandur-stripe: I got rid of this test. I think it's been long enough.

brandur-stripe:
Looks like 8 tests that need fixing:

```
Tests in error: 
  testCustomerCreate(com.stripe.functional.CustomerTest)
  testCustomerCardAddition(com.stripe.functional.CustomerTest)
  testCreateCardThroughCollection(com.stripe.functional.CustomerTest)
  testCustomerCardUpdate(com.stripe.functional.CustomerTest)
  testCustomerCardDelete(com.stripe.functional.CustomerTest)
  testCustomerCreatePerCallAPIKey(com.stripe.functional.CustomerTest)
  testInvalidAddressZipTest(com.stripe.functional.ChargeTest)
  testInvalidAddressLine1Test(com.stripe.functional.ChargeTest)
```

src/test/java/com/stripe/BaseStripeFunctionalTest.java
```diff
@@ -72,8 +72,13 @@ public static Customer createDefaultCustomerWithPlan(Plan plan)
...
     public static void setUp() {
         Stripe.apiKey = "tGN0bIwXnHdwOa85VABjPdSn8nWY7G7I"; // stripe public
+
+        // Peg the API version so that it can be varied independently of the
+        // one set on the test account.
+        Stripe.apiVersion = "2017-02-14";
+
         // test key
-        supportedRequestOptions = RequestOptions.builder().setStripeVersion("2015-02-16").build();
+        supportedRequestOptions = RequestOptions.builder().setStripeVersion(Stripe.apiVersion).build();
```
brandur-stripe: I'm honestly not 100% sure what the intention of this variable was. I _think_ that it was maybe to assemble some basic options and test it against the occasional request. It seems a little superfluous though TBH, and we may want to strip it out on another PR at some point.

src/test/java/com/stripe/functional/ChargeTest.java
```diff
@@ -170,8 +171,11 @@ public void testInvalidAddressZipTest() throws StripeException {
 		invalidChargeParams.put("card", invalidCardParams);
 		Charge charge = Charge.create(invalidChargeParams, supportedRequestOptions);
 		assertEquals(charge.getPaid(), true);
-		assertEquals(charge.getCard().getAddressZipCheck(), "fail");
-		assertEquals(charge.getCard().getAddressLine1Check(), "pass");
+
+		assertThat(charge.getSource(), instanceOf(Card.class));
+		Card card = (Card)charge.getSource();
```
ob-stripe: Much cleaner! 👍

src/test/java/com/stripe/functional/DisputeTest.java
```diff
@@ -208,17 +209,19 @@ public void testSubmitEvidence() throws StripeException, InterruptedException {
 
     @Test
     public void testSubmitOldStyleEvidence() throws StripeException, InterruptedException {
+		RequestOptions options = RequestOptions.builder().setStripeVersion("2014-11-20").build();
```
ob-stripe: Looks like you're using tabs here when the rest of the file uses spaces (though the file probably should be using tabs...)

brandur-stripe: Oh man. I really need a linter or something. Fixed in d6e8cdd.

ob-stripe:
It's not really related to the PR, but I noticed that one of the tests [failed](https://travis-ci.org/stripe/stripe-java/jobs/219400082#L1415-L1417) with a `NullPointerException` [here](https://github.com/stripe/stripe-java/blob/brandur-fixed-api-version/src/test/java/com/stripe/functional/ChargeTest.java#L220). My guess is that something happened on the API's end, but we probably should assert that `c.getBalanceTransactionObject()` is non-null to get a proper assertion failure instead of an exception.

Other than that, lgtm! Nice technical debt payoff :)

brandur-stripe:
> It's not really related to the PR, but I noticed that one of the tests failed with a NullPointerException here. My guess is that something happened on the API's end, but we probably should assert that c.getBalanceTransactionObject() is non-null to get a proper assertion failure instead of an exception.

Yeah totally. I think even all the tests were failing with the equivalent of a null reference exception because they were trying to do things like `charge.getCards().getData().get(0)`. `getCards` had become `getSources` and was throwing an exception as you tried to change the accessors.

The problem's going to be widespread enough that it's going to take a fair bit of yak shaving to fully address ... we probably need some kind of assertion helper in there somewhere.

Anyway, thanks for taking a look!

PR #365, opened 2017-04-12
Balance transaction deserialization with support for expandable source objects

kion:
# The Task

Implement balance transactions view with metadata (e.g. customer name, payment summary) rendered for records of type "charge". To make it clear, `Charge` objects we create are populated with metadata we need to store. This task is only about extracting that metadata from balance transaction records corresponding to charges.

Note: I'm aware balance transactions do not support metadata, but according to API docs, "source" object on balance transaction is expandable and is capable of returning full nested structure of corresponding charge, transfer, etc.

# The Problem

If "data.source" expansion is requested, returned JSON data structure looks correct - it does contain nested objects.

However, stripe-java unfortunately does not support this currently.

Consider the following API call:

`Map<String, Object> params = new HashMap<String, Object>();`
`params.put("expand", Arrays.asList(new String[]{ "data.source.metadata" }));`
`BalanceTransactionCollection transactions = BalanceTransaction.list(params);`

It worked perfectly without "expand" param set (but obviously didn't contain the nested objects we needed to get), but as long as param is set, it fails with the following exception:

> com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 15 column 18 path $.data[0].source

This happens when execution flow reaches the following line within `LiveStripeResponseGetter._request` method:

`T resource = APIResource.GSON.fromJson(rBody, clazz);`

Obviously Gson fails to properly deserialize balance transaction JSON structure that has nested source object, b/c `BalanceTransaction.source` is defined as `String`. 

To make it clear, I'm aware that stripe-java has support for expandable fields, but the challenge there is that just changing "source" type on `BalanceTransaction` from `String` to `ExpandableField<...>` wouldn't work b/c nested "source" objects within returned JSON might have *different types* (i.e. some are charges, others are transfers, adjustments, etc), while `ExpandableField` requires single generic type (`<? extends HasId>`).

# The Solution

Note: I tried to solve this in a different way initially - see pull request https://github.com/stripe/stripe-java/pull/361 - , which worked just fine with JSON structures returned by end-point requests, but was incompatible with incoming JSON structures for webhooks (was throwing exception on `APIResource.GSON.fromJson(jsonBody, Event.class)` call); this new solution supports both and looks a bit more elegant too.

Here's a list of changes I made to solve the problem:

- added a bunch of transient properties for various possible nested "source" object types (e.g. “sourceCharge", “sourceTransfer", etc.) to `BalanceTransaction` class (I only added a few, might need to support more in the future) which would be populated based on the type of corresponding "source" object (others would remain null); so, for example, if nested "source" object was of type "charge", you'd be able to call `balanceTransaction.getSourceCharge()` to get populated `Charge` instance
- implemented new `BalanceTransactionDeserializer` class that supports nested "source" objects deserialization for balance transactions and registered it within `APIResource.GSON` (i.e. mapped it to `BalanceTransaction` class via `GsonBuilder`)

I only needed support for nested "charge" objects, but also added support for "transfer" and "refund" ones - for the sake of having an example of how to extend that logic to support more nested object types, if needed.

**NOTE**: I did these changes on separate branch - **balance-transaction-deserializer** - which is based on v3.11.0 of code-base b/c that's the version we use in production currently (i.e. version our code-base is currently integrated with); changes are pretty straightforward though and shouldn't be hard to merge into master branch (GitHub indicates there are no conflicts).

**IMPORTANT**: as far as I can tell, these are non-breaking changes that should be safe to merge in - they should not break any existing code that uses stripe-java, i.e. if someone calls `balanceTransaction.getSource()` in their code, they would still be getting String ID, while those who need full nested source object would rather call one of the new methods for the appropriate "source" object type, e.g. `balanceTransaction.getSourceCharge()` to get nested `Charge` object; `BalanceTransactionDeserializer ` only handles nested "source" object for balance transaction if one is present and only handles types it knows how to handle - in all other cases it does nothing besides of setting "source" object ID (which it always does, of course), so it won't throw exceptions in case no/new/unknown "source" objects are present in JSON data returned by API end-points; and, finally, just FYI: `BalanceTransactionTest` still passes without issues (I did not do any changes to it).

### DIFF
--- a/src/main/java/com/stripe/model/BalanceTransaction.java
+++ b/src/main/java/com/stripe/model/BalanceTransaction.java
@@ -1,5 +1,8 @@
 package com.stripe.model;
 
+import java.util.List;
+import java.util.Map;
+
 import com.stripe.Stripe;
 import com.stripe.exception.APIConnectionException;
 import com.stripe.exception.APIException;
@@ -8,9 +11,6 @@ import com.stripe.exception.CardException;
 import com.stripe.exception.InvalidRequestException;
 import com.stripe.net.APIResource;
 import com.stripe.net.RequestOptions;
-
-import java.util.List;
-import java.util.Map;
 
 public class BalanceTransaction extends APIResource implements HasId {
 	String id;
@@ -26,7 +26,11 @@ public class BalanceTransaction extends APIResource implements HasId {
 	String source;
 	String status;
 	String type;
-
+	
+	transient Charge sourceCharge;
+	transient Transfer sourceTransfer;
+	transient Refund sourceRefund;
+	
 	@Deprecated
 	TransferCollection sourcedTransfers;
 
@@ -146,6 +150,30 @@ public class BalanceTransaction extends APIResource implements HasId {
 	public void setType(String type) {
 		this.type = type;
 	}
+	
+	public Charge getSourceCharge() {
+		return sourceCharge;
+	}
+
+	public void setSourceCharge(Charge sourceCharge) {
+		this.sourceCharge = sourceCharge;
+	}
+
+	public Transfer getSourceTransfer() {
+		return sourceTransfer;
+	}
+
+	public void setSourceTransfer(Transfer sourceTransfer) {
+		this.sourceTransfer = sourceTransfer;
+	}
+
+	public Refund getSourceRefund() {
+		return sourceRefund;
+	}
+
+	public void setSourceRefund(Refund sourceRefund) {
+		this.sourceRefund = sourceRefund;
+	}
 
 	public static BalanceTransaction retrieve(String id) throws AuthenticationException,
 			InvalidRequestException, APIConnectionException, CardException,
@@ -200,4 +228,5 @@ public class BalanceTransaction extends APIResource implements HasId {
 			APIConnectionException, CardException, APIException {
 		return list(params, options);
 	}
+	
 }
--- a/src/main/java/com/stripe/model/BalanceTransaction.java
+++ b/src/main/java/com/stripe/model/BalanceTransaction.java
@@ -27,9 +27,7 @@ public class BalanceTransaction extends APIResource implements HasId {
 	String status;
 	String type;
 	
-	transient Charge sourceCharge;
-	transient Transfer sourceTransfer;
-	transient Refund sourceRefund;
+	transient HasId sourceObject;
 	
 	@Deprecated
 	TransferCollection sourcedTransfers;
@@ -151,30 +149,15 @@ public class BalanceTransaction extends APIResource implements HasId {
 		this.type = type;
 	}
 	
-	public Charge getSourceCharge() {
-		return sourceCharge;
+	public void setSourceObject(HasId sourceObject) {
+		this.sourceObject = sourceObject;
 	}
 
-	public void setSourceCharge(Charge sourceCharge) {
-		this.sourceCharge = sourceCharge;
+	@SuppressWarnings("unchecked")
+	public <O> O getSourceObject() {
+		return (O) sourceObject;
 	}
-
-	public Transfer getSourceTransfer() {
-		return sourceTransfer;
-	}
-
-	public void setSourceTransfer(Transfer sourceTransfer) {
-		this.sourceTransfer = sourceTransfer;
-	}
-
-	public Refund getSourceRefund() {
-		return sourceRefund;
-	}
-
-	public void setSourceRefund(Refund sourceRefund) {
-		this.sourceRefund = sourceRefund;
-	}
-
+		
 	public static BalanceTransaction retrieve(String id) throws AuthenticationException,
 			InvalidRequestException, APIConnectionException, CardException,
 			APIException {
--- a/src/main/java/com/stripe/model/BalanceTransaction.java
+++ b/src/main/java/com/stripe/model/BalanceTransaction.java
@@ -23,11 +23,9 @@ public class BalanceTransaction extends APIResource implements HasId {
 	Long fee;
 	List<Fee> feeDetails;
 	Integer net;
-	String source;
+	ExpandableField<HasId> source;
 	String status;
 	String type;
-	
-	transient HasId sourceObject;
 	
 	@Deprecated
 	TransferCollection sourcedTransfers;
@@ -112,14 +110,6 @@ public class BalanceTransaction extends APIResource implements HasId {
 		this.net = net;
 	}
 
-	public String getSource() {
-		return source;
-	}
-
-	public void setSource(String source) {
-		this.source = source;
-	}
-
 	/**
 	 * @deprecated
 	 * Recent API versions no longer return this field (https://stripe.com/docs/upgrades#2017-01-27).
@@ -149,15 +139,35 @@ public class BalanceTransaction extends APIResource implements HasId {
 		this.type = type;
 	}
 	
-	public void setSourceObject(HasId sourceObject) {
-		this.sourceObject = sourceObject;
+	public String getSource() {
+		if (this.source == null) {
+			return null;
+		}
+		return this.source.getId();
 	}
 
-	@SuppressWarnings("unchecked")
-	public <O> O getSourceObject() {
-		return (O) sourceObject;
+	public void setSource(String sourceID) {
+		this.source = setExpandableFieldID(sourceID, this.source);
 	}
-		
+	
+	public HasId getSourceObject() {
+		if (this.source == null) {
+			return null;
+		}
+		return this.source.getExpanded();
+	}
+
+	public void setSourceObject(HasId o) {
+		this.source = new ExpandableField<HasId>(o.getId(), o);
+	}
+
+	public <O extends HasId> O getSourceObjectAs() {
+		if (this.source == null) {
+			return null;
+		}
+		return (O) this.source.getExpanded();
+ 	}
+	
 	public static BalanceTransaction retrieve(String id) throws AuthenticationException,
 			InvalidRequestException, APIConnectionException, CardException,
 			APIException {
--- a/src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
+++ b/src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
@@ -50,17 +50,19 @@ public class BalanceTransactionDeserializer implements JsonDeserializer<BalanceT
 			sourceId = sourceIdEl != null ? sourceIdEl.getAsString() : null;
 			JsonElement val = sourceJsonObject.get("object");
 			if (val != null) {
+				Class<? extends HasId> sourceObjClass = null;
 				if ("charge".equals(val.getAsString())) {
-					Charge charge = context.deserialize(source, Charge.class);
-					balanceTransaction.setSourceCharge(charge);
+					sourceObjClass = Charge.class;
 				} else if ("transfer".equals(val.getAsString())) {
-					Transfer transfer = context.deserialize(source, Transfer.class);
-					balanceTransaction.setSourceTransfer(transfer);
+					sourceObjClass = Transfer.class;
 				} else if ("refund".equals(val.getAsString())) {
-					Refund refund = context.deserialize(source, Refund.class);
-					balanceTransaction.setSourceRefund(refund);
+					sourceObjClass = Refund.class;
 				}
 				// TODO support other source types (?)
+				if (sourceObjClass != null) {
+					HasId sourceObj = context.deserialize(source, sourceObjClass);
+					balanceTransaction.setSourceObject(sourceObj);
+				}
 			}
 		} else if (!source.isJsonNull()) {
 			throw new JsonParseException("Source field on a balance transaction was a non-primitive, non-object type.");
--- a/src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
+++ b/src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2003-2010 SPORTSVITE, LLC. All Rights Reserved.
+ *
+ * This software is the confidential and proprietary information of
+ * SPORTSVITE, LLC and certain third parties ("Confidential Information").
+ * You shall not disclose such Confidential Information and shall
+ * use it only in accordance with the terms of the license agreement
+ * you entered into with SPORTSVITE, LLC.
+ *
+ * SPORTSVITE, LLC MAKES NO REPRESENTATIONS OR WARRANTIES ABOUT THE SUITABILITY OF
+ * THE SOFTWARE, EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE,
+ * OR NON-INFRINGEMENT. SPORTSVITE, LLC SHALL NOT BE LIABLE FOR ANY DAMAGES
+ * SUFFERED BY LICENSEE AS A RESULT OF USING, MODIFYING OR DISTRIBUTING
+ * THIS SOFTWARE OR ITS DERIVATIVES.
+ */
+package com.stripe.model;
+
+import java.lang.reflect.Type;
+
+import com.google.gson.FieldNamingPolicy;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonPrimitive;
+
+/**
+ * @author kion
+ */
+public class BalanceTransactionDeserializer implements JsonDeserializer<BalanceTransaction> {
+
+	@Override
+	public BalanceTransaction deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+		Gson gson = new GsonBuilder()
+				.setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+				.create();
+		
+		if (json.isJsonNull()) {
+			return null;
+		}
+
+		if (!json.isJsonObject()) {
+			throw new JsonParseException("BalanceTransaction type was not an object, which is problematic.");
+		}
+		
+		JsonObject btAsJsonObject = json.getAsJsonObject();
+
+		JsonElement source = btAsJsonObject.get("source");
+		
+		btAsJsonObject.remove("source");
+
+		BalanceTransaction balanceTransaction = gson.fromJson(json, typeOfT);
+		
+		String sourceId = null;
+
+		if (source.isJsonPrimitive()) {
+			JsonPrimitive sourceJsonPrimitive = source.getAsJsonPrimitive();
+			if (!sourceJsonPrimitive.isString()) {
+				throw new JsonParseException("Source field on a balance transaction was a primitive non-string type.");
+			}
+			sourceId = sourceJsonPrimitive.getAsString();
+		} else if (source.isJsonObject()) {
+			JsonObject sourceJsonObject = source.getAsJsonObject();
+			JsonElement sourceIdEl = sourceJsonObject.get("id");
+			sourceId = sourceIdEl != null ? sourceIdEl.getAsString() : null;
+			JsonElement val = sourceJsonObject.get("object");
+			if (val != null) {
+				if ("charge".equals(val.getAsString())) {
+					Charge charge = context.deserialize(source, Charge.class);
+					balanceTransaction.setSourceCharge(charge);
+				} else if ("transfer".equals(val.getAsString())) {
+					Transfer transfer = context.deserialize(source, Transfer.class);
+					balanceTransaction.setSourceTransfer(transfer);
+				} else if ("refund".equals(val.getAsString())) {
+					Refund refund = context.deserialize(source, Refund.class);
+					balanceTransaction.setSourceRefund(refund);
+				}
+				// TODO support other source types (?)
+			}
+		} else if (!source.isJsonNull()) {
+			throw new JsonParseException("Source field on a balance transaction was a non-primitive, non-object type.");
+		}
+		
+		balanceTransaction.setSource(sourceId);
+
+		return balanceTransaction;
+	}
+
+}
--- a/src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
+++ b/src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
@@ -1,19 +1,3 @@
-/*
- * Copyright (c) 2003-2010 SPORTSVITE, LLC. All Rights Reserved.
- *
- * This software is the confidential and proprietary information of
- * SPORTSVITE, LLC and certain third parties ("Confidential Information").
- * You shall not disclose such Confidential Information and shall
- * use it only in accordance with the terms of the license agreement
- * you entered into with SPORTSVITE, LLC.
- *
- * SPORTSVITE, LLC MAKES NO REPRESENTATIONS OR WARRANTIES ABOUT THE SUITABILITY OF
- * THE SOFTWARE, EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE,
- * OR NON-INFRINGEMENT. SPORTSVITE, LLC SHALL NOT BE LIABLE FOR ANY DAMAGES
- * SUFFERED BY LICENSEE AS A RESULT OF USING, MODIFYING OR DISTRIBUTING
- * THIS SOFTWARE OR ITS DERIVATIVES.
- */
 package com.stripe.model;
 
 import java.lang.reflect.Type;
@@ -28,9 +12,6 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 import com.google.gson.JsonPrimitive;
 
-/**
- * @author kion
- */
 public class BalanceTransactionDeserializer implements JsonDeserializer<BalanceTransaction> {
 
 	@Override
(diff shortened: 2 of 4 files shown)


brandur-stripe:
Hi @kion, thanks again for the continued attention on this problem.

So I think your solution is totally workable. I was going through the code though, and noticed that we do have some precedent this type of polymorphism in the instance of an event's `data`/`object` field [1].

The way this has been previously implemented is to store a basic `StripeObject` on the class like this [2]:

``` java
	public StripeObject getObject() {
		return object;
	}

	public void setObject(StripeObject object) {
		this.object = object;
	}
```

Which can then be used through a typecast like:

``` java
		Account account = (com.stripe.model.Account)event.getData().getObject();
```

The way this is implemented is through a custom deserializer similar to what you've already done. First, a static map gets built [3], then it's used to deserialize a named class [4].

I think you could make an argument as to which approach is overall better, type-specific fields like you've done versus cast-based polymorphism like in event, but I'm kind of thinking that we should stick to the latter just because the library already has precedent for it. Thoughts?

[1] https://stripe.com/docs/api/java#event_object-data-object
[2] https://github.com/stripe/stripe-java/blob/a389f6d7173d735d18692d4fc4314b12e5392ee3/src/main/java/com/stripe/model/EventData.java#L9,L15
[3] https://github.com/stripe/stripe-java/blob/a389f6d7173d735d18692d4fc4314b12e5392ee3/src/main/java/com/stripe/model/EventDataDeserializer.java#L19,L62
[4] https://github.com/stripe/stripe-java/blob/a389f6d7173d735d18692d4fc4314b12e5392ee3/src/main/java/com/stripe/model/EventDataDeserializer.java#L128,L133

kion:
Hi @brandur-stripe, thanks for looking into this and coming back to me.

So, while I think your reasoning makes sense and code base consistency indeed matters a lot, let me outline a couple of arguments:

* type casting isn't the best practice in general and while it was a necessary evil in earlier ages of Java, since generics were added to the language spec, it's generally advised to avoid type casting whenever possible; this case is a bit different though, b/c source object could be of a different type in each specific case...

So, if you'd insist on taking single type-agnostic (i.e. plain Object) property approach, I'd recommend the following generics based tweak for the getter method to avoid explicit type casting within code bases using stripe-java.

I'm basically saying that if type casting needs to be used at all, let's rather make it a generics based and have it in one spot only.

So, `BalanceTransaction` would then have the following:

```java
transient Object sourceObject;

public void setSourceObject(Object sourceObject) {
  this.sourceObject = sourceObject;
}

public <O> O getSourceObject() {
  return (O) sourceObject;
}
```

(`BalanceTransactionDeserializer` would only need a minor tweak to use this single setter instead of one of the type-specific ones)

Just to make it clear, without this you'd need to use code like the following one within code base that uses stripe-java:

```java
Charge charge = null;
Transfer transfer = null;
Refund refund = null;

if (transaction.getSourceObject() instanceof Charge) {
  charge = (Charge) transaction.getSourceObject();
} else if (transaction.getSourceObject() instanceof Charge) {
  transfer = (Transfer) transaction.getSourceObject();
} else if (transaction.getSourceObject() instanceof Charge) {
  refund = (Refund) transaction.getSourceObject();
}
```

With generics based approach, it would rather look like this:

```java
Charge charge = null;
Transfer transfer = null;
Refund refund = null;

if (transaction.getSourceObject() instanceof Charge) {
  charge = transaction.getSourceObject();
} else if (transaction.getSourceObject() instanceof Charge) {
  transfer = transaction.getSourceObject();
} else if (transaction.getSourceObject() instanceof Charge) {
  refund = transaction.getSourceObject();
}
```

Or, if you knew what exact type of course object was, you'd be able to simply do this:

`Charge charge = transaction.getSourceObject();`

or this:

`Transfer transfer = transaction.getSourceObject();`

etc.

This is still not perfect though b/c you need to do `instanceof` checks, which, similar to type casting, isn't a recommended practice.

So, just FYI, I actually did consider this approach initially, but decided to go with type-specific getters, b/c it would (IMO) make it a bit easier to write/read/maintain code that needs to use stripe-java. Consider how code for the examples above turns into more straightforward form with implementation I provided initially:

```java
Charge charge = null;
Transfer transfer = null;
Refund refund = null;

if (transaction.getSourceCharge() != null) {
  charge = transaction.getSourceCharge();
} else if (transaction.getSourceCharge() != null) {
  transfer = transaction.getSourceTransfer();
} else if (transaction.getSourceCharge() != null) {
  refund = transaction.getSourceRefund();
}
```

* now, no matter which approach we choose, it wouldn't be too hard to deal with related code in Java (and I'd agree if you argue code above doesn't really look all that different for either of the cases... which is true, but my point there is more about avoiding type casing and `instanceof` checks usage), but if you need to deal with `BalanceTransaction` entities in JSPs, let's say - it would become a bit more of an inconvenience; type checking wouldn't always be required there, but in case it's needed (for example, if I need to show metadata for balance transaction charge source objects and some other information for balance transaction transfer source objects in the same spot of the web page, e.g. same column within table view), it wouldn't look like an elegant code to write/maintain, IMO... so, instead of all those type checks or checks against type-specific properties (or even wrapping `BalanceTransaction` into separate POJO only to make it easier to handle it within view logic), I'd rather prefer simple code like this:

```jsp
<s:if test="%{#transaction.sourceCharge != null}">
  ${transaction.sourceCharge.metadata.get('summary')}
</s:if>
<s:elseif test="%{#transaction.sourceTransfer != null}">
  <!-- render transfer specific data -->
</s:elseif>
<s:elseif test="%{#transaction.sourceRefund != null}">
  <!-- render refund specific data -->
</s:elseif>
```

That being said, if you still think single type-agnostic property approach would fit existing stripe-java code base better, I'd be fine with that too (even though I personally think it'd be more convenient to have multiple type-specific getters for the reasons I outlined above).

In fact, we could even have both :) So, if `getSourceObject()` needs to be there for the code base consistency sake (and maybe some other reasons?), so be it. But why not also add some convenience methods (e.g. `getSourceCharge()`, etc.) to make it more flexible and easier to use? These methods would simply do what I outlined above, but it would still be better to have it all in one place (i.e. as convenience methods within `BalanceTransaction` class) instead of having it all over the place within code bases that use stripe-java.

Either way, it wouldn't be hard to make these tweaks, so let me know what you think would be the best approach and I'll tweak my initial changes accordingly.

brandur-stripe:
Hi @kion, thanks as usual for the thoughtful response.

So, yes, I'd totally be on board for something like a generic-based version of what I suggested above along-side the one without. I'd suggest that we might even want to go as far as introducing a `BalanceTransactionSourceInterface` and assign it to the relevant objects. That way you could plausibly just always retrieve an object with that, and do an operation on a common field (to be clear though, maybe `BalanceTransactionSourceInterface` could inherit `HasId`, but it wouldn't be necessary for not to go through and look for all common fields).

I think a possible disadvantage of adding the `getSourceCharge`-style helpers is that it's more likely that a catch all `else` statement will be needed. If I look at this:

``` java
if (transaction.getSourceObject() instanceof Charge) {
  charge = transaction.getSourceObject<Charge>();
} else if (transaction.getSourceObject() instanceof Refund) {
  refund = transaction.getSourceObject<Refund>();
} else if (transaction.getSourceObject() instanceof Transfer) {
  transfer = transaction.getSourceObject<Transfer>();
}
```

It's quite obvious to me that an `else` should be in there because it could be a different type.

On the other hand, this might be exhaustive already for all I know:

``` java
if (transaction.getSourceCharge() != null) {
  charge = transaction.getSourceCharge();
} else if (transaction.getSourceTransfer() != null) {
  transfer = transaction.getSourceTransfer();
} else if (transaction.getSourceRefund() != null) {
  refund = transaction.getSourceRefund();
}
```

In the case of BT's `source` this is especially important because it's allowed to be one of a dozen different types (along with the three you've listed, it can also be a fee, fee refund, or transfer reversal for example). It's fairly important that people include an `else` in case it comes up that they receive a resource type that they're not handling because we might even add more to the list of possibilities over time.

kion:
Hi @brandur-stripe,

Thanks for spending more time on this and coming back to me.

> It's quite obvious to me that an else should be in there because it could be a different type.

Not quite sure I get this one, perhaps b/c it looks more like a subjective thing to me (for the version I proposed, I think you could just as well argue the same: it's quite obvious that an else should be in there b/c it could be a different property that was populated). Nevertheless, I think both approaches are acceptable and both look quite obvious (I don't think anyone would have issues with either), my point was more about cleaner code and absense of type casting / instanceof.

But hey, it's your code base, so I guess you know better how things should look like there.

Everything else you said makes sense to me.

So, I'll spend some more time tweaking it as you suggested / we agreed and will come back to you later.

kion:
@brandur-stripe: 

So, I've tweaked it as was requested.

Just FYI: I did NOT add `BalanceTransactionSourceInterface` just yet b/c this would imply some changes to the existing types like Charge, Transfer, etc, which would need to extend `BalanceTransactionSourceInterface` instead of `HasId`. In such a case though we'd really need to rather add that to *all* classes that could be nested within `BalanceTransaction`. I'm not sure I know what the requirements for this would be and as long as it's also not really clear at this point what the shared properties `BalanceTransactionSourceInterface` would hold, besides of object ID, I decided to simply use `HasId` interface instead (which those classes like Charge, Transfer, etc already implement). If shared properties need to be implemented in the future, it could indeed be updated to `BalanceTransactionSourceInterface` instead, but for now it seems to make more sense to keep it simple (until we have more specific requirements for what `BalanceTransactionSourceInterface` could look like). Makes sense?

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonPrimitive;
+
+public class BalanceTransactionDeserializer implements JsonDeserializer<BalanceTransaction> {
+
+	@Override
+	public BalanceTransaction deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+		Gson gson = new GsonBuilder()
+				.setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+				.create();
```
brandur-stripe: Is it possible to use `APIResource.GSON` instead of defining a new one here? It looks like this is what the `EventDataDeserializer` does.

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
```diff
@@ -0,0 +1,76 @@
...
+		
+		if (json.isJsonNull()) {
+			return null;
+		}
+
+		if (!json.isJsonObject()) {
+			throw new JsonParseException("BalanceTransaction type was not an object, which is problematic.");
+		}
```
brandur-stripe: Just in the interest of brevity if we can achieve it, do you know if either of these cases is ever likely to happen? I notice that some of the other deserializers have managed to get away without them.

src/main/java/com/stripe/model/BalanceTransaction.java
```diff
@@ -146,7 +148,16 @@ public String getType() {
...
 	}
+	
+	public void setSourceObject(HasId sourceObject) {
+		this.sourceObject = sourceObject;
+	}
 
+	@SuppressWarnings("unchecked")
+	public <O> O getSourceObject() {
+		return (O) sourceObject;
+	}
```
brandur-stripe: @kion We've fairly recently introduced a special type for expandable fields like this that I think we can potentially reuse here. What do you think about this change? (Should apply cleanly to your branch.)

``` patch
diff --git a/src/main/java/com/stripe/model/BalanceTransaction.java b/src/main/java/com/stripe/model/BalanceTransaction.java
index bd86330..79a504f 100644
--- a/src/main/java/com/stripe/model/BalanceTransaction.java
+++ b/src/main/java/com/stripe/model/BalanceTransaction.java
@@ -23,12 +23,10 @@ public class BalanceTransaction extends APIResource implements HasId {
 	Long fee;
 	List<Fee> feeDetails;
 	Integer net;
-	String source;
+	ExpandableField<HasId> source;
 	String status;
 	String type;
 	
-	transient HasId sourceObject;
-	
 	@Deprecated
 	TransferCollection sourcedTransfers;
 
@@ -112,14 +110,6 @@ public class BalanceTransaction extends APIResource implements HasId {
 		this.net = net;
 	}
 
-	public String getSource() {
-		return source;
-	}
-
-	public void setSource(String source) {
-		this.source = source;
-	}
-
 	/**
 	 * @deprecated
 	 * Recent API versions no longer return this field (https://stripe.com/docs/upgrades#2017-01-27).
@@ -148,14 +138,34 @@ public class BalanceTransaction extends APIResource implements HasId {
 	public void setType(String type) {
 		this.type = type;
 	}
-	
-	public void setSourceObject(HasId sourceObject) {
-		this.sourceObject = sourceObject;
+
+	public String getSource() {
+		if (this.source == null) {
+			return null;
+		}
+		return this.source.getId();
+	}
+
+	public void setSource(String sourceID) {
+		this.source = setExpandableFieldID(sourceID, this.source);
 	}
 
-	@SuppressWarnings("unchecked")
-	public <O> O getSourceObject() {
-		return (O) sourceObject;
+	public HasId getSourceObject() {
+		if (this.source == null) {
+			return null;
+		}
+		return this.source.getExpanded();
+	}
+
+	public void setSourceObject(HasId o) {
+		this.source = new ExpandableField<HasId>(o.getId(), o);
+	}
+
+	public <O extends HasId> O getSourceObjectAs() {
+		if (this.source == null) {
+			return null;
+		}
+		return (O) this.source.getExpanded();
 	}
 		
 	public static BalanceTransaction retrieve(String id) throws AuthenticationException,
```

Naming the generic accessor `getSourceObjectAs` gives us a plausible avenue to change `EventData` to have a similar method without having to introduce a backwards incompatible change.

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
```diff
@@ -0,0 +1,76 @@
...
+				if (sourceObjClass != null) {
+					HasId sourceObj = context.deserialize(source, sourceObjClass);
+					balanceTransaction.setSourceObject(sourceObj);
+				}
+			}
+		} else if (!source.isJsonNull()) {
+			throw new JsonParseException("Source field on a balance transaction was a non-primitive, non-object type.");
+		}
+		
+		balanceTransaction.setSource(sourceId);
```
brandur-stripe: I think moving over to `ExpandableField` as suggested above will allow us to avoid setting two fields in the case of an expanded object.

brandur-stripe:
> Not quite sure I get this one, perhaps b/c it looks more like a subjective thing to me (for the version I proposed, I think you could just as well argue the same: it's quite obvious that an else should be in there b/c it could be a different property that was populated). Nevertheless, I think both approaches are acceptable and both look quite obvious (I don't think anyone would have issues with either), my point was more about cleaner code and absense of type casting / instanceof.
>
> But hey, it's your code base, so I guess you know better how things should look like there.

Just to be clear, it definitely _is_ subjective :) I think my argument is somewhat plausible, but certainly not definitive. The more important aspect to me is to try and keep design consistency within the library's API if we possibly can.

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
kion: So, I actually tried to do it this way initially, but `BalanceTransactionTest` was failing with `java.lang.StackOverflowError` in that case... Not sure why exactly, but APIResource.GSON just doesn't work there, while new Gson instance does.

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
kion: So, I actually did it this way b/c I found that similar approach was used within code base earlier. I don't know if it's ever likely to happen. I think I would need to have a perfect knowledge of all data structures returned by API endpoints to be able to answer this. Assuming it's being used in some other spots on purposes and for a good reason, I'd rather keep it there.

src/main/java/com/stripe/model/BalanceTransaction.java
kion: Ha. So that's actually pretty much what my very first implementation was based on, remember? See https://github.com/stripe/stripe-java/pull/361 for more details.

The problem with it was that it's not compatible with webhooks b/c data structure that is submitted to webhook handler doesn't support expandable fields and it was failing b/c of that.

That's the reason I closed that pull request and created this one following the new approach. This one works with both API endpoints (i.e. when you request to expand balance transaction source object) and webhook handlers (i.e. when data structure that does not follow expandable source object format is submitted to webhook handler and you need to deserialize Event that contains nested balance transaction).

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
kion: See my comment above on why expandable fields would not work in this case.

kion:
Hi @brandur-stripe, thanks for reviewing the changes and coming back to me!

I left a reply to each of you comments above.

> The more important aspect to me is to try and keep design consistency within the library's API if we possibly can.

No worries at all. I totally understand that.

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
brandur-stripe: Alright, it seems like we're doing this in a few other places too, so that should be fine. Thanks!

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
brandur-stripe: Okay, can't hurt.

src/main/java/com/stripe/model/BalanceTransaction.java
brandur-stripe: Sorry the delayed reply here @kion! I'm doing a horrible job getting this through.

> The problem with it was that it's not compatible with webhooks b/c data structure that is submitted to webhook handler doesn't support expandable fields and it was failing b/c of that.

If you understand why this is, could you elaborate a little more?

I saw the error on your other PR, but it seems to me that this should be working. Basically the `ExpandableField` works [by checking whether a value is a primitive or an object](https://github.com/stripe/stripe-java/blob/0c6c358a670af7967edfe60a9441c67b5495d7da/src/main/java/com/stripe/model/ExpandableFieldDeserializer.java#L17,L39) and should be functional even for data that comes in via webhook. I think we should see if we can get whatever problem this is fixed before falling back to a different solution.

kion: Hi @brandur-stripe, 

So, I went ahead and double-checked it with the code changes you provided above.

It still fails with the same exception upon the following call:

```java
Event event = APIResource.GSON.fromJson(json, Event.class);
```

(where `json` is JSON structure coming in via webhook)

It throws the following exception:

```
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was STRING at path $.balance_transactions[0].source
```

It seems like stripe-java deserialization works in a way that would not handle objects within array (in this case - `balance_transactions` array property) same way non-array properties are handled.

Here’s JSON structure coming in via webhook that it fails to deserialize:

```json
{
  "id": "evt_1AF5AmKac2qaDcCZULVeSfjz",
  "object": "event",
  "api_version": "2017-02-14",
  "created": 1493747548,
  "data": {
    "object": {
      "id": "dp_1AF5AmKac2qaDcCZHvOjyrDo",
      "object": "dispute",
      "amount": 5000,
      "balance_transaction": "txn_1AF5AmKac2qaDcCZm8tBQt6I",
      "balance_transactions": [
        {
          "id": "txn_1AF5AmKac2qaDcCZm8tBQt6I",
          "object": "balance_transaction",
          "amount": -5000,
          "available_on": 1494288000,
          "created": 1493747548,
          "currency": "usd",
          "description": "Chargeback withdrawal for ch_1AF5AlKac2qaDcCZ4Iu07uIA",
          "fee": 1500,
          "fee_details": [
            {
              "amount": 1500,
              "application": null,
              "currency": "usd",
              "description": "Dispute fee",
              "type": "stripe_fee"
            }
          ],
          "net": -6500,
          "source": "dp_1AF5AmKac2qaDcCZHvOjyrDo",
          "status": "pending",
          "type": "adjustment"
        }
      ],
      "charge": "ch_1AF5AlKac2qaDcCZ4Iu07uIA",
      "created": 1493747548,
      "currency": "usd",
      "evidence": {
        "access_activity_log": null,
        "billing_address": "1400 Key Blvd #102, Arlington, VA, 22209, US",
        "cancellation_policy": null,
        "cancellation_policy_disclosure": null,
        "cancellation_rebuttal": null,
        "customer_communication": null,
        "customer_email_address": "asdf@asdfasdfasdfasdfasdf.com",
        "customer_name": "First Last",
        "customer_purchase_ip": "111.111.111.111",
        "customer_signature": null,
        "duplicate_charge_documentation": null,
        "duplicate_charge_explanation": null,
        "duplicate_charge_id": null,
        "product_description": null,
        "receipt": null,
        "refund_policy": null,
        "refund_policy_disclosure": null,
        "refund_refusal_explanation": null,
        "service_date": null,
        "service_documentation": null,
        "shipping_address": null,
        "shipping_carrier": null,
        "shipping_date": null,
        "shipping_documentation": null,
        "shipping_tracking_number": null,
        "uncategorized_file": null,
        "uncategorized_text": null
      },
      "evidence_details": {
        "due_by": 1495151999,
        "has_evidence": false,
        "past_due": false,
        "submission_count": 0
      },
      "is_charge_refundable": false,
      "livemode": false,
      "metadata": {
      },
      "reason": "general",
      "status": "needs_response"
    }
  },
  "livemode": false,
  "pending_webhooks": 1,
  "request": "req_AaFg0ircrTmUJ5",
  "type": "charge.dispute.created"
}
```

I don’t know why there’s both `balance_transaction` (String ID) and `balance_transactions` (array which contains that only balance transaction, but also includes the rest of the transaction properties) within data coming in via webhook, but in any case - it fails while trying to process the very first element of that array.

kion:
Hi @brandur-stripe, any update on this?

kion:
Hi @brandur-stripe, any progress on this?

I'll be talking to our CTO this week about our future plans regarding Stripe integration with our platform and it would be great to know whether changes in this PR are going to be merged into stripe-java master branch eventually.

As I mentioned before to some other guys from Stripe tech support, these changes are crucial for us to have, but applying them on top of each new version of stripe-java we upgrade to in the future isn't the route we'd like to take (we are always extra careful about upgrades like this, especially as we already have customers using our Stripe integration on production)... 

So, knowing whether these changes are going to be merged in would make it much easier for us to define our future strategy regarding stripe-java/API upgrades... I really hope we wouldn't end up using customized version of the older/outdated v3.11 lib forever :)

Any sort of estimation on this PR would be greatly appreciated!

brandur-stripe:
Hey @kion,

Once again, my apologies for the delay here. I wanted to set aside some time to dig into that issue with using `ExpandableField` for `source` under balance transactions.

So I should have seen this earlier, but I think I found the problem: because in each deserializer we're creating a new `GSON` instance that doesn't have support for `ExpandableField`. I was able to get the patch above working with your sample event using this patch:

``` patch
diff --git a/src/main/java/com/stripe/model/BalanceTransaction.java b/src/main/java/com/stripe/model/BalanceTransaction.java
index bd86330..79a504f 100644
--- a/src/main/java/com/stripe/model/BalanceTransaction.java
+++ b/src/main/java/com/stripe/model/BalanceTransaction.java
@@ -23,12 +23,10 @@ public class BalanceTransaction extends APIResource implements HasId {
 	Long fee;
 	List<Fee> feeDetails;
 	Integer net;
-	String source;
+	ExpandableField<HasId> source;
 	String status;
 	String type;
 	
-	transient HasId sourceObject;
-	
 	@Deprecated
 	TransferCollection sourcedTransfers;
 
@@ -112,14 +110,6 @@ public class BalanceTransaction extends APIResource implements HasId {
 		this.net = net;
 	}
 
-	public String getSource() {
-		return source;
-	}
-
-	public void setSource(String source) {
-		this.source = source;
-	}
-
 	/**
 	 * @deprecated
 	 * Recent API versions no longer return this field (https://stripe.com/docs/upgrades#2017-01-27).
@@ -148,14 +138,34 @@ public class BalanceTransaction extends APIResource implements HasId {
 	public void setType(String type) {
 		this.type = type;
 	}
-	
-	public void setSourceObject(HasId sourceObject) {
-		this.sourceObject = sourceObject;
+
+	public String getSource() {
+		if (this.source == null) {
+			return null;
+		}
+		return this.source.getId();
+	}
+
+	public void setSource(String sourceID) {
+		this.source = setExpandableFieldID(sourceID, this.source);
 	}
 
-	@SuppressWarnings("unchecked")
-	public <O> O getSourceObject() {
-		return (O) sourceObject;
+	public HasId getSourceObject() {
+		if (this.source == null) {
+			return null;
+		}
+		return this.source.getExpanded();
+	}
+
+	public void setSourceObject(HasId o) {
+		this.source = new ExpandableField<HasId>(o.getId(), o);
+	}
+
+	public <O extends HasId> O getSourceObjectAs() {
+		if (this.source == null) {
+			return null;
+		}
+		return (O) this.source.getExpanded();
 	}
 		
 	public static BalanceTransaction retrieve(String id) throws AuthenticationException,
diff --git a/src/main/java/com/stripe/model/DisputeDataDeserializer.java b/src/main/java/com/stripe/model/DisputeDataDeserializer.java
index 591dfb0..e0497da 100644
--- a/src/main/java/com/stripe/model/DisputeDataDeserializer.java
+++ b/src/main/java/com/stripe/model/DisputeDataDeserializer.java
@@ -16,6 +16,7 @@ public class DisputeDataDeserializer implements JsonDeserializer<Dispute> {
 	public Dispute deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
 		Gson gson = new GsonBuilder()
+				.registerTypeAdapter(ExpandableField.class, new ExpandableFieldDeserializer())
 				.create();
 		if (json.isJsonNull()) {
 			return null;
```

I think we should apply that to the PR you have here and get this merged in. What do you think?

kion:
Hi @brandur-stripe, I just tested your patch on our code-base and it seems to work just fine - thanks for spending time on looking into it!

So I went ahead and applied those changes to this PR's branch.

Hope this is now looking good and can be merged in?

brandur-stripe:
Yeah. I'm going to write one more test out of band, but this looks good. Thanks!

brandur-stripe:
Released as 5.2.0.

kion:
@brandur-stripe Great job, thank you!

PR #366, opened 2017-04-21
allow to set connection and read timeout per request.

agori:
It's important to be able to set a reasonable timeout for http connections. If production server has a network problem or Stripe serve is down, a reasonable timeout could prevent the calling thread from being blocked for huge amount of time. 

Timeout values were previously hard coded in LiveStripeResponseGetter, and the values were frankly quite high (30 seconds for connection and 80 for reads).

Example:

```
RequestOptions options = RequestOptions.builder()
    .setConnectionTimeout(2000)
    .setReadTimeout(5000)
    .build()

Customer.create(params, options);

```

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -149,7 +149,7 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
 		} else {
 			conn = (HttpURLConnection) stripeURL.openConnection();
 		}
-		conn.setConnectTimeout(options.getConnectionTimeout());
+		conn.setConnectTimeout(options.getConnectTimeout());
 		conn.setReadTimeout(options.getReadTimeout());
 		conn.setUseCaches(false);
 		for (Map.Entry<String, String> header : getHeaders(options).entrySet()) {
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -149,8 +149,8 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
 		} else {
 			conn = (HttpURLConnection) stripeURL.openConnection();
 		}
-		conn.setConnectTimeout(30 * 1000);
-		conn.setReadTimeout(80 * 1000);
+		conn.setConnectTimeout(options.getConnectionTimeout());
+		conn.setReadTimeout(options.getReadTimeout());
 		conn.setUseCaches(false);
 		for (Map.Entry<String, String> header : getHeaders(options).entrySet()) {
 			conn.setRequestProperty(header.getKey(), header.getValue());
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -3,8 +3,13 @@ package com.stripe.net;
 import com.stripe.Stripe;
 
 public class RequestOptions {
+	
+	private final static int DEFAULT_CONNECTION_TIMEOUT = 30 * 1000;    	
+	
+	private final static int DEFAULT_READ_TIMEOUT = 80 * 1000;
+    
 	public static RequestOptions getDefault() {
-		return new RequestOptions(Stripe.apiKey, Stripe.apiVersion, null, null, 30 * 1000, 80 * 1000);
+		return new RequestOptions(Stripe.apiKey, Stripe.apiVersion, null, null, DEFAULT_CONNECTION_TIMEOUT, DEFAULT_READ_TIMEOUT);
 	}
 
 	private final String apiKey;
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -4,19 +4,23 @@ import com.stripe.Stripe;
 
 public class RequestOptions {
 	public static RequestOptions getDefault() {
-		return new RequestOptions(Stripe.apiKey, Stripe.apiVersion, null, null);
+		return new RequestOptions(Stripe.apiKey, Stripe.apiVersion, null, null, 30 * 1000, 80 * 1000);
 	}
 
 	private final String apiKey;
 	private final String stripeVersion;
 	private final String idempotencyKey;
 	private final String stripeAccount;
+	private final int connectionTimeout;
+	private final int readTimeout;
 
-	private RequestOptions(String apiKey, String stripeVersion, String idempotencyKey, String stripeAccount) {
+	private RequestOptions(String apiKey, String stripeVersion, String idempotencyKey, String stripeAccount, int connectionTimeout, int readTimeout) {
 		this.apiKey = apiKey;
 		this.stripeVersion = stripeVersion;
 		this.idempotencyKey = idempotencyKey;
 		this.stripeAccount = stripeAccount;
+		this.connectionTimeout = connectionTimeout;
+		this.readTimeout = readTimeout;
 	}
 
 	public String getApiKey() {
@@ -33,6 +37,14 @@ public class RequestOptions {
 
 	public String getStripeAccount() {
 		return stripeAccount;
+	}
+	
+	public int getReadTimeout() {
+	    return readTimeout;
+	}
+	
+	public int getConnectionTimeout() {
+	    return connectionTimeout;
 	}
 
 	@Override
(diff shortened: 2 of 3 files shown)


src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -4,19 +4,23 @@
 
 public class RequestOptions {
 	public static RequestOptions getDefault() {
-		return new RequestOptions(Stripe.apiKey, Stripe.apiVersion, null, null);
+		return new RequestOptions(Stripe.apiKey, Stripe.apiVersion, null, null, 30 * 1000, 80 * 1000);
```
ob-stripe: Would you mind declaring two new constants `DEFAULT_READ_TIMEOUT = 80*1000` and `DEFAULT_CONNECTION_TIMEOUT = 30*1000` and using the named constants here?

ob-stripe:
Hi @agori, thanks a lot for this! Our other language libraries do allow passing custom timeouts, so I'm definitely in favor of this change.

I left a small comment, and we'd also need to add a test for this. It would be great if you could write one, but otherwise @brandur-stripe or I will do it.

src/main/java/com/stripe/net/RequestOptions.java
agori: Thanks for commenting on this. I'll add this and the test case over the weekend, if you don't mind.

ob-stripe: Not at all! Thanks a lot.

agori:
@ob-stripe please help :) . 
Tests in TimeoutTest pass when they are run individually but fail otherwise.
Even following code fails to throw an Exception in the second call:

```
Token.create(defaultTokenParams); //first call with defaults
RequestOptions options = RequestOptions.builder().setConnectionTimeout(1).build();
Token.create(defaultTokenParams, options); // second call
```
Is there any "caching"  I am not aware of?

ob-stripe:
I suspect that the HTTP client library uses persistent connections. We call `disconnect()` after every request ([here](https://github.com/agori/stripe-java/blob/7b37e38fc8b064f289ede7e85c796ed5d5fa7a2b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java#L382)) but it looks like this simply "indicates that other requests to the server are unlikely in the near future" and may not actually close the connection.

I'm not sure how to actually force a connection close -- I'll try digging some more.

agori:
I see now, JDK does some form of pooling behind the scenes (keep alive), so it skips the second attempt to connect. Read timeout work great instead.

I have found out that if you add a sufficient delay between test cases, the connection is not reused anymore.
So the following code passes:

```
Token.create(defaultTokenParams); //first call with defaults
Thread.sleep(5000); //inject delay
RequestOptions options = RequestOptions.builder().setConnectionTimeout(1).build();
Token.create(defaultTokenParams, options); // second call
```
I agree that injecting 5 seconds of delay is quite arbitrary and could make the test brittle, so I won't commit this trick for now. 

Let me know how you prefer to proceed.

ob-stripe:
Yeah, I can't find a clean way to force a new connection.

I agree that introducing a delay is less than ideal. A simple trick we could do is to connect to a different host that isn't used anywhere else in the test suite. I'm pretty sure the host doesn't even need to exist (if the name resolution counts towards the connection timeout -- not sure if that's the case).

Can you try calling `Stripe.overrideApiBase("https://timeout-test.stripe.com");` at the beginning of the test and `Stripe.overrideApiBase("https://api.stripe.com");` at the end, just to check if it works as expected? (Actually, we should probably make `apiBase` settable via `RequestOptions`...)

agori:
I see your point but a call to timeout-test would, well, timeout after 30 seconds :) (tested)
Further, if it works, it works more for the delay (30 seconds) than because we change the URL. Are you sure the JDK is pooling just the last URL? 

Anyway, what about using your idea but in this way?

```
        @Test(expected=APIConnectionException.class, timeout=2000)
	public void testConnectionTimeoutIsConsidered() throws StripeException {
		String nonExistentUrl = "https://timeout-test.stripe.com";
		int tooShortTimeoutInMillis = 1;
		RequestOptions options = RequestOptions.builder()
				.setConnectionTimeout(tooShortTimeoutInMillis)
				.build();
		
		
		Stripe.overrideApiBase(nonExistentUrl);
		Token.create(defaultTokenParams, options);
	}
```
The test should throw an `APIConnectionException` in less than 2 seconds. If I do not set setConnectionTimeout(1) the test fails because it timeouts after 30 seconds and more (notice the annotation @Test(...., timeout=2000)).

agori:
@ob-stripe have you missed my last comment? I'd like to close this and ditch my customised project :)
If you do not agree with my proposal, just let me know please.

ob-stripe:
Hey @agori, sorry, didn't mean to leave you hanging!

Actually, messing with the API base is not a good idea as it's a global setting so it could impact other tests (e.g. when running tests in parallel).

What we really want to do here is force a new connection. This is apparently possible by adding a `Connection: Close` header, but unfortunately there's no interface that would let us do this from a test and I'm reluctant to add a new interface just for this.

At this point I think we might just want to lose the connection timeout test :/

cc @brandur-stripe in case he has a better idea

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -149,8 +149,8 @@ protected PasswordAuthentication getPasswordAuthentication() {
 		} else {
 			conn = (HttpURLConnection) stripeURL.openConnection();
 		}
-		conn.setConnectTimeout(30 * 1000);
-		conn.setReadTimeout(80 * 1000);
+		conn.setConnectTimeout(options.getConnectionTimeout());
```
brandur-stripe: What do you think about keeping the naming of `Connect` instead of adding `Connection`? Besides just consistency, I think that the former is more accurate given that it's a verb (like read), and the timeout is the maximum time allowed as that action is being taken.

brandur-stripe:
Nice idea! I left one comment above, but this is looking very good. Thanks for the PR.

> At this point I think we might just want to lose the connection timeout test :/

Yeah ... I'd also recommend just removing it. I'm really not crazy about tests that rely on time-based side effects to operate because it just seems like the fragility is not worth the guarantee they're providing. It would be nice to get access to the actual `HttpURLConnection` object here and checking that its timeout has been set, but I don't think it's worth the trouble here.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
agori: Ok roger. I'll fix this.

agori:
Ok I am going to remove the connect time out test. I'll leave the read timeout one instead.

As a note, actually I think it would be possible to remove completely `Stripe.getApiBase` static property in favour of `RequestOptions.getApiBase` (options are always available as parameter of the current method or in the caller method). But I understand it would cause a huge refactory, at least in term of number lines touched, and break compatibility.

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -3,20 +3,29 @@
 import com.stripe.Stripe;
 
 public class RequestOptions {
+	
+	private final static int DEFAULT_CONNECTION_TIMEOUT = 30 * 1000;
```
brandur-stripe: Lets make this one `CONNECT` (instead of `CONNECTION`) as well.

brandur-stripe:
Thanks for the change!

> As a note, actually I think it would be possible to remove completely Stripe.getApiBase static property in favour of RequestOptions.getApiBase (options are always available as parameter of the current method or in the caller method). But I understand it would cause a huge refactory, at least in term of number of touched lines, and break compatibility.

Yeah, I expect that's true. In general, we kind of have a cross-library convention right now that configuration lives on the top-level `Stripe` class/module. Like you mention though, it makes sense to leave it there if only for purposes of backwards compatibility.

brandur-stripe:
And actually, speaking of the top-level `Stripe` class: I was just thinking that it's fairly customary for configuration like an API key to go there, and something like connect/read timeouts is likely to be one of those things that users need to only set once.

What do you think about putting the default configuration on `Stripe` and having `RequestOptions` pull from that by default (and optionally overridden with a builder).

agori:
I'll give it a try today or tomorrow (unless you prefer to step in?). 

I still need different timeouts per operation, because for example in my system the card is saved in the same http request, while payments are submitted in a queue (so the former should have a much shorter timeout). But I understand most of the time one timeout could fit all the situations.

brandur-stripe:
> I still need different timeouts per operation, because for example in my system the card is saved in the same http request, while payments are submitted in a queue (so the former should have a much shorter timeout). But I understand most of the time one timeout could fit all the situations.

Yeah, I suspect it will be useful for the general case.  I didn't mention it before, but it's also partly for purposes of consistency: the timeout setters on stripe-ruby for example are available at the `Stripe` module level.

Alright, I'm going to pull this into another branch and add a minor fix on top of that.

PR #372, opened 2017-05-01
Add `three_d_secure` accessors on Card model

### LINKED ISSUE #371: No three_d_secure attribute on the Card object
According to the docs the Card object should have a three_d_secure attribute.

https://stripe.com/docs/api#card_object-three_d_secure
https://stripe.com/docs/testing#3ds-cards

I have looked around the code and it doesn't seem like that is being exposed anywhere in the stripe-java API.  Am I missing something or is this an oversight.

I would switch over to the Source api (rather than creating Cards) however I can't see how to add a Source to a Customer so that option seems out as well.

brandur-stripe:
Adds missing `three_d_secure` accessors on the Card mode. They're supposed to have been there for a while now, and it's just an oversight that they're currently missing.

Fixes #371.

[1] https://stripe.com/docs/api#card_object-three_d_secure

r? @ob-stripe

### DIFF
--- a/src/main/java/com/stripe/model/Card.java
+++ b/src/main/java/com/stripe/model/Card.java
@@ -34,6 +34,7 @@ public class Card extends ExternalAccount {
 	String name;
 	String recipient;
 	String status;
+	ThreeDSecure threeDSecure;
 	String tokenizationMethod;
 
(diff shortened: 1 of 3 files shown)


ob-stripe:
lgtm!

PR #374, opened 2017-05-24
Add `account` under the event model

brandur-stripe:
The `user_id` field is being renamed to `account`. I've left the `user_id` accessors in here for backwards compatibility.

r? @remi-stripe

### DIFF
--- a/src/test/java/com/stripe/model/EventTest.java
+++ b/src/test/java/com/stripe/model/EventTest.java
@@ -31,11 +31,11 @@ public class EventTest extends BaseStripeTest {
 
 		assertEquals(reserializedEvent.getId(), event.getId());
 		assertEquals(reserializedEvent.getObject(), event.getObject());
+		assertEquals(reserializedEvent.getAccount(), event.getAccount());
 		assertEquals(reserializedEvent.getApiVersion(), event.getApiVersion());
 		assertEquals(reserializedEvent.getCreated(), event.getCreated());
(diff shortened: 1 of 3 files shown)


src/test/java/com/stripe/model/EventTest.java
```diff
@@ -31,11 +31,11 @@ public void serializesToJson() throws IOException {
 
 		assertEquals(reserializedEvent.getId(), event.getId());
 		assertEquals(reserializedEvent.getObject(), event.getObject());
+		assertEquals(reserializedEvent.getAccount(), event.getAccount());
```
remi-stripe: how does the new test works if you haven't changed the fixture's json?

remi-stripe: Update: that test doesn't even support the old `user_id` key so I guess we're good though testing it explicitly might be worth it.

remi-stripe:
LGTM (though a test change wouldn't hurt)

src/test/java/com/stripe/model/EventTest.java
brandur-stripe: I fully admit that it kind of sucks, but it works because both values are `null`.

brandur-stripe: Alright, I updated the fixture in 304fc78.

brandur-stripe:
Lost track of this one a few weeks back unfortunately, but it's out now! Released as 5.1.0.

PR #375, opened 2017-05-25
Add support for login_links on the Account resource

remi-stripe:
This adds support for creating a login_link on an Account. Also adds support for the new `type` property/parameter on the Account resource instead of `managed`

r? @brandur


brandur-stripe:
Released as 4.8.0.

PR #376, opened 2017-05-25
Add event request object

brandur-stripe:
Event's `request` field has changed from being a string to being an object that contains a request ID and an idempotency key.

I don't think there's an easy way of making this backwards compatible, and since it's probably not in all that common of use, I've just changed the type completely here.

r? @remi-stripe ~~One thing I noticed while testing is that if you load an event from an old API version (before the type changed) with this code, you get a GSON error because it's expecting an object instead of a string. I don't really like that if a user upgrades this SDK and is still on an old API version, that could throw an error. The right answer is probably to lock stripe-java to a version like we do for stripe-go, but I'm not sure I want to shave that yak right now. Any thoughts on a better way to handle this?~~

### DIFF
--- a/src/main/java/com/stripe/model/EventRequestDeserializer.java
+++ b/src/main/java/com/stripe/model/EventRequestDeserializer.java
@@ -1,0 +1,34 @@
+package com.stripe.model;
+
+import com.google.gson.FieldNamingPolicy;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+import com.google.gson.reflect.TypeToken;
+
+import java.lang.reflect.Type;
+import java.util.List;
+
+public class EventRequestDeserializer implements JsonDeserializer<EventRequest> {
+
+	public EventRequest deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+		throws JsonParseException {
+		Gson gson = new GsonBuilder()
+			.setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+			.create();
+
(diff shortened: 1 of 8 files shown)


brandur-stripe:
@remi-stripe I ended up patching this for backwards compatibility. Can you take a look? Thanks!

src/main/java/com/stripe/model/EventRequestDeserializer.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+import com.google.gson.reflect.TypeToken;
+
+import java.lang.reflect.Type;
+import java.util.List;
+
+public class EventRequestDeserializer implements JsonDeserializer<EventRequest> {
+
+	public static final Type REFUND_LIST_TYPE = new TypeToken<List<FeeRefund>>() {}.getType();
```
remi-stripe: I think this needs to be removed

brandur-stripe: Arg, it's been a long day. Thanks!

brandur-stripe:
Released as 4.9.0.

ob-stripe:
Nice, but isn't this a breaking change? Every `event.getRequest()` would need to be updated to `event.getRequest().getId()`, right?

brandur-stripe:
> Nice, but isn't this a breaking change? Every event.getRequest() would need to be updated to event.getRequest().getId(), right?

Ah, shoot, absolutely right. I dropped the ball on semver, heh.

So I'd kind of propose as a fix that I release `HEAD` as 5.0.0. I thought about yanking 4.9.0, but I think that it the churn that would cause for users who have already locked to it in the last week is worse than the mislabeled version. I will retroactively edit the CHANGELOG to highlight the error though. Sound okay?

ob-stripe:
sgtm

brandur-stripe:
Great! Thank you. Released as 5.0.0 and I've amended the `CHANGELOG`.

PR #381, opened 2017-06-09
Expandable attributes

ob-stripe:
r? @brandur-stripe 

This PR adds support for expansion on all API-expandable attributes that are not supported yet.

The library should now support deserializing every expandable attribute ~save for [`parent`](https://stripe.com/docs/api#order_item_object-parent) on order_item which needs to be handled more or less like #365 (the expanded object can be either a SKU or a discount)~.

It also adds `retrieve()` methods with support for passing parameters to resources with at least one expandable attributes, so users can actually send the `expand` parameter when retrieving a single resource.


src/main/java/com/stripe/model/ChargeRefundCollectionDeserializer.java
```diff
@@ -18,6 +20,7 @@ public ChargeRefundCollection deserialize(JsonElement json, Type typeOfT, JsonDe
 		throws JsonParseException {
 		Gson gson = new GsonBuilder()
 			.setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+			.registerTypeAdapter(ExpandableField.class, new ExpandableFieldDeserializer())
```
ob-stripe: I had to add the ExpandableField deserializer to both `ChargeRefundCollectionDeserializer` and `FeeRefundColelctionDeserializer`. There's probably a better way of doing this but it would require some refactoring.

brandur-stripe:
OMG. Sorry OB, I didn't get a chance to take a look at this today, but this is **SWEET**. Really great work.

ob-stripe:
After investigating a bit more, it appears that when expanding `item.parent` on order objects, the order_item's `parent` attribute will only be expanded if it's a SKU, not if it's a discount. This means that the attribute can be expanded with a simple `ExpandableField<SKU>` since the expanded object has only one possible type.

src/main/java/com/stripe/model/OrderItem.java
```diff
@@ -1,11 +1,13 @@
 package com.stripe.model;
 
-public class OrderItem extends StripeObject {
+import com.stripe.net.APIResource;
```
ob-stripe: I had to make `OrderItem` extend `APIResource` instead of `StripeObject` so it can use `ExpandableField`. Technically, order_items are a subresource and not a first-order API resource, but I don't think there's any ill effect to making it extend `APIResource`.

That said, maybe we ought to move `ExpandableField` so that it's usable from `StripeModel` instead?

src/test/resources/com/stripe/model/balance_transaction_expansions.json
```diff
@@ -18,11 +18,55 @@
...
-      "balance_transactions": []
+    "id": "dp_1AF5AmKac2qaDcCZHvOjyrDo",
+    "object": "dispute",
+    "amount": 5000,
+    "balance_transaction": "txn_1AF5AmKac2qaDcCZm8tBQt6I",
+    "balance_transactions": [],
+    "charge": "ch_1AF5AlKac2qaDcCZ4Iu07uIA",
+    "created": 1497287247,
+    "currency": "usd",
+    "evidence": {
```
ob-stripe: I completed the model to be a full dispute object.

Prior to [16780a4](https://github.com/stripe/stripe-java/pull/381/commits/16780a4b6b74cfbd73bb17ea052f1b43fb8b4b84), `BalanceTransaction` only supported expanding `source` to a charge, transfer or refund. Since I added support for all possible types, including dispute, `DisputeDataDeserializer` is now correctly invoked when deserializing this balance transaction, and this deserializer expects an `evidence` key.

brandur-stripe:
> After investigating a bit more, it appears that when expanding item.parent on order objects, the order_item's parent attribute will only be expanded if it's a SKU, not if it's a discount. This means that the attribute can be expanded with a simple ExpandableField<SKU> since the expanded object has only one possible type.

So to me this sounds a little like an implementation quirk, and not necessarily the behavior that we'd like to happen. On another PR we recently introduced a pattern whereby balance transaction sources are an `ExpandableField<HasId>` and we could potentially apply a similar pattern here. Thoughts?

src/main/java/com/stripe/model/OrderItem.java
brandur-stripe: So since order items _are_ API resources, I think it's perfectly okay to change their subclass here.

We probably should make `ExpandableField` available on `StripeObject` though.

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
```diff
@@ -50,15 +65,8 @@ public BalanceTransaction deserialize(JsonElement json, Type typeOfT, JsonDeseri
...
-				if ("charge".equals(val.getAsString())) {
-					sourceObjClass = Charge.class;
-				} else if ("transfer".equals(val.getAsString())) {
-					sourceObjClass = Transfer.class;
-				} else if ("refund".equals(val.getAsString())) {
-					sourceObjClass = Refund.class;
-				}
-				// TODO support other source types (?)
+				String type = val.getAsString();
+				Class<? extends HasId> sourceObjClass = sourceObjMap.get(type);
```
brandur-stripe: Very nice improvement here :)

src/main/java/com/stripe/model/BalanceTransactionDeserializer.java
```diff
@@ -12,30 +10,47 @@
...
+	static {
+		sourceObjMap.put("application_fee", ApplicationFee.class);
+		sourceObjMap.put("charge", Charge.class);
+		sourceObjMap.put("dispute", Dispute.class);
+		sourceObjMap.put("fee_refund", FeeRefund.class);
+		sourceObjMap.put("payout", Payout.class);
+		sourceObjMap.put("refund", Refund.class);
+		sourceObjMap.put("transfer", Transfer.class);
+		sourceObjMap.put("transfer_reversal", Reversal.class);
+	}
```
brandur-stripe: Thanks a lot for finishing this off! Much cleaner implementation too.

src/main/java/com/stripe/model/ChargeRefundCollectionDeserializer.java
brandur-stripe: Yeah, +1. The thing where we manually instantiate new GSONs all over the place that only have a set of hand-picked custom deserializer support seems super brittle to me and we should fix it, but this seems like the right thing to do for now.

brandur-stripe:
Once again, great work here OB! And bonus points for making it backwards compatible!

The one thing that struck me on this one is that it would be nice to have a few more tests around deserializing each type of model with and without expansions, but given how sorry the shape of model tests is right now, it's definitely not fair to dump all of that on you right now.

I wouldn't be surprised if we missed something somewhere, but I'd say that given that this PR is big enough that it could be prone to rot, we should just go ahead and bring it in and iterate on problems or omissions as they come up. What do you think?

ob-stripe:
> On another PR we recently introduced a pattern whereby balance transaction sources are an ExpandableField<HasId> and we could potentially apply a similar pattern here. Thoughts?

This is actually what I did first. It's when I tried to test it that I realized that only SKUs were expanded and not discounts.

I could certainly reintroduce the custom `OrderItemDeserializer` I wrote (which was basically a copy/paste of `BalanceTransactionDeserializer`), but if it's only going to expect one expanded class for the `parent` attribute (i.e. `SKU`), I thought we might as well simply use `ExpandableField<SKU>`.

> The one thing that struck me on this one is that it would be nice to have a few more tests around deserializing each type of model with and without expansions, but given how sorry the shape of model tests is right now, it's definitely not fair to dump all of that on you right now.

Agreed on both counts :p 

> I wouldn't be surprised if we missed something somewhere, but I'd say that given that this PR is big enough that it could be prone to rot, we should just go ahead and bring it in and iterate on problems or omissions as they come up. What do you think?

Yes, agreed. It's definitely possible I missed some things (I did use the OpenAPI spec to find all expandable attributes but it was still done manually). If that's the case, we can always tackle that later though.

brandur-stripe:
> I could certainly reintroduce the custom OrderItemDeserializer I wrote (which was basically a copy/paste of BalanceTransactionDeserializer), but if it's only going to expect one expanded class for the parent attribute (i.e. SKU), I thought we might as well simply use ExpandableField<SKU>.

Can we dig into this a bit more? I just looked at the code and it certainly _looks_ like discounts are also supposed to be expandable. It's possible they're not, but it seems like that would be more of a bug than a feature?

I guess my point is that even if only SKUs are expandable, if it looks like an implementation quirk rather than something that was purposely designed, we should probably prefer `HasId` here just to minimize the probability of having to introduce a breaking change somewhere down the line.

ob-stripe:
@brandur-stripe That makes sense. I updated the PR to use a custom deserializer for order items.

The test is unchanged except that I used `getParentObjectAs()` instead of `getParentObject()` to avoid casting.

brandur-stripe:
> The test is unchanged except that I used getParentObjectAs() instead of getParentObject() to avoid casting.

Nice! Thanks for re-using the balance transaction conventions here. This looks great to me.

brandur-stripe:
Released as 5.3.0.

PR #382, opened 2017-06-13
Add support for ephemeral keys

grey-stripe:
This adds the `EphemeralKey` resource to the bindings.  Currently, the only supported operations are creation and deletion.

Because Java is statically typed, we define a custom JSON deserializer `EphemeralKeyDeserializer` so that we can preserve the raw JSON so it can be passed back to the frontend.  Otherwise, any unrecognized fields would be lost by deserialization/reserialization.

These bindings do not have the same behavior regarding the API version as the other bindings.  The API version is set via the `RequestOptions` object, but the object's API version is copied from the globally configured `Stripe.apiVersion`, so we can't tell if an API version was explicitly specified on the `RequestOptions`.  I doubt this is worth changing the behavior for, so I've left it as-is, but we will still throw an `IllegalArgumentException` if the resulting API version is `null`.

### DIFF
--- a/src/main/java/com/stripe/model/EphemeralKey.java
+++ b/src/main/java/com/stripe/model/EphemeralKey.java
@@ -20,7 +20,7 @@ public class EphemeralKey extends APIResource implements HasId {
 	Boolean livemode;
 	String secret;
 	List<AssociatedObject> associatedObjects;
-    transient JsonObject rawJson;
+	transient JsonObject rawJson;
 
 	public String getId() {
 		return id;
@@ -95,7 +95,7 @@ public class EphemeralKey extends APIResource implements HasId {
 	public EphemeralKey delete()
 			throws AuthenticationException, InvalidRequestException,
 			APIConnectionException, CardException, APIException {
-        return delete((RequestOptions) null);
+		return delete((RequestOptions) null);
 	}
 
 	public EphemeralKey delete(RequestOptions options)
--- a/src/main/java/com/stripe/model/EphemeralKey.java
+++ b/src/main/java/com/stripe/model/EphemeralKey.java
@@ -89,6 +89,10 @@ public class EphemeralKey extends APIResource implements HasId {
 	public static EphemeralKey create(Map<String, Object> params, RequestOptions options)
 			throws AuthenticationException, InvalidRequestException,
 			APIConnectionException, CardException, APIException {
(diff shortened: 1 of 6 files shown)


src/main/java/com/stripe/model/EphemeralKey.java
```diff
@@ -0,0 +1,131 @@
...
+		return delete((RequestOptions) null);
+	}
+
+	public EphemeralKey delete(RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return request(RequestMethod.DELETE, instanceURL(EphemeralKey.class, this.id), null, EphemeralKey.class, options);
+	}
+
+	public static class AssociatedObject extends StripeObject {
```
brandur-stripe: So I think it's convention to have all models (even submodels) at the top level of the `model` package (for example, `EventData` and `ChargeOutcomeRule`). Just to keep inline with that pattern, I think we should rename this to something like `EphemeralKeyAssociatedObject`and put it there.

src/main/java/com/stripe/model/EphemeralKey.java
```diff
@@ -0,0 +1,131 @@
...
+
+	public List<AssociatedObject> getAssociatedObjects() {
+		return associatedObjects;
+	}
+
+	public void setAssociatedObjects(List<AssociatedObject> associatedObjects) {
+		this.associatedObjects = associatedObjects;
+	}
+
+	public JsonObject getRawJson() {
```
brandur-stripe: Hmm, I'm not too sure that we should export a GSON type as part of our public API. What do you think about making this a more opaque data type (like you did with `[]byte` in Go)?

brandur-stripe:
Left a couple comments, but looks largely good to me! Thanks.

grey-stripe:
I changed the raw JSON field to a string and moved `EphemeralKeyAssociatedObject` to its own file.

brandur-stripe:
Thanks!

Sorry for not calling this out earlier, but would you also mind adding a basic model test [in this style](https://github.com/stripe/stripe-java/blob/master/src/test/java/com/stripe/model/BalanceTransactionTest.java). It's a little hit or miss whether we have them right now, but I think this is roughly the direction that we should moving towards for testing (and away from live API calls), especially when we have custom deserializers.

PR #385, opened 2017-06-22
adding automatically generated build files

mrmcduff-stripe:
r? @brandur-stripe 
cc @teich-stripe 

Just showing how we can add gradle builds to our stripe-java repo. This _might_ work for travis too - I'll see when it runs.


brandur-stripe:
@mrmcduff-stripe Nice! Thanks for taking a shot at this!

I gotta say that even if the speed improvements aren't too different, just the more sane output alone might be a reason to switch. If you look at [a sample Gradle build](https://travis-ci.org/stripe/stripe-java/jobs/245986910) versus [a sample Maven build](https://travis-ci.org/stripe/stripe-java/jobs/245540804) they're ~1300 lines difference! (~200 vs. ~1500)

I guess a couple questions/notes:

* Would we safely be able to remove our `pom.xml` with this coming in?
* I'm not 100% sure we can kill Java 6, but it actually does seem somewhat plausible now that I look. It looks like < 0.5% of registered Java requests to the API over the last 24 hours were using it.
* Are there any other downsides to the change that we should know about?

mrmcduff-stripe:
you can remove the Pom, but we might need to come up with some version of the deploy.gradle script I have in stripe-android to ship your artifact (it actually generates a pom file and sends that to sonatype).

Otherwise the maven file would have no more affect on the build (provided you were inclined to build with gradle).

We can actually keep java 6 if we fall back to an earlier version of gradle (i think 3.5 still supports 6?), but then you lose the really cool command line interface that gradle 4.0 has.

This won't affect at all how our users would be able to import the artifact (they would still be free to use Maven). I'd have to google around some to figure out if there are any drawbacks that get to people.

brandur-stripe:
Thanks!

> Otherwise the maven file would have no more affect on the build (provided you were inclined to build with gradle).

Ah yeah. I guess the part that would concern me here is that you'd be sharing a lot duplicate information between `pom.xml` and `build.gradle`. That could lead to problems like where you update one but forget to update the other (for example). Overall it'd be preferable to just have one or the other, so I'd be curious to see if we could rig up something similar to your Android solution.

mrmcduff-stripe:
Update: keeping Java 6 would require going back to gradle 2.1.14

ob-stripe:
We removed support for Java 6 back in #383 + #395, so I took the liberty of rebasing and updating this PR.

I had to make a few changes to `.travis.yml` to work around a couple of issues with OpenJDK 7:
- use `dist: precise` to work around a TLS error: https://github.com/gradle/gradle/issues/2421
- use the `hosts` and `hostname` addons to work around a buffer overflow error: https://github.com/travis-ci/travis-ci/issues/5227#issuecomment-165131913

brandur-stripe:
Nice @olivierbellone :)

@mrmcduff-stripe @oliver-stripe Would one of you be willing to try and get this one across the finish line by finalizing any remaining items? It seems like the build's working now, but it'd be nice to (1) get the plumbing in to do new releases from Gradle, and (2) pull out the rest of the old Maven infrastructure. I don't think all this necessarily has to happen on PR, but ideally someone would be keeping an eye on making sure that we rolled the whole project through to completion.

mrmcduff-stripe:
@brandur-stripe -- I can schedule some time for this next week. I'm also on a thread with the folks from Gradle who mentioned that the publishing script in this PR should create a POM file that is functionally identical to the one we generate via Maven. My original concern was that a sub-file in the directory didn't appear with this script, but did via Maven, and I had no idea how to test whether or not that was the same (both worked with a sample project).

brandur-stripe:
> I can schedule some time for this next week. I'm also on a thread with the folks from Gradle who mentioned that the publishing script in this PR should create a POM file that is functionally identical to the one we generate via Maven. My original concern was that a sub-file in the directory didn't appear with this script, but did via Maven, and I had no idea how to test whether or not that was the same (both worked with a sample project).

Nice! That'd be great.

Yeah, I'm not sure about the different in subdirectory. I'm not sure I'd be worried about necessarily getting this before and after bit-perfect as long as Gradle's able to push an artifact that's valid and works as expected when people upgrade.

PR #387, opened 2017-06-27
Add parameters when calling `pay` on an invoice

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries @ark-stripe

Users can now provide a `source` parameter when paying invoices, but the `pay` method currently does not let users provide any parameters. This PR fixes that.

Not a breaking change thanks to static typing.

### DIFF
--- a/src/main/java/com/stripe/model/Invoice.java
+++ b/src/main/java/com/stripe/model/Invoice.java
@@ -384,6 +384,12 @@ public class Invoice extends APIResource implements MetadataStore<Invoice>, HasI
 		return this.pay((RequestOptions) null);
 	}
 
+	public Invoice pay(Map<String, Object> params) throws AuthenticationException,
+			InvalidRequestException, APIConnectionException, CardException,
+			APIException {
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/model/Invoice.java
```diff
@@ -488,7 +494,12 @@ public Invoice pay(String apiKey) throws AuthenticationException,
 	public Invoice pay(RequestOptions options) throws AuthenticationException,
```
brandur-stripe: I think maybe the one thing that we should do here is deprecate this usage since it's inconsistent with other method calls now. I'll take care of it in master.

brandur-stripe:
LGTM.

brandur-stripe:
Released as 5.6.0.

src/main/java/com/stripe/model/Invoice.java
ob-stripe: Great point. Thanks!

PR #390, opened 2017-07-06
Use IntelliJ to clean up the code base

kjc-stripe:
Most of our test code still uses spaces instead of tabs. Let's fix that. Also do some general cleanup as well.

Shift + Cmd + A -> Code Cleanup


shale-stripe:
LGTM

PR #392, opened 2017-07-06
Provide a type-safe API method interface

kjc-stripe:
The current interface for creating and updating records requires manual construction of HashMaps. 

```java
Map<String, Object> planParams = new HashMap<String, Object>();
planParams.put("amount", 5000);
planParams.put("interval", "month");
planParams.put("name", "Gold large");
planParams.put("currency", "usd");
planParams.put("id", "gold-large");
Plan.create(planParams);
```

This pull requests uses Google AutoValue to generate a typed interface for API requests. Creating a plan can now be down in a type-safe manner.

```java
CreatePlan params = CreatePlan.builder()
    .amount(5000)
    .interval("month")
    .name("Gold large")
    .currency("usd")
    .id("gold-large")
    .build()
Plan.create(params);
```

### Breaking changes
If you were calling a method with null before, you'll need to cast or pass an empty value object now.

```diff
-Account account = Account.create(null);
+Account account = Account.create(new HashMap<String, Object>());
```

### TODO
- [x] Support nullable arguments
- [x] Update all methods to include a signature with the correct ValueObject

### Future work
- Support nested input parameters
- Use annotations to instead of `toMap`
- Type safe interface for retrieval methods (`all`, `retrieve`, etc.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,6 +15,17 @@
     </licenses>
     <dependencies>
         <dependency>
+            <groupId>com.google.auto.value</groupId>
+            <artifactId>auto-value</artifactId>
+            <version>1.2</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <version>3.0.2</version>
+        </dependency>
+        <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
             <version>2.2.4</version>
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -7,6 +7,9 @@ import com.google.gson.annotations.SerializedName;
 import com.stripe.exception.*;
 import com.stripe.net.APIResource;
 import com.stripe.net.RequestOptions;
+import com.stripe.params.CreateAccount;
+import com.stripe.params.RejectAccount;
+import com.stripe.params.UpdateAccount;
 
 public class Account extends APIResource implements HasId, MetadataStore<Account> {
 	String id;
@@ -408,5 +411,41 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 		public String getSecret() {
 			return secret;
 		}
+	}
+
+	public Account reject(RejectAccount params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return reject(params, (RequestOptions) null);
+	}
+
+	public Account reject(RejectAccount params, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return reject(params.toMap(), options);
+	}
+
+	public static Account create(CreateAccount params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return create(params, (RequestOptions) null);
+	}
+
+	public static Account create(CreateAccount params, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return create(params.toMap(), options);
+	}
+
+	public Account update(UpdateAccount params)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return update(params, (RequestOptions) null);
+	}
+
+	public Account update(UpdateAccount params, RequestOptions options)
+			throws AuthenticationException, InvalidRequestException,
+			APIConnectionException, CardException, APIException {
+		return update(params.toMap(), options);
 	}
 }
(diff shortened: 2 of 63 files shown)


kjc-stripe:
So I'm not super happy that I needed to pull in a jsr jar to get the `@Nullable` annotation.

kjc-stripe:
Alright, now that this is passing, I'd be interested in hearing feedback. cc @evan-stripe @olivier-stripe @mrmcduff-stripe

mrmcduff-stripe:
I've never seen Google's AutoValue package before -- is there a reason it makes all the fields package-private instead of private? I also really wish that the `toMap` function it (very helpfully) creates was an `@Override` of some interface so that you could know it was there in a group of superclass types.

Otherwise, seems like a good thing. Is this lib particularly large in build size?

kjc-stripe:
@mrmcduff-stripe the `toMap` method is actually my doing. I'm using a script to generate all the files in the `params` package. I can add a `Mappable` interface for all of our auto values, though I'm not sure how useful that would be.

mrmcduff-stripe:
It looks like the only thing in the diff is the conversion to using this new style of model -- just wanted to confirm that before I lgtm-stamp a 3000+ line diff :) @kjc-stripe

kjc-stripe:
Sorry, I should have included a note to all the reviewers. I'm looking for feedback on the proposed interface changes, especially the backwards incompatible change. Basically, a go/no-go decision for this interface. If you vote no-go, I'd love to hear your concerns.

kjc-stripe:
Also, here's the code I used to generate this PR https://gist.github.com/kjc-stripe/6824d1081808906f82f4a0c366bf45b1

mrmcduff-stripe:
I'm all for it. Will wait for others to have a chance to look it over

evan-stripe:
No objections from me

remi-stripe:
I think we need to find a viable solution to the API version pinning only when using the Builder pattern before we can merge. Some of those named parameters only work with the latest API version but the map solution works with all.

kjc-stripe:
So, I sadly don't have to the time right now to get this to completion. I'm going to close this until that happens, or someone else wants to take. 

@remi-stripe If you have time, I'd love to hear your thoughts on how we should do API pinning

remi-stripe:
@kjc-stripe As long as it's behind the builder pattern I think it should be doable to pin to a specific API version. Part of the issue is that if we update it often (and we should) we end up with major versions constantly similarly to the Go library which is not the best solution if we want people to upgrade often.

I'd say that the first step would be to discuss what we want out of this new behavior and make some decisions. This should help unblock shipping this afterwards. cc @brandur-stripe

brandur-stripe:
> So, I sadly don't have to the time right now to get this to completion. I'm going to close this until that happens, or someone else wants to take.

Nooooooo.

Alright, I'm going to try and solicit some interest internally. We have one person coming on who might be in need of a project, so there's hope.

> I'd say that the first step would be to discuss what we want out of this new behavior and make some decisions. This should help unblock shipping this afterwards. cc @brandur-stripe

Yeah. Those ideas have been getting bounced around for a while, but I guess we've never decided on anything. I personally suspect that the major version bumps aren't the end of the world. People don't tend to upgrade their dependencies that often anyway, and when they do, (1) we do a pretty job of documenting the changes, and (2) often changes will be in parts of the library they're not using directly anyway. Deprecating fields also has some nice properties around user friendliness as well; on an API version upgrade instead of fields just silently becoming `null` because they've been removed, at least users will get a compiler error instead.

Remi, what do you think is the best way forward here? Do you want me to write an internal RFC? I have a feeling that this is the type of change that might be relatively uncontentious, but could do well with a rubber stamp if that's the case.

remi-stripe:
> Remi, what do you think is the best way forward here? Do you want me to write an internal RFC? I have a feeling that this is the type of change that might be relatively uncontentious, but could do well with a rubber stamp if that's the case.

I mostly agree. I think we need agreement, likely internally, on the best path forward and how to handle API version changes. We do this in an ad-hoc/random way on stripe-go right now. If two libraries (three if we count stripe-dotnet) end up with enforced API version, we need better documentation to actively upgrade the API version in the bindings whenever a new one is released.

Java is also a bit specific because we will enforce the API version but only for a specific type of integration using the builder. The other solution would be to take this as an opportunity to rework the library completely, only support the builder method moving forward and stop supporting old API versions and custom parameters.

brandur-stripe:
> we need better documentation to actively upgrade the API version in the bindings whenever a new one is released.

Yeah definitely. One thing to maybe note is that this wouldn't too much different than what we're doing right now (or what we should be doing right now) because we're still adding new things to the library, just without the version bump.

> The other solution would be to take this as an opportunity to rework the library completely, only support the builder method moving forward and stop supporting old API versions and custom parameters.

Yeah, this might be a good idea. I suspect that we should wait until we have a much stronger reason for doing it though, because we're going to create some pretty huge legacy if we cut over to something that's totally incompatible.

kjc-stripe:
This pull request is now out-dated and stale. I'm going to close it, but we should open it up in the future when we want to tackle this again.

PR #393, opened 2017-07-11
Use standard library implementation for comparing signatures

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Now that we've dropped support for Java 6, we can use `MessageDigest.isEqual()` to compare signatures (the implementation was vulnerable in early Java 6 versions, but is secure as of [Java 6u17](http://www.oracle.com/technetwork/java/javase/6u17-141447.html)).


brandur-stripe:
Great clean up :) LGTM.

PR #394, opened 2017-07-11
Force UTF-8 encoding when verifying signatures

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This was reported by a user. [`getBytes()`](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#getBytes()) without any parameters will use the platform's default encoding. Since the header and payload sent by Stripe will always UTF-8, we should force the encoding when decoding these strings.

### DIFF
--- a/src/main/java/com/stripe/net/Webhook.java
+++ b/src/main/java/com/stripe/net/Webhook.java
@@ -4,6 +4,7 @@ import com.stripe.exception.SignatureVerificationException;
 import com.stripe.model.Event;
 import com.stripe.model.StripeObject;
 
+import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Good point.

I'm going to ship this as a patch release, but hmm, it's somewhat debatable given the addition of a new exception type will break compilation in some cases.

brandur-stripe:
Released as 5.7.1.

ob-stripe:
I think a patch release is fine -- only the interfaces for `computeSignature()` and `computeHmacSHA256()` were modified, and users shouldn't be calling those directly. (The former is private, the latter is public though that's mostly out of convenience for tests.)

PR #395, opened 2017-07-11
Drop support for Java < 1.7

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

A few updates since we officially no longer support Java 6.

We also probably can get rid of some code in `StripeSSLSocketFactory` as Java 7 should always support TLS 1.2 (though it still needs to be explicitly enabled as it isn't used by default), but any such changes should be handled in a separate PR and tested thoroughly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: java
 
 jdk:
-  - openjdk6
   - openjdk7
   - oraclejdk7
   - oraclejdk8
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks Olivier!

> We also probably can get rid of some code in StripeSSLSocketFactory as Java 7 should always support TLS 1.2 (though it still needs to be explicitly enabled as it isn't used by default), but any such changes should be handled in a separate PR and tested thoroughly.

Good call.

PR #398, opened 2017-08-29
OAuth support

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds native support for OAuth in the Java library.

The implementation is mostly based on the existing ones (cf. https://github.com/stripe/stripe-ruby/pull/523). One notable difference is that I split the existing `_request` method in `LiveStripeResponseGetter` into two different `_request` (for normal API requests) and `_oauthRequest` (for OAuth requests) methods, with the common code factorized into a new `_rawRequest` method.

I did this because if I had simply updated the `_request` method to handle OAuth errors (like I did in the other languages implementations), then I'd have to update every request method in every model to account for the new possible OAuth exceptions (even though in practice, OAuth exceptions will never occur in normal API requests, but the compiler has no way of knowing this).

There might be a smarter / more elegant way of handling this. Open to all suggestions here!

### DIFF
--- a/src/main/java/com/stripe/model/oauth/TokenResponse.java
+++ b/src/main/java/com/stripe/model/oauth/TokenResponse.java
@@ -1,0 +1,33 @@
+package com.stripe.model.oauth;
+
+import com.stripe.model.StripeObject;
+
+public class TokenResponse extends StripeObject {
+	Boolean livemode;
+	String scope;
+	String stripeUserId;
+
+	public Boolean getLivemode() {
+		return livemode;
+	}
+
+	public void setLivemode(Boolean livemode) {
+		this.livemode = livemode;
+	}
+
+	public String getScope() {
+		return scope;
+	}
+
+	public void setScope(String scope) {
+		this.scope = scope;
+	}
+
+	public String getStripeUserId() {
+		return stripeUserId;
+	}
+
+	public void setStripeUserId(String stripeUserId) {
+		this.stripeUserId = stripeUserId;
+	}
+}
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -8,6 +8,13 @@ import com.stripe.exception.PermissionException;
 import com.stripe.exception.RateLimitException;
 import com.stripe.exception.CardException;
 import com.stripe.exception.InvalidRequestException;
+import com.stripe.exception.oauth.InvalidClientException;
+import com.stripe.exception.oauth.InvalidGrantException;
+//import com.stripe.exception.oauth.InvalidRequestException;
+import com.stripe.exception.oauth.InvalidScopeException;
+import com.stripe.exception.oauth.OAuthException;
+import com.stripe.exception.oauth.UnsupportedGrantTypeException;
+import com.stripe.exception.oauth.UnsupportedResponseTypeException;
 import com.stripe.model.StripeCollectionInterface;
 
 import javax.net.ssl.HttpsURLConnection;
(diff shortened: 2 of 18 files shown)


ob-stripe:
Tests are failing because of an unrelated issue. They should pass once #399 is merged and this PR is rebased.

src/main/java/com/stripe/model/oauth/TokenResponse.java
```diff
@@ -0,0 +1,98 @@
...
+		return stripeUserId;
+	}
+
+	public void setStripeUserId(String stripeUserId) {
+		this.stripeUserId = stripeUserId;
+	}
+
+	/**
+	 * @deprecated Use `stripe_user_id` with the `Stripe-Account` header (https://stripe.com/docs/connect/authentication#stripe-account-header)
+	 */
```
brandur-stripe: Is the idea to bring in deprecated fields for older API versions? I wonder if we should wait until someone complains ... :)

src/main/java/com/stripe/net/OAuth.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		if (clientId == null) {
+			throw new AuthenticationException(
+					"No client_id provided. (HINT: set client_id key using 'Stripe.client_id = <CLIENT-ID>'. "
+							+ "You can find your client_ids in your Stripe dashboard at "
+							+ "https://dashboard.stripe.com/account/applications/settings, "
+							+ "after registering your account as a platform. See "
+							+ "https://stripe.com/docs/connect/standard-accounts for details, "
+							+ "or email support@stripe.com if you have any questions.",
```
brandur-stripe: I think we'll have to change this error message right?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -605,6 +653,33 @@ private static void handleAPIError(String rBody, int rCode, String requestId)
...
+			case "invalid_request":
+				throw new com.stripe.exception.oauth.InvalidRequestException(code, description, requestId, rCode, null);
+			case "invalid_scope":
+				throw new InvalidScopeException(code, description, requestId, rCode, null);
+			case "unsupported_grant_type":
+				throw new UnsupportedGrantTypeException(code, description, requestId, rCode, null);
+			case "unsupported_response_type":
+				throw new UnsupportedResponseTypeException(code, description, requestId, rCode, null);
+			default:
+				throw new APIException(code, requestId, rCode, null);
```
brandur-stripe: Great use of `switch`!

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -321,6 +340,12 @@ static String createQuery(Map<String, Object> params)
 		String charge;
 	}
 
+	private static class OAuthError {
```
brandur-stripe: I think a short comment on this one indicating that it's used to receive an error before converting it to a more suitable user-facing exception would be helpful.

brandur-stripe:
Amazing work OB, thanks so much! I left a couple comments, but this looks great to me.

src/main/java/com/stripe/model/oauth/TokenResponse.java
ob-stripe: Yeah, I was being conservative here, but since this is a new feature as far as the Java library is concerned, we might as well not declare the deprecated fields at all. I'll update the PR.

src/main/java/com/stripe/net/OAuth.java
ob-stripe: It's the same error message as in the other language implementations (e.g. in [stripe-ruby](https://github.com/stripe/stripe-ruby/blob/a645a78cd028554165e44760d531f946db0ea1cb/lib/stripe/oauth.rb#L18-L24)). What do you think needs to be changed?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
ob-stripe: :+1:

ob-stripe:
Alright, I pushed a couple of commits:

- I removed the deprecated fields from the token response, and updated the test to use a model file containing a full token response (to ensure that the deprecated fields are silently ignored by the library)

- I added some Javadoc to all the OAuth methods and exception classes, including `OAuthException`, and added a short comment above `OAuthError` in `LiveStripeResponseGetter` as you suggested

ptal @brandur-stripe

src/main/java/com/stripe/net/OAuth.java
brandur-stripe: @ob-stripe Sorry, I should have been more specific. The config is called `Stripe.clientId` in Java (not `Stripe.client_id`).

brandur-stripe:
> Alright, I pushed a couple of commits:

Nice work again!

LGTM pending minor change above.

ptal @ob-stripe

src/main/java/com/stripe/net/OAuth.java
ob-stripe: 🤦‍♂️  Of course, thanks!

ob-stripe:
@brandur-stripe I fixed the error messages and squashed the commits. ptal

brandur-stripe:
Released as 5.8.0.

PR #399, opened 2017-08-29
Remove Alipay account functional tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Those tests used to create Alipay tokens, which was never officially documented and is no longer possible (Alipay payments must now use the sources API and go through a redirect flow), so I removed the failing tests.


brandur-stripe:
Good call. +1. Thank you again!

PR #406, opened 2017-10-02
Only set value on DNS cache TTL if it's been modified

### LINKED ISSUE #405: LiveStripeResponseGetter globally changes the DNS caching policy
LiveStripeResponseGetter changes the DNS caching policy from "implementation specific" (which in Java 1.7+ is 30 seconds) to "forever" (*-1*).

By default, the system property `networkaddress.cache.ttl` is `null` but after first API request it becomes `-1`. This changes globally the DNS caching and **indirectly affects other network related code**.

The following fragments in `com.stripe.net.LiveStripeResponseGetter#_rawRequest`  introduce the side-effect:

```java
try {
	originalDNSCacheTTL = java.security.Security
			.getProperty(DNS_CACHE_TTL_PROPERTY_NAME);
	// disable DNS cache
	java.security.Security
			.setProperty(DNS_CACHE_TTL_PROPERTY_NAME, "0");
} catch (SecurityException se) {
	allowedToSetTTL = false;
}
```
and at the end:
```java 
if (allowedToSetTTL) {
	if (originalDNSCacheTTL == null) {
		// value unspecified by implementation
		// DNS_CACHE_TTL_PROPERTY_NAME of -1 = cache forever
		java.security.Security.setProperty(
				DNS_CACHE_TTL_PROPERTY_NAME, "-1");
	} else {
		java.security.Security.setProperty(
				DNS_CACHE_TTL_PROPERTY_NAME, originalDNSCacheTTL);
	}
}
```
I suggest to leave the default value in case of DNS resolution strategy is not explicitly set. 
For example:
```java
try {
    originalDNSCacheTTL = java.security.Security
            .getProperty(DNS_CACHE_TTL_PROPERTY_NAME);
    // disable DNS cache
    if (originalDNSCacheTTL != null) {
        java.security.Security
                .setProperty(DNS_CACHE_TTL_PROPERTY_NAME, "0");
    }
} catch (SecurityException se) {
    allowedToSetTTL = false;
}
```
and 
```java 
if (allowedToSetTTL && originalDNSCacheTTL != null) {
    java.security.Security.setProperty(
            DNS_CACHE_TTL_PROPERTY_NAME, originalDNSCacheTTL);
}
```

brandur:
While making a request, stripe-java sets the DNS cache's TTL to zero. I
traced the commit back to 01fd2dba2ec1885f68c52355813eece3975ba3f1, but
unfortunately the author didn't leave us a shred of information as to
why they did this. I'd guess that it was was done because people were
setting long cache times and had originally resolved the host
incorrectly. I don't think it's related to security, because your TLS
certificate validation will still be active even if an attacker managed
to poison the DNS cache.

Here we take svilenivanovsmule's suggestion in #405 and only modify the
cache TTL *if* the user has overridden it. Java has pretty reasonable
defaults for this, and we shouldn't really need to touch it (unless
someone can think of a good reason why we would).

I think another very plausible option here is to just strip out all of
this code completely -- even someone who's written Java for only about
four minutes should've been able to tell that this code is not even
remotely thread-safe as written. Given that fact and no known reason why
this should be here, taking it out might be the right way to go.

Thoughts on that @ob-stripe @remi-stripe?

Fixes #405.


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -425,9 +425,18 @@ private static StripeResponse _rawRequest(
 		try {
 			originalDNSCacheTTL = java.security.Security
 					.getProperty(DNS_CACHE_TTL_PROPERTY_NAME);
-			// disable DNS cache
-			java.security.Security
-					.setProperty(DNS_CACHE_TTL_PROPERTY_NAME, "0");
+			// Disable the DNS cache if the user
```
ob-stripe: Looks like the comment got truncated.

ob-stripe:
Seems reasonable. Thanks for adding a comment so the next person investigating this doesn't have to wonder!

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
brandur-stripe: Oops. Fixing!

PR #410, opened 2017-10-09
Update Source attributes

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Updates Source attributes:
- support for new [`statement_descriptor`](https://stripe.com/docs/api#source_object-statement_descriptor) attribute
- `verification` was renamed to [`code_verification`](https://stripe.com/docs/api#source_object-code_verification)

The latter is technically a breaking change, but there are no payment methods using it just yet so it should be fine to release this as a minor version.

### DIFF
--- a/src/main/java/com/stripe/model/Source.java
+++ b/src/main/java/com/stripe/model/Source.java
@@ -13,19 +13,18 @@ import java.util.Map;
 public class Source extends ExternalAccount {
 	Long amount;
 	String clientSecret;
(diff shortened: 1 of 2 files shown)


brandur-stripe:
> The latter is technically a breaking change, but there are no payment methods using it just yet so it should be fine to release this as a minor version.

+1. Thanks for the heads up on that one!

brandur-stripe:
Released as 5.21.0.

PR #411, opened 2017-10-09
[WIP] Support for deleted fields in expansions

### LINKED ISSUE #409: Deserializing resources should work even if the resource is deleted
When you delete a resource in Stripe, it is returned as a limited object with just an `id` and `deleted: true`. In Java, those resources are custom classes called `DeletedAccount` for example.

When you retrieve a resource and expand specific fields, some of those fields might represent a deleted resource. If you use a custom deserializer though you can end up with a crash because the deserializer did not account for deleted resource.

An example is trying to retrieve a `Payout` that has been sent to a bank account (external account) deleted since then. The resource returned is not deserialized properly.

The callstack is
```
Exception in thread "main" java.lang.NullPointerException
	at com.stripe.model.ExternalAccountTypeAdapterFactory$1.read(ExternalAccountTypeAdapterFactory.java:41)
	at com.stripe.model.ExternalAccountTypeAdapterFactory$1.read(ExternalAccountTypeAdapterFactory.java:33)
	at com.google.gson.TypeAdapter$1.read(TypeAdapter.java:199)
	at com.google.gson.Gson.fromJson(Gson.java:795)
	at com.google.gson.Gson.fromJson(Gson.java:859)
	at com.google.gson.Gson$2.deserialize(Gson.java:131)
	at com.stripe.model.ExpandableFieldDeserializer.deserialize(ExpandableFieldDeserializer.java:36)
	at com.stripe.model.ExpandableFieldDeserializer.deserialize(ExpandableFieldDeserializer.java:8)
	at com.google.gson.TreeTypeAdapter.read(TreeTypeAdapter.java:58)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:93)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:172)
	at com.google.gson.Gson.fromJson(Gson.java:795)
	at com.google.gson.Gson.fromJson(Gson.java:761)
	at com.google.gson.Gson.fromJson(Gson.java:710)
	at com.google.gson.Gson.fromJson(Gson.java:682)
	at com.stripe.net.LiveStripeResponseGetter._request(LiveStripeResponseGetter.java:447)
	at com.stripe.net.LiveStripeResponseGetter.request(LiveStripeResponseGetter.java:66)
	at com.stripe.net.APIResource.request(APIResource.java:161)
	at com.stripe.model.Payout.retrieve(Payout.java:271)
```

The code causing the issue lives [here](https://github.com/stripe/stripe-java/blob/da0f8208e7505b635508d09e7bd38686ed9f9790/src/main/java/com/stripe/model/ExternalAccountTypeAdapterFactory.java#L41) though there might be other custom deserializers impacted.

ob-stripe:
r? @brandur-stripe 

Attempts to fix #409.

The API doesn't return an `object` attribute for deleted objects. Rather than trying to cast to the correct `DeletedClass`, I simply added a `deleted` boolean (which provides exactly the same information).

The downside is that we'd have to add an `isAttributeDeleted` accessor for every expandable attribute where the API can return a deleted resource. Users would then have to call this method before trying to access the expanded object.

Tagged as WIP because I'm not sure if this is the right approach for this issue. I've only added the accessor for payouts' [`destination`](https://stripe.com/docs/api#payout_object-destination) attribute for now.

### DIFF
--- a/src/main/java/com/stripe/model/Payout.java
+++ b/src/main/java/com/stripe/model/Payout.java
@@ -123,6 +123,13 @@ public class Payout extends APIResource implements MetadataStore<Payout>, HasId 
 		this.destination = new ExpandableField<ExternalAccount>(c.getId(), c);
 	}
 
+	public Boolean isDestinationDeleted() {
+		if (this.destination == null) {
+			return false;
+		}
+		return this.destination.isDeleted();
+	}
+
 	public String getFailureBalanceTransaction() {
 		if (this.failureBalanceTransaction == null) {
 			return null;
--- a/src/main/java/com/stripe/model/ExpandableField.java
+++ b/src/main/java/com/stripe/model/ExpandableField.java
@@ -7,14 +7,24 @@ package com.stripe.model;
 public class ExpandableField<T extends HasId> {
 	private String id;
 	private T expandedObject;
+	Boolean deleted;
 
 	public ExpandableField(String id, T expandedObject) {
+		this(id, expandedObject, false);
+	}
+
+	public ExpandableField(String id, T expandedObject, Boolean deleted) {
 		this.id = id;
 		this.expandedObject = expandedObject;
+		this.deleted = deleted;
 	}
 
 	public boolean isExpanded() {
 		return expandedObject != null;
+	}
+
+	public boolean isDeleted() {
+		return deleted;
 	}
 
 	public String getId() {
--- a/src/main/java/com/stripe/model/ExpandableFieldDeserializer.java
+++ b/src/main/java/com/stripe/model/ExpandableFieldDeserializer.java
@@ -30,10 +30,21 @@ public class ExpandableFieldDeserializer implements JsonDeserializer<ExpandableF
 			// Get the `id` out of the response
 			JsonObject fieldAsJsonObject = json.getAsJsonObject();
 			String id = fieldAsJsonObject.getAsJsonPrimitive("id").getAsString();
+
+			// Check if the expanded object is deleted
+			JsonElement deletedElement = fieldAsJsonObject.get("deleted");
+			if (deletedElement != null) {
+				Boolean deleted = deletedElement.getAsBoolean();
+				if (deleted) {
+					expandableField = new ExpandableField(id, null, true);
+					return expandableField;
+				}
+			}
+
 			// We need to get the type inside the generic ExpandableField to make sure fromJson correctly serializes
 			// the JsonObject:
 			Type clazz = ((ParameterizedType) typeOfT).getActualTypeArguments()[0];
-			expandableField = new ExpandableField(id, (HasId) context.deserialize(json, clazz));
+			expandableField = new ExpandableField(id, (HasId) context.deserialize(json, clazz), false);
 			return expandableField;
 		}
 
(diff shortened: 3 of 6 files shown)


src/main/java/com/stripe/model/Payout.java
```diff
@@ -123,6 +123,13 @@ public void setDestinationObject(ExternalAccount c) {
 		this.destination = new ExpandableField<ExternalAccount>(c.getId(), c);
 	}
 
+	public Boolean isDestinationDeleted() {
+		if (this.destination == null) {
+			return false;
```
ob-stripe: Note that this means that we'd return `false` when the object is not expanded, even though in that case we don't know if the resource is deleted or not. Maybe it'd be better to raise an exception, e.g. "you should not call this method if you did not expand the attribute"?

brandur-stripe:
Thanks for taking a crack at it Olivier!

Man, looking through this, it seems to me that the best option overall would be to add an `object` field to deleted objects in the backend. Does anyone know if there's a good reason that they don't have one? (My guess is that this was some kind of technical limitation.)

ob-stripe:
While it would let us unambiguously deserialize to the correct `DeletedClass`, users would still need to manually check if the resource is deleted, then call the correct method to get the correct instance (`Class` or `DeletedClass`), right?

Can you think of a way to more elegantly handle this?

remi-stripe:
@brandur-stripe @ob-stripe The whole issue here is that the API does not return the `object` property on a deleted class right? Should we simply change this on Stripe's side?

This does not sound problematic as the other library automatically cast it to the right type right? I've just confirmed with PHP but it seemed to work.

ob-stripe:
No, even if the API did return the `object` property for deleted objects, we would still need to update the `ExpandableField` implementation in the Java library to account for this. Right now `ExpandableField` expects a single class, e.g. `ExternalAccount`.

We'd need to update the implementation to handle either `ExternalAccount` or `DeletedExternalAccount`, and users would then need to check which type they get. This would be more or less the same than what I did in this PR, except that here I'm simply returning a boolean flag instead of the `DeletedClass` instance.

Alternatively, we could get rid of all the `DeletedClass`es and simply add a `deleted` attribute on deletable resources. That would make things much simpler , but it would be a breaking change.

remi-stripe:
Sorry, what I'm saying is that if we returned the `object` property you would know that it's an `ExternalAccount`. And we would then get rid _entirely_ of the `DeletedResource` concept. You would do what you do in all languages: check if deleted is true before accessing other fields. Instead of checking the class of the resource itself.

ob-stripe:
Ah, gotcha. Yes, it would be a breaking change but I think we should bite that bullet.

remi-stripe:
It would be a breaking change simply for Java though right? I think that's the only library taking this approach.

ob-stripe:
I think stripe-dotnet might do something similar, as it has a [`StripeDeleted`](https://github.com/stripe/stripe-dotnet/blob/96cdbaffa5609ab6f8c4925342247c5a8be6f0ac/src/Stripe.net/Entities/StripeDeleted.cs) class.

remi-stripe:
Ah nice catch. So there are 2 potential approaches then:
- Any deleted resource is represented by `StripeDeleted` and we need a custom deserializer in Java to match what we do in .NET
- The API changes to return the `object` property on all resources that are deleted and we remove any reference to a deleted resource in both Java and .NET and all resources now point to the normal resource with `Deleted: true`.

For what it's worth, even in Ruby/PHP/etc. we get a `StripeObject` in general for a deleted resource that is expanded. So for example if you retrieve an invoice and expand the customer and that customer was deleted, you don't get a Customer object for `invoice.customer`.

ob-stripe:
Closing in favor of #553.

PR #412, opened 2017-10-09
Support for detaching sources

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @stan-stripe 

- Adds support for detaching sources. We can't override `delete()` because the method would need to return a `DeletedExternalAccount` instance, but in the case of sources the API will simply return the detached source object. I added a `detach()` method instead (which is more descriptive of what the method does anyway) and overrode `delete()` to raise an exception with an explicit message.

- Ensures that all requests on existing sources (save for detaching sources) use the `/v1/sources/src_...` endpoint even if the source is attached to a customer (rather than using `/v1/customers/cus_.../sources/src_...` in the latter case). While both URLs would work for update requests, using the customer URL wouldn't work for verify requests.

### DIFF
--- a/src/main/java/com/stripe/model/ExternalAccount.java
+++ b/src/main/java/com/stripe/model/ExternalAccount.java
@@ -23,8 +23,16 @@ public class ExternalAccount extends APIResource implements HasId, MetadataStore
 		return id;
 	}
 
+	public void setId(String id) {
+		this.id = id;
+	}
+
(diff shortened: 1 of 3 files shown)


brandur-stripe:
Thanks for the additional explanation in the PR here OB! Looks good to me.

brandur-stripe:
Released as 5.21.0.

PR #415, opened 2017-10-25
Fix a failing test case in stripe-java's build

brandur:
This test case seems to have been trying to fetch zero items of a
collection and then comparing against the returned `data`'s count.
Unfortunately for our build's reliability, for one the parameter is
actually called `limit`, and two it only supports numbers of one or
more [1].

I believe that the only thing keeping this build running until now is
that the collection of application fees for the account could be
reliably assumed to be zero.

This patch changes what we assert on for the return list -- we move over
to a stable value; in this case the collection's URL.

[1] https://stripe.com/docs/api/curl#list_application_fees-limit

cc @mrmcduff-stripe


brandur-stripe:
Thanks @mrmcduff-stripe!

PR #417, opened 2017-10-26
Add support for listing source_transactions

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries @stan-stripe

Adds support for the `/v1/sources/src_.../source_transactions` endpoint.

`source_transaction` objects, like `source` objects, have a `type` attribute and an attribute named with the value of `type` and containing another map. This was previously handled in `SourceDeserializer`. In order to avoid duplicating that code, I turned it into a generic `TypeDataDeserializer` used by both the `Source` and `SourceTransaction` classes.


src/main/java/com/stripe/model/TypeDataDeserializer.java
```diff
@@ -0,0 +1,53 @@
+package com.stripe.model;
+
+import com.google.gson.*;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+import java.util.HashMap;
+
+public class TypeDataDeserializer<T extends HasTypeData> implements JsonDeserializer<T> {
```
brandur-stripe: Would you mind giving this a slightly more specific name and adding a doc block explaining roughly what it does to the class? I think that "type data deserializer" is just a little too generic given that this is all related to sources.

Sorry definitely a nit given that you were not the one to originally cause these issues :$

src/main/java/com/stripe/model/HasTypeData.java
```diff
@@ -0,0 +1,8 @@
+package com.stripe.model;
+
+import java.util.Map;
+
+public interface HasTypeData {
```
brandur-stripe: Good call on the interface!

I wonder if we should call this `HasSourceTypeData` or something too? (Again, will likely always be sources specific.)

brandur-stripe:
Left a couple comments, but great work as always!

ob-stripe:
I renamed the various `TypeData` references to `SourceTypeData`, and left a comment on the deserializer class as request. ptal @brandur-stripe

brandur-stripe:
Thanks Olivier! Shipping.

brandur-stripe:
Released as 5.22.0.

PR #424, opened 2017-11-10
Upgrade to Mockito 2

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Bumped the test dependencies:

- JUnit to 4.12 (unfortunately, JUnit 5 requires Java 1.8+, so we'll be stuck on the 4.x branch for a while)

- Mockito to 2.12.0

The latter fixes an "illegal reflection operation" warning when running tests with Java 9.

### DIFF
--- a/src/test/java/com/stripe/BaseStripeTest.java
+++ b/src/test/java/com/stripe/BaseStripeTest.java
@@ -113,7 +113,7 @@ public class BaseStripeTest {
 				Mockito.<Map<String, Object>>any(),
 				Mockito.<Class<T>>any(),
 				Mockito.any(APIResource.RequestType.class),
-				Mockito.any(RequestOptions.class))).thenReturn(APIResource.GSON.fromJson(response, clazz));
+				Mockito.<RequestOptions>any())).thenReturn(APIResource.GSON.fromJson(response, clazz));
 	}
 
 	public static <T> void stubOAuth(Class<T> clazz, String response) throws StripeException {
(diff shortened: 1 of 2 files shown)


src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -113,7 +113,7 @@
 				Mockito.<Map<String, Object>>any(),
 				Mockito.<Class<T>>any(),
 				Mockito.any(APIResource.RequestType.class),
-				Mockito.any(RequestOptions.class))).thenReturn(APIResource.GSON.fromJson(response, clazz));
+				Mockito.<RequestOptions>any())).thenReturn(APIResource.GSON.fromJson(response, clazz));
```
ob-stripe: Had to change this because in Mockito 2, the `any(SomeClass.class)` no longer matches `null`.

src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -123,54 +123,48 @@
...
 	}
 
-	public static class ParamMapMatcher extends ArgumentMatcher<Map<String, Object>> {
+	public static class ParamMapMatcher implements ArgumentMatcher<Map<String, Object>> {
```
ob-stripe: `ArgumentMatcher` is now an interface in Mockito 2.

src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -123,54 +123,48 @@
...
 		private Map<String, Object> other;
 
 		public ParamMapMatcher(Map<String, Object> other) {
 			this.other = other;
 		}
 
 		/* Treat null references as equal to empty maps */
-		public boolean matches(Object obj) {
-			if (obj == null) {
+		public boolean matches(Map<String,Object> paramMap) {
```
ob-stripe: The compared object is now cast to the generic type directly in the signature of the `matches` method.

src/test/java/com/stripe/BaseStripeTest.java
brandur-stripe: Makes sense. Thanks.

src/test/java/com/stripe/BaseStripeTest.java
brandur-stripe: Nice improvement.

brandur-stripe:
Awesome. Thank you for the commentary too!

PR #426, opened 2017-11-13
Autogenerated getters and setters on models

marctrem:
Inculsion of Lombok as a compile-time dependency

Signed-off-by: Marc-André Tremblay <marc@tremblay.io>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .settings
 tags
 target
+out
 .idea
 *.iml
 
--- a/build.gradle
+++ b/build.gradle
@@ -26,8 +26,7 @@ repositories {
 }
 
 dependencies {
-    compileOnly group: 'org.projectlombok', name: 'lombok', version: '1.16.16'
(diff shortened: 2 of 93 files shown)


marctrem:
Hey,

I just wanted to go through the source of a client library before interviewing. At the same time I wanted to see if I could improve something. Removal of boilerplate seemed like something great until I noticed that Lombok is not _yet_ supported in JDK9. The yet-to-be-released version of Lombok will suport it.

Also, most occurrences of _retrieve_, _update_, _reject_, _list_ and update could also be removed since they are mostly boilerplate.

No hard feelings if it gets rejected! :)

marctrem:
Oops, noticed it is a duplicate of #302.

ob-stripe:
Hey @marctrem, thanks a lot for your contribution and sorry you put in so much work before noticing we already had a PR opened for this!

There are some concerns around using Lombok (support for Java 9 is one of them, but not the most important one) and we're exploring other options to reduce the huge amount of boilerplate in the library. Do feel free to chime in in #302, user opinions are valid data points that are useful to us!

PR #428, opened 2017-11-16
Improve gradle output in Travis

ob-stripe:
r? @dpetrovics-stripe (dev-platform run, feel free to re-assign)

~By default, gradle outputs the test results in an HTML report file. This is rather inconvenient since we cannot consult the contents of the file on Travis.~

~This PR uses this project: https://github.com/mendhak/Gradle-Travis-Colored-Output to enable a pretty console output that works locally and on Travis.~

~Slight FUD: the project does not have an explicit license, so not sure if we can just pull this in. I've filed an issue on the repo: https://github.com/mendhak/Gradle-Travis-Colored-Output/issues/4~
~edit: The project's author has updated the repository to release the code under the MIT license, so I think we're good!~

See comment below: https://github.com/stripe/stripe-java/pull/428#issuecomment-346647838

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,3 +33,10 @@ dependencies {
 }
 
 apply from: 'deploy.gradle'
+
+test {
+    testLogging {
+        events "passed", "skipped", "failed"
+        exceptionFormat "full"
+    }
+}


grey-stripe:
This looks great, but a few nits:

1. When I first ran this, I saw some extra characters bleed into the test results:
image

It looks like this is due to not clearing the line after issuing the `\r`

2. When I run this with filtering for a specific suite, it will still output skipped suites:
image

It might be better to lazily output the suite name on the first test run

3. The display is a bit confusing when successes, errors, and skipped tests are next to each other:
image

Maybe add some whitespace after the failed test error message, and change the skipped test indicator to a single character-width thing so the columns still line up.

ob-stripe:
Thanks for the excellent feedback @grey-stripe!

Some of these issues would be pretty easy to fix, others not so much. In particular, the progress bar updates "bleeding" into the output seems difficult to fix.

I've updated the PR with a more basic Gradle `testLogging` configuration. The output is less fancy but is (IMO) a good compromise.

I'm a bit sad that there's no color output on Travis, but this is apparently due to Travis automatically setting `TERM=dumb` for Gradle projects, specifically to disable Gradle's fancy progress bars that are not supported by the Travis terminal. Unfortunately, this also disables colors -- it doesn't seem like Gradle supports enabling colors but not progress bars.

It's not that big of an issue: failed tests should still be easily identifiable by the presence of a stack trace.

grey-stripe:
LGTM

PR #432, opened 2017-12-05
Make charge field on Dispute objects expandable

### LINKED ISSUE #431: Dispute.charge is not Expandable
Dispute.charge is ```Expandable``` according to API docs: https://stripe.com/docs/api#dispute_object

However the current implementation expects a string only: https://github.com/stripe/stripe-java/blob/e2d2fb0fb9cf2a9a6b34e6d69e5993b08f88c6d5/src/main/java/com/stripe/model/Dispute.java#L19

Which causes deserialization with expanded charge to fail with:
```
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected STRING but was BEGIN_OBJECT
```
I have an adjustment balance transaction for a Dispute that causes the entire BalanceTransactionCollection load to fail because of that.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes #431.

~No tests because expansion itself is already tested and I've verified in the OpenAPI spec that the `charge` attribute is expandable in dispute objects.~

### DIFF
--- a/src/main/java/com/stripe/model/Dispute.java
+++ b/src/main/java/com/stripe/model/Dispute.java
@@ -16,7 +16,7 @@ public class Dispute extends APIResource implements HasId {
 	String object;
 	Long amount;
 	List<BalanceTransaction> balanceTransactions;
-	String charge;
+	ExpandableField<Charge> charge;
 	Long created;
 	String currency;
 	EvidenceSubObject evidenceSubObject; // `evidence`
@@ -77,11 +77,25 @@ public class Dispute extends APIResource implements HasId {
 	}
 
 	public String getCharge() {
-		return charge;
+		if (charge == null) {
+			return null;
+		}
+		return charge.getId();
 	}
 
-	public void setCharge(String charge) {
-		this.charge = charge;
+	public void setCharge(String chargeID) {
+		this.charge = setExpandableFieldID(chargeID, this.charge);
+	}
+
+	public Charge getChargeObject() {
+		if (this.charge == null) {
+			return null;
+		}
+		return this.charge.getExpanded();
+	}
+
+	public void setChargeObject(Charge charge) {
+		this.charge = new ExpandableField<Charge>(charge.getId(), charge);
 	}
 
 	public Long getCreated() {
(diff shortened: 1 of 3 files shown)


PeterDimov:
Thanks for the quick turnaround on this! I'm just testing effectively the same change locally and discovered that the following is also needed:
```diff
diff --git a/src/main/java/com/stripe/model/DisputeDataDeserializer.java b/src/main/java/com/stripe/model/DisputeDataDeserializer.java
index 882e685..8646f0d 100644
--- a/src/main/java/com/stripe/model/DisputeDataDeserializer.java
+++ b/src/main/java/com/stripe/model/DisputeDataDeserializer.java
@@ -17,6 +17,8 @@ public class DisputeDataDeserializer implements JsonDeserializer<Dispute> {
                Gson gson = new GsonBuilder()
                                .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
                                .registerTypeAdapter(ExpandableField.class, new ExpandableFieldDeserializer())
+                               .registerTypeAdapter(Source.class, new SourceTypeDataDeserializer<Source>())
+                               .registerTypeAdapterFactory(new ExternalAccountTypeAdapterFactory())
                                .create();
                if (json.isJsonNull()) {
                        return null;
```
Without it, Charge.source is deserialized as ExternalAccount instead of the correct dynamic type. This pattern is used a lot, so there could be lingering issues like this in other scenarios.

ob-stripe:
Hey @PeterDimov, you're entirely correct! Thanks for catching this. I've updated the PR with the fix you suggested, and a test to check to ensure that the deserialization is working correctly.

PeterDimov:
Thanks Olivier!

You probably don't need the binary files under .gradle?

The way the *Deserializer classes create a new GsonBuilder instead of using APIResource.GSON (which has all adapters configured on it) leaves the door open for more bugs like this. Using APIResource.GSON directly doesn't work as that causes recursion. I can't offer an elegant solution at the moment but maybe that's worth filing a separate issue for?

brandur-stripe:
LGTM. Thanks OB.

> The way the *Deserializer classes create a new GsonBuilder instead of using APIResource.GSON (which has all adapters configured on it) leaves the door open for more bugs like this. Using APIResource.GSON directly doesn't work as that causes recursion. I can't offer an elegant solution at the moment but maybe that's worth filing a separate issue for?

Yeah, great point — we need to get this figured out. I've logged #433 to track this problem.

PR #434, opened 2017-12-15
Upgrade Gradle, and use new plugin DSL

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Upgrade Gradle to latest version (4.4), and use [new plugin DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block) in `build.gradle`.


brandur-stripe:
Super clean! LGTM.

PR #436, opened 2017-12-18
Fix Account API request methods

### LINKED ISSUE #435: Account.delete(options) bug
```
	public DeletedAccount delete(RequestOptions options)
		return delete(null, null);
	}
```

It doesn't pass `options` to the next method.

https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Account.java#L343

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

- Fixes the `Account.delete(RequestOptions options)` method to properly pass `options`
- Adds an `Account.reject(Map<String, Object> params, RequestOptions options)` method

Fixes #435.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -322,12 +322,6 @@ public class Account extends APIResource implements HasId, MetadataStore<Account
 		return update(params, null);
 	}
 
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Ah, amazing how long that problem was in there :$ Thanks Olivier!

Issue #433, opened 2017-12-06
Consolidate GSON configurations in a recursion-friendly way


brandur-stripe:
As mentioned in #432, it's a problem that each of our deserializer classes (e.g. `ChargeRefundCollectionDeserializer`) creates its own GSON instance because it leaves the door open to forgetting to register certain type adapters that are included in `APIResource.GSON`. Currently we have this working optimistically so that hopefully any type adapters needed by any particular deserializer will be configured there.

Ideally, type adapters should all be registered in one place. Using `APIResource.GSON` has been a problem because if used from a deserializer, it will cause issues with recursion.

I'm not sure what the right answer for this is off hand, but GSON is a really big library with lots of usage, so there must be a better way to do this.

codylerum:
@brandur-stripe I took a glace at this and I don't see a path where you can get to having a single GSON instance without changing the hierarchy.

The issue comes down to using GSON to deserialize the same type as the deserializer 

https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/ChargeRefundCollectionDeserializer.java#L38

So currently you need to have a second instance of GSON which does not register the adapter for the specific type the deserializer is typed to. If you do then it just loops and stack-overflows.

So you need to have two different types for ChargeRefundCollection. Something like:

ChargeRefundCollectionAsArray
ChargeRefundCollectionAsObject

ob-stripe:
I'm closing this issue: the root cause described here hasn't been fixed, but as of stripe-java v8.0.0, the library is pinned to a specific API version. As a result, we were able to drop most of the custom deserializers that we used to have to handle the changes caused by different API versions, and so the problem is mostly gone.

PR #439, opened 2017-12-30
update to a current version of gson

codylerum:
The current version of gson used `2.2.4` is quite old (2013-05-13) and is starting to cause conflicts with tools like selenium.

Version `2.8.2` is still compatible with Java 6 and should not have any breaking changes. Just a lot of bug fixes and new features. Some that may be helpful with #433

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ repositories {
 }
 
 dependencies {
-    compile group: 'com.google.code.gson', name: 'gson', version:'2.2.4'
+    compile group: 'com.google.code.gson', name: 'gson', version:'2.8.2'
     testCompile group: 'com.google.guava', name: 'guava', version:'18.0'
     testCompile group: 'junit', name: 'junit', version:'4.12'
     testCompile group: 'org.mockito', name: 'mockito-core', version:'2.12.0'


ob-stripe:
Hi @codylerum, thanks for the PR! It looks good to me. Assigning to @brandur-stripe to confirm and release a new version soon.

brandur-stripe:
Thanks @codylerum!

Looking through [the changelog](https://github.com/google/gson/blob/master/CHANGELOG.md), it doesn't seem like anything particularly risky was introduced or changed. I'm willing to trust the test suite and say that this upgrade is fine.

> Some that may be helpful with #433

@codylerum Just OOC, do you know if there was a particular change that you think would be helpful with #433? I didn't see anything obvious like "recursive type registering support", but I might have missed something.

(Regardless, I suspect that this is something that could be resolved given a couple of solid hours spent on the problem.)

codylerum:
@brandur-stripe I didn't see anything off the top of my head but I was going to try and take a peak at it this month. It does seem odd that you have to define setup a GSON instance inside a deserializer that may be used by another GSON instance.

I don't use GSON directly anywhere however so I have very limited knowledge of it.

brandur-stripe:
> @brandur-stripe I didn't see anything off the top of my head but I was going to try and take a peak at it this month. It does seem odd that you have to define setup a GSON instance inside a deserializer that may be used by another GSON instance.

Cool! Thanks for digging and the reply. We'll get to it at some point as well :)

(This PR now released as 5.28.0.)

PR #443, opened 2018-01-11
Add support for coveralls

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds support for coveralls.io in stripe-java.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
 
 jdk:
   - openjdk7
+  - openjdk8
   - oraclejdk8
   - oraclejdk9
 
@@ -25,10 +26,8 @@ notifications:
   email:
     on_success: never
 
-sudo: false
-
 script:
-  - ./gradlew clean test
+  - ./gradlew clean cobertura coveralls
 
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
--- a/src/test/java/com/stripe/model/StandardizationTest.java
+++ b/src/test/java/com/stripe/model/StandardizationTest.java
@@ -64,6 +64,10 @@ public class StandardizationTest {
 				if (method.getName().startsWith("get")) {
 					continue;
 				}
+				// Skip internal methods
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -9,6 +9,7 @@ addons:
 
 jdk:
   - openjdk7
+  - openjdk8
```
ob-stripe: Opportunistically added openjdk8 to the list of tested JDKs.

.travis.yml
```diff
@@ -25,10 +26,8 @@ notifications:
   email:
     on_success: never
 
-sudo: false
```
ob-stripe: This was a leftover that should have been removed in #440.

.travis.yml
```diff
@@ -25,10 +26,8 @@ notifications:
...
-
 script:
-  - ./gradlew clean test
+  - ./gradlew clean cobertura coveralls
```
ob-stripe: The `cobertura` command will both run tests and generate the coverage reports. In theory, it should be possible to separate the two with the `test` and `coberturaReport` commands, but I couldn't make the latter work for some reason.

src/test/java/com/stripe/model/StandardizationTest.java
```diff
@@ -64,6 +64,10 @@ public void allNonDeprecatedMethodsTakeOptions() throws IOException, NoSuchMetho
 				if (method.getName().startsWith("get")) {
 					continue;
 				}
+				// Skip internal methods
+				if (method.getName().startsWith("_")) {
```
ob-stripe: Cobertura instruments the code to generate the coverage reports. Rather than targeting Cobertura's methods specifically, I opted to simply ignore all methods that start with an underscore -- the assumption being that such methods are very likely internal and out of scope of the standardization test.

brandur-stripe: Cool. Works for me. Thanks for explaining!

brandur-stripe:
LGTM.

PR #447, opened 2018-02-07
Support new Product+Plan API

pantera-stripe:
Add support for new Product and Plan API (`2018-02-05`)

- Added tests showing create and update of products `Product` with `type = service`
- Added tests creating a plan and product at once
- Added new fields to `Plan` and `Product`

r? @brandur-stripe @tmaxwell-stripe
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Plan.java
+++ b/src/main/java/com/stripe/model/Plan.java
@@ -20,14 +20,17 @@ public class Plan extends APIResource implements MetadataStore<Plan>, HasId {
 	Integer intervalCount;
 	Boolean livemode;
 	Map<String, String> metadata;
-	String name;
 	String nickname;
-	String statementDescriptor;
-	Integer trialPeriodDays;
 	String product;
 
 	@Deprecated
 	String statementDescription;
+	@Deprecated
+	String name;
+	@Deprecated
+	Integer trialPeriodDays;
+	@Deprecated
+	String statementDescriptor;
 
 	public String getId() {
 		return id;
@@ -101,14 +104,6 @@ public class Plan extends APIResource implements MetadataStore<Plan>, HasId {
 		this.metadata = metadata;
 	}
 
-	public String getName() {
-		return name;
-	}
-
-	public void setName(String name) {
-		this.name = name;
-	}
-
 	public String getNickname() {
 		return nickname;
 	}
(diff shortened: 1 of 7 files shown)


src/main/java/com/stripe/model/Plan.java
```diff
@@ -21,8 +21,10 @@
 	Boolean livemode;
 	Map<String, String> metadata;
 	String name;
```
ob-stripe: Can you move the `name`, `statementDescriptor` and `trialPeriodDays` attributes to the bottom of the list with a `@Deprecated` tag?

src/main/java/com/stripe/model/Plan.java
```diff
@@ -107,6 +109,14 @@ public void setName(String name) {
 		this.name = name;
```
ob-stripe: Can you move the accessors for `name`, `statementDescriptor` and `trialPeriodDays` at the bottom of the list, with a comment like e.g. `@deprecated Prefer using the `product` field (https://stripe.com/docs/upgrades#2018-02-05)`?

src/main/java/com/stripe/model/Product.java
```diff
@@ -29,6 +29,8 @@
 	SKUCollection skus;
 	Long updated;
 	String url;
+	String statementDescriptor;
```
ob-stripe: Aside from `id` and `object` which are always at the top of the list, we keep the attributes and their respective accessors sorted in alphabetical order.

src/main/java/com/stripe/model/Product.java
```diff
@@ -166,6 +168,18 @@ public void setURL(String url) {
...
+	}
+
+	public void setStatementDescriptor(String statementDescriptor) {
+		this.statementDescriptor = statementDescriptor;
+	}
+
+	public String getType() {
+		return type;
+	}
+
```
ob-stripe: You're missing a setter for `type` (not a big deal since setters are only used in tests, but we have setters ~everywhere so might as well be consistent).

src/main/java/com/stripe/model/Subscription.java
```diff
@@ -34,6 +34,7 @@
 	Double taxPercent;
 	Long trialEnd;
 	Long trialStart;
+	Long billingCycleAnchor;
```
ob-stripe: See above re. alphabetical order.

src/test/java/com/stripe/functional/ServiceProductTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.HashMap;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class ServiceProductTest extends BaseStripeFunctionalTest {
+    @Test
+    public void testProductCreate() throws StripeException {
+        Product product = Product.create(defaultServiceProductParams);
```
ob-stripe: In newer tests we usually mock the API instead of sending real requests.

Not necessarily a big issue since we're going to port the test suite to use stripe-mock eventually, so just an FYI -- feel free to ignore :)

pantera-stripe:
Made the requested changes
r? @ob-stripe

PR #448, opened 2018-02-12
Document Back

cat-stripe:
Updates the java client bindings to support document being a hash.


ob-stripe:
@cat-stripe Unfortunately the Java library does not target a specific API version, so we need to be able to support the API returning `document` both as a string or a hash.

You can take a look at how the `evidence` property is handled on the `Dispute` model:
- there's an `evidenceSubObject` attribute containing `evidence` as the "new" hash
- and a (deprecated) `evidence` attribute containing `evidence` as the "old" string
- the custom deserializer `DisputeDataDeserializer` is used to fill the correct attribute depending on what's returned by the API

I'm sorry this is a probably a little more complex than you anticipated, feel free to ping me if you need any help!

ob-stripe:
Support for `document_back` was implemented in #521, so closing this one.

PR #452, opened 2018-02-20
Use stripe-mock for tests

ob-stripe:
Updated version of #384.


ob-stripe:
I've updated the PR to add the ability to verify requests, and also stub requests (see the functional test for `FileUpload` for an example with stubbed requests).

I think we have a good base to start porting all tests to stripe-mock. To summarize, for each model class, we want:
- a functional test class that tests all API methods (using stripe-mock whenever possible, or stubbed requests otherwise)
- a model test class that tests deserialization. If the model has at least one expandable attribute, there should be two tests: one without expanding any attributes, and one expanding all attributes

If we want to be exhaustive, we should also have deserialization tests for collection classes.

@brandur-stripe @remi-stripe Any thoughts or questions? Anything I missed before we start porting all tests?

brandur-stripe:
> I think we have a good base to start porting all tests to stripe-mock. To summarize, for each model class, we want:
>
> * a functional test class that tests all API methods (using stripe-mock whenever possible, or stubbed requests otherwise)
> * a model test class that tests deserialization. If the model has at least one expandable attribute, there should be two tests: one without expanding any attributes, and one expanding all attributes
If we want to be exhaustive, we should also have deserialization tests for collection classes.

Yep, that sounds absolutely perfect. I love that it'll get us a pretty exhaustive test of all possible expansions in a pretty direct way too.

> If we want to be exhaustive, we should also have deserialization tests for collection classes.

IMO, it'd be fine just to make sure that this works for just one collection, but it's up to you :)

> @brandur-stripe @remi-stripe Any thoughts or questions? Anything I missed before we start porting all tests?

Just that this is going to be a monstrous amount of work :) The existing diff looks great.

src/test/java/com/stripe/functional/AccountTest.java
```diff
@@ -1,7 +1,11 @@
...
-import com.stripe.BaseStripeFunctionalTest;
+import static org.junit.Assert.assertNotNull;
+
+import java.io.IOException;
+import com.stripe.BaseStripeMockTest;
 import com.stripe.exception.StripeException;
+import com.stripe.net.APIResource;
 import com.stripe.model.*;
 import junit.framework.Assert;
 import org.junit.Test;
```
ob-stripe: Kind of a nit, but at some point in the future I'd like to enforce Google's Java style in the project. Can you reorder the import like in the other tests?

1. All static imports first
2. All imports in aphabetical order (`com.` / `org.` prefixes are ignored)
3. Separate import "groups" with a new line (in most cases the groups will be java, junit, then stripe)
4. Don't use `.*`, explicitly import all the classes you need.

src/test/java/com/stripe/functional/AccountTest.java
```diff
@@ -12,22 +16,132 @@
...
+			APIResource.RequestMethod.GET,
+			String.format("/v1/accounts/%s", RESOURCE_ID)
+		);
+	}
 
-		Map<String, Object> accountParams = new HashMap<String, Object>();
-		accountParams.put("limit", 3);
-		AccountCollection accountCollection = Account.all(accountParams);
+	@Test
+	public void testAccountUpdate() throws StripeException {
```
ob-stripe: Rename to `testUpdate()`

src/test/java/com/stripe/functional/AccountTest.java
```diff
@@ -12,22 +16,132 @@
...
+		account.delete();
+
+		verifyRequest(
+			APIResource.RequestMethod.DELETE,
+			String.format("/v1/accounts/%s", account.getId())
+		);
+	}
+
+	@Test
+	public void testAccountReject() throws StripeException {
```
ob-stripe: Rename to `testReject()`

src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -21,14 +19,12 @@
 import static org.mockito.Mockito.*;
```
ob-stripe: Same comment about imports as above.

src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -21,14 +19,12 @@
...
 
 public class AccountTest extends BaseStripeMockTest {
-	public static final String RESOURCE_ID = "acct_123";
-
 	@Test
 	public void testDeserialize() throws StripeException, IOException {
-		String json = getResourceAsString("account.json");
+		String json = getResourceAsString("/api_fixtures/account.json");
```
ob-stripe: Rather than deserializing from an embedded JSON fixture, I'd rather we use stripe-mock whenever possible.

src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -21,14 +19,12 @@
...
 		Account acc = APIResource.GSON.fromJson(json, Account.class);
 
-		assertEquals("acct_1032D82eZvKYlo2C", acc.getId());
+		assertEquals("acct_1234", acc.getId());
```
ob-stripe: I don't think it's necessary to check the value of every field (especially if you're using stripe-mock instead of an embedded fixture file). Mostly what we're testing here is that `fromJson` does not raise an exception.

In the other deserialization tests, I used just two asserts: `assertNotNull(resource)` + `assertNotNul(resource.getId())` to check for the unlikely case where deserialization would not raise an exception but return `null`, or return an instance with `null` values in attributes.

For models with submodels (e.g. `AccountPayoutSchedule`), we should create a separate deserialization test for the submodel, e.g. in `model/AccountPayoutScheduleTest.java`:
```java
String accountData = getFixture("/v1/accounts/acct_123");
String data = getDataAt(accountData, "payout_schedule");
AccountPayoutSchedule object = APIResource.GSON.fromJson(data, AccountPayoutSchedule.class);
assertNotNull(object);
assertNotNull(object.getInterval());
```

(cf. `ChargeOutcomeTest.java` in the first commit)

src/test/java/com/stripe/model/AccountTest.java
remi-stripe: I disagree partially because stripe-mock return non Custom accounts which won't deserialize a lot of the more complex sub-elements like `legal_entity` and `external_accounts`

remi-stripe: I'm happy to do both but that deserialization via stripe-mock happens in most of the other tests since we retrieve an account before reject/update/etc.

src/test/java/com/stripe/model/AccountTest.java
remi-stripe: That seems pretty painful to maintain for the Account resource and all its sub-resources and various versions.

ob-stripe: They are separate classes in the code, so they should be tested separately. It will certainly be a bit painful to cover all models (mostly due to the fact that our coverage is pretty spotty right now), but once that's done, maintenance should be minimal.

remi-stripe:
@ob-stripe PTAL. I addressed all your comments (I think) and changed the Account model tests to only use stripe-mock's fixture. Then I moved LegalEntityTest to the new framework

ob-stripe:
👍  Nice, looks great! Thanks for also deleting the old fixtures that are no longer used after you updated the tests.

src/test/java/com/stripe/functional/IdempotentTest.java
```diff
@@ -1,68 +0,0 @@
-package com.stripe.functional;
```
remi-stripe: @ob-stripe We need to look into moving those tests to a different level to make sure we are actively verifying this works but it's the wrong way to test idempotency

src/test/java/com/stripe/functional/StripeResponseTest.java
```diff
@@ -35,12 +35,13 @@ public void testResponseIncluded() throws
...
         .build();
-    Customer cus = Customer.create(defaultCustomerParams, requestOptions);
+    Customer cus = Customer.create(null, requestOptions);
     cus = Customer.retrieve(cus.getId(), requestOptions);
     StripeResponse resp = cus.getLastResponse();
     assertThat(resp, instanceOf(StripeResponse.class));
     assertEquals(200, resp.code());
-    assertEquals(idempotencyKey, resp.idempotencyKey());
+    // TODO: ob add this to stripe-mock
+    // assertEquals(idempotencyKey, resp.idempotencyKey());
```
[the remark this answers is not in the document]
brandur-stripe: Well, it still doesn't do anything too smart with idempotency keys, but I just patched stripe-mock to reflect an idempotency key that you sent back to you like the real API does. Patch in https://github.com/stripe/stripe-mock/pull/77 and release in 0.17.0.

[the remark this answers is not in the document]
remi-stripe: Damn you're so fast, I had barely started the code lol.

src/test/java/com/stripe/model/ExternalAccountTest.java
```diff
@@ -48,52 +37,44 @@ public void testUnknownExternalAccountRetrieval() throws StripeException, IOExce
...
-
-    assertEquals(true, bankAccount instanceof ExternalAccount);
+    String data = getResourceAsString("/api_fixtures/customer_with_external_account.json");
+    Customer cus = APIResource.GSON.fromJson(data, Customer.class);
+    assertNotNull(cus);
+    ExternalAccount ea = cus.getSources().getData().get(0);
 
-    stubNetwork(ExternalAccount.class,
-        "{\"id\": \"extacct_123\", \"object\": \"unknown_external_account\"}");
+    assertEquals(true, ea instanceof ExternalAccount);
```
ob-stripe: You can use `assertTrue` here.

ob-stripe:
Okay, I think this is ~ready.

Coverage numbers are still fairly low primarily because of the large number of untested deprecated methods (I've tried configuring the Cobertura plugin to ignore those, but it doesn't seem to be working too well) and some setter methods (which should also be excluded from coverage but I haven't looked into it yet).

Note that there is one non-test change in this PR: I added a new method `Stripe.overrideUploadBase()` in order to be able to use stripe-mock to test file uploads. We already had similar methods for the API and for OAuth anyway.

src/test/java/com/stripe/functional/ApplicationFeeTest.java
```diff
@@ -1,22 +1,59 @@
...
-    ApplicationFeeCollection fees = ApplicationFee.list(listParams);
-    assertEquals("/v1/application_fees", fees.getURL());
+    listParams.put("limit", 1);
+
+    ApplicationFeeCollection resources = ApplicationFee.list(listParams);
+
+    assertNotNull(resources);
+    verifyRequest(
+        APIResource.RequestMethod.GET,
+        String.format("/v1/application_fees")
```
remi-stripe: should we verify that params are passed?

src/test/java/com/stripe/functional/BalanceTransactionTest.java
```diff
@@ -1,34 +1,43 @@
...
+  public void testList() throws StripeException {
+    Map<String, Object> listParams = new HashMap<String, Object>();
+    listParams.put("limit", 1);
+
+    BalanceTransactionCollection resources = BalanceTransaction.list(listParams);
+
+    assertNotNull(resources);
+    verifyRequest(
+        APIResource.RequestMethod.GET,
+        String.format("/v1/balance/history")
```
remi-stripe: should we verify that params are sent? (won't repeat it for all retrieve, might be worth fixing all)

src/test/java/com/stripe/functional/BankAccountTest.java
```diff
@@ -0,0 +1,166 @@
...
+    verifyRequest(
+        APIResource.RequestMethod.POST,
+        String.format("/v1/customers/%s/sources", customer.getId()),
+        params
+    );
+  }
+
+  @Test
+  public void testRetrieve() throws StripeException {
+    final Customer customer = Customer.retrieve(CUSTOMER_ID);
```
remi-stripe: why do we have `final` here and not on any of the other examples? Should we fix them all? (annoying I know but consistency is king)

src/test/java/com/stripe/functional/BankAccountTest.java
```diff
@@ -0,0 +1,166 @@
...
+    ExternalAccountCollection stubbedCollection = new ExternalAccountCollection();
+    List<ExternalAccount> stubbedData = new LinkedList<ExternalAccount>();
+    stubbedData.add(stubbedBankAccount);
+    stubbedCollection.setData(stubbedData);
+    stubRequest(
+        APIResource.RequestMethod.GET,
+        String.format("/v1/customers/%s/sources", customer.getId()),
+        params,
+        ExternalAccountCollection.class,
+        stubbedCollection.toJson()
```
remi-stripe: I don't really understand this one. Why not a fixture that is a list already? I think that's what I did for other tests.

src/test/java/com/stripe/functional/CardTest.java
```diff
@@ -1,20 +1,150 @@
...
+public class CardTest extends BaseStripeTest {
+  public static final String CUSTOMER_ID = "cus_123";
+  public static final String CARD_ID = "card_123";
+
+  private Card getCardFixture(Customer customer) throws IOException, StripeException {
+    final Card card = APIResource.GSON.fromJson(
+        getResourceAsString("/api_fixtures/card.json"), Card.class);
+    card.setCustomer(customer.getId());
+
+    return card;
```
remi-stripe: you likely want a comment to explain that this is because stripe-mock is mean to you.

src/test/java/com/stripe/functional/CardTest.java
```diff
@@ -1,20 +1,150 @@
...
+    Map<String, Object> params = new HashMap<String, Object>();
+    params.put("source", "btok_123");
+
+    // stripe-mock returns a BankAccount instance instead of a Card
+    stubRequest(
+        APIResource.RequestMethod.POST,
+        String.format("/v1/customers/%s/sources", customer.getId()),
+        params,
+        Card.class,
+        getResourceAsString("/api_fixtures/card.json")
```
remi-stripe: why not `getCardFixture`?

src/test/java/com/stripe/functional/CardTest.java
```diff
@@ -1,20 +1,150 @@
...
+  public void testRetrieve() throws IOException, StripeException {
+    final Customer customer = Customer.retrieve(CUSTOMER_ID);
+
+    // stripe-mock returns a BankAccount instance instead of a Card
+    stubRequest(
+        APIResource.RequestMethod.GET,
+        String.format("/v1/customers/%s/sources/%s", customer.getId(), CARD_ID),
+        null,
+        Card.class,
+        getResourceAsString("/api_fixtures/card.json")
```
remi-stripe: same as above

src/test/java/com/stripe/functional/CardTest.java
```diff
@@ -1,20 +1,150 @@
...
+    List<ExternalAccount> stubbedData = new LinkedList<ExternalAccount>();
+    stubbedData.add(stubbedCard);
+    stubbedCollection.setData(stubbedData);
+    stubRequest(
+        APIResource.RequestMethod.GET,
+        String.format("/v1/customers/%s/sources", customer.getId()),
+        params,
+        ExternalAccountCollection.class,
+        stubbedCollection.toJson()
+    );
```
remi-stripe: same as with the bank account, why not a real fixture?

src/test/java/com/stripe/functional/CouponTest.java
```diff
@@ -1,128 +1,89 @@
...
-    Map<String, Object> customerWithCouponParams = new HashMap<String, Object>();
-    customerWithCouponParams.put("coupon", coupon.getId());
-    Customer customer = Customer.create(customerWithCouponParams,
-        Stripe.apiKey);
-    assertEquals(customer.getDiscount().getCoupon().getId(), coupon.getId());
-
-    customer.deleteDiscount();
-    assertNull(Customer.retrieve(customer.getId()).getDiscount());
-  }
+    resource.delete();
```
remi-stripe: should we test `CouponDeleted` like you did for Card (if so, applies to all delete)

src/test/java/com/stripe/functional/FeeRefundTest.java
```diff
@@ -0,0 +1,85 @@
...
+        String.format("/v1/application_fees/%s/refunds/%s", fee.getId(), refund.getId()),
+        params
+    );
+  }
+
+  @Test
+  public void testList() throws StripeException {
+    final ApplicationFee fee = ApplicationFee.retrieve(FEE_ID);
+
+    final FeeRefundCollection reversals = fee.getRefunds().list(null);
```
remi-stripe: rename the variable to `resources` instead (bad copy paste I think)

src/test/java/com/stripe/functional/IdempotentTest.java
remi-stripe: bumping this just in case

src/test/java/com/stripe/functional/LoginLinkTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    stubRequest(
+        APIResource.RequestMethod.POST,
+        String.format("/v1/accounts/%s/login_links", account.getId()),
+        null,
+        LoginLink.class,
+        getResourceAsString("/api_fixtures/login_link.json")
+    );
+
+    final LoginLink link = account.getLoginLinks().create();
```
remi-stripe: I wonder if it'd be cleaner to have an account fixture with the `login_links` collection instead

src/test/java/com/stripe/functional/OrderReturnTest.java
```diff
@@ -0,0 +1,39 @@
...
+        String.format("/v1/order_returns/%s", RESOURCE_ID)
+    );
+  }
+
+  @Test
+  public void testList() throws StripeException {
+    final OrderReturnCollection orderReturns = OrderReturn.list(null);
+
+    assertNotNull(orderReturns);
+    assertEquals(1, orderReturns.getData().size());
```
remi-stripe: should we do this assert on other list endpoints (I don't think I've seen it yet)
also why do we expect `1` when we're not even passing a limit?

src/test/java/com/stripe/functional/OrderTest.java
```diff
@@ -0,0 +1,106 @@
...
+    verifyRequest(
+        APIResource.RequestMethod.POST,
+        String.format("/v1/orders/%s", order.getId()),
+        params
+    );
+  }
+
+  @Test
+  public void testList() throws StripeException {
+    final OrderCollection orders = Order.list(null);
```
remi-stripe: should be called `resources`

src/test/java/com/stripe/functional/OrderTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class OrderTest extends BaseStripeTest {
+  public static final String RESOURCE_ID = "or_123";
+
+  @Test
+  public void testCreate() throws StripeException {
+    Map<String, Object> params = new HashMap<String, Object>();
+    params.put("currency", "usd");
+
+    final Order order = Order.create(params);
```
remi-stripe: should be called `resource`

src/test/java/com/stripe/functional/PlanTest.java
```diff
@@ -2,205 +2,91 @@
...
+    params.put("interval", "month");
+    params.put("name", "Sapphire Elite");
 
-  @Test
-  public void testCustomerCreateWithPlan() throws StripeException {
-    Plan plan = Plan.create(getUniquePlanParams());
-    Customer customer = createDefaultCustomerWithPlan(plan);
-    assertEquals(customer.getSubscriptions().getData().get(0).getPlan().getId(), plan.getId());
-  }
+    Plan plan = Plan.create(params);
```
remi-stripe: should be called `resource`

src/test/java/com/stripe/functional/ProductTest.java
```diff
@@ -0,0 +1,106 @@
...
+    params.put("name", "Test Name");
+    params.put("description", "This is a description");
+    params.put("caption", "This is a caption");
+    params.put("attributes", attributes);
+    params.put("url", "http://example.com");
+    params.put("shippable", true);
+    params.put("package_dimensions", packageDimensions);
+    params.put("type", "good");
+
+    final Product product = Product.create(params);
```
remi-stripe: should be called `resource`

src/test/java/com/stripe/functional/ReversalTest.java
```diff
@@ -0,0 +1,85 @@
...
+        String.format("/v1/transfers/%s/reversals/%s", transfer.getId(), reversal.getId()),
+        params
+    );
+  }
+
+  @Test
+  public void testList() throws StripeException {
+    final Transfer transfer = Transfer.retrieve(TRANSFER_ID);
+
+    final TransferReversalCollection reversals = transfer.getReversals().list(null);
```
remi-stripe: should be called `resources`

src/test/java/com/stripe/functional/SKUTest.java
```diff
@@ -0,0 +1,103 @@
+package com.stripe.functional;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+import com.stripe.BaseStripeTest;
+import com.stripe.exception.StripeException;
+import com.stripe.model.DeletedSKU;
+import com.stripe.model.SKU;
```
remi-stripe: flagging for next major version, are we renaming to Sku? 😱

src/test/java/com/stripe/functional/SKUTest.java
```diff
@@ -0,0 +1,103 @@
...
+    assertNotNull(deletedSku);
+    verifyRequest(
+        APIResource.RequestMethod.DELETE,
+        String.format("/v1/skus/%s", sku.getId())
+    );
+  }
+
+  @Test
+  public void testList() throws StripeException {
+    SKUCollection skus = SKU.list(null);
```
remi-stripe: should be called `resources`

src/test/java/com/stripe/functional/SKUTest.java
```diff
@@ -0,0 +1,103 @@
...
+    Map<String, Object> params = new HashMap<String, Object>();
+    params.put("active", true);
+    params.put("attributes", attributes);
+    params.put("price", 499);
+    params.put("currency", "usd");
+    params.put("inventory", inventory);
+    params.put("product", "prod_123");
+    params.put("image", "http://example.com/foo.png");
+
+    SKU sku = SKU.create(params);
```
remi-stripe: should be called `resource`

src/test/java/com/stripe/functional/SourceTest.java
```diff
@@ -1,40 +1,137 @@
...
+  @Test
+  public void testCreate() throws StripeException {
+    Map<String, Object> owner = new HashMap<String, Object>();
+    owner.put("email", "jenny.rosen@example.com");
+    Map<String, Object> params = new HashMap<String, Object>();
+    params.put("type", "ach_credit_transfer");
+    params.put("currency", "usd");
+    params.put("owner", owner);
+
+    Source source = Source.create(params);
```
remi-stripe: should be called resource

src/test/java/com/stripe/functional/SubscriptionItemTest.java
```diff
@@ -1,106 +1,92 @@
...
-    SubscriptionItem subscriptionItem = createDefaultSubscriptionItem(subscription);
-
-    assertEquals(subscriptionItem.getPlan().getName(), "J Bindings Plan");
+  public void testCreate() throws StripeException {
+    Map<String, Object> params = new HashMap<String, Object>();
+    params.put("plan", "plan_123");
+    params.put("subscription", "cus_123");
+    params.put("quantity", 99);
+
+    SubscriptionItem subscriptionItem = SubscriptionItem.create(params);
```
remi-stripe: should be called `resource`

src/test/java/com/stripe/functional/SubscriptionTest.java
```diff
@@ -3,220 +3,108 @@
...
-    Map<String, Object> subUpdateParams = new HashMap<String, Object>();
-    subUpdateParams.put("plan", plan2.getId());
-    sub = sub.update(subUpdateParams);
-    assertEquals(plan2.getId(), sub.getPlan().getId());
-
-    // Cancel
-    sub = sub.cancel(null);
-    assertNotNull(sub.getCanceledAt());
-  }
+    Subscription subscription = Subscription.create(params);
```
remi-stripe: should be called resource.

src/test/java/com/stripe/functional/TimeoutTest.java
```diff
@@ -1,23 +0,0 @@
-package com.stripe.functional;
```
remi-stripe: restarting that I think we should find a way to test this. Maybe stripe-mock could have an endpoint that explicitly waits X seconds to respond (as a parameter) and we'd put a timeout that is X - 1 ?

src/test/java/com/stripe/functional/TopupTest.java
```diff
@@ -0,0 +1,80 @@
...
+  @Test
+  public void testCreate() throws StripeException {
+    Map<String, Object> params = new HashMap<String, Object>();
+    params.put("amount", 100);
+    params.put("currency", "usd");
+    params.put("source", "src_123");
+    params.put("description", "creating a topup");
+    params.put("statement_descriptor", "topup");
+
+    Topup topup = Topup.create(params);
```
remi-stripe: should be named `resource`

src/test/java/com/stripe/functional/TokenTest.java
```diff
@@ -1,82 +1,49 @@
...
+  public void testCreate() throws StripeException {
+    Calendar now = Calendar.getInstance();
+    Map<String, Object> card = new HashMap<String, Object>();
+    card.put("number", "4242424242424242");
+    card.put("exp_month", now.get(Calendar.MONTH));
+    card.put("exp_year", now.get(Calendar.YEAR) + 1);
+    Map<String, Object> params = new HashMap<String, Object>();
+    params.put("card", card);
+
+    Token token = Token.create(params);
```
remi-stripe: should be named `resource`

src/test/java/com/stripe/functional/TransferTest.java
```diff
@@ -1,103 +1,78 @@
...
-    Transfer created = Transfer.create(getTransferParams());
-    try {
-      // we expect an InvalidRequestException here (caught by JUnit),
-      // because in test mode, transfers are automatically sent.
-      created.cancel(RequestOptions.getDefault());
-      Assert.fail();
-    } catch (InvalidRequestException ire) {
-      // do nothing
-    }
+    Transfer transfer = Transfer.create(params);
```
remi-stripe: should be named `resource`

src/test/java/com/stripe/functional/UsageRecordTest.java
```diff
@@ -1,65 +1,39 @@
...
+    params.put("quantity", 10);
+    params.put("subscription_item", SUBCRIPTION_ITEM_ID);
+    params.put("timestamp", System.currentTimeMillis() / 1000L);
+
+    UsageRecord resource = UsageRecord.create(params, null);
+
+    // UsageRecord.create() has a non-standard behavior: it uses the `subscription_item` element
+    // in the params map to form the endpoint's URL. We need to remove it from the map because
+    // it won't be sent as a URL-encoded parameter.
+    params.remove("subscription_item");
```
remi-stripe: flagging that we should fix this in the next major version and move back to a parameter for the id

src/test/java/com/stripe/model/AccountTest.java
```diff
@@ -1,195 +1,19 @@
...
-
-    String jsonLink = resource("login_link.json");
-    stubNetwork(LoginLink.class, jsonLink);
-    acc.getLoginLinks().create();
-    verifyPost(LoginLink.class, "https://api.stripe.com/v1/accounts/acct_EXPRESS/login_links", (RequestOptions) null);
+  public void testDeserialize() throws Exception {
+    String data = getFixture("/v1/accounts/acct_123");
+    Account resource = APIResource.GSON.fromJson(data, Account.class);
+    assertNotNull(resource);
+    assertNotNull(resource.getId());
```
remi-stripe: we likely want specific edge-case as fixtures to ensure deserialization works as needed.

src/test/java/com/stripe/model/BalanceTransactionTest.java
```diff
@@ -1,32 +1,30 @@
...
-  public void testDeserializeExpansions() throws StripeException, IOException {
-    String json = resource("balance_transaction_expansions.json");
-    BalanceTransaction transaction = APIResource.GSON.fromJson(json, BalanceTransaction.class);
-    assertEquals("txn_1AF5AmKac2qaDcCZm8tBQt6I", transaction.getId());
-
-    Dispute dp = transaction.getSourceObjectAs();
-    assertNotNull(dp);
-    assertEquals("dp_1AF5AmKac2qaDcCZHvOjyrDo", dp.getId());
+  public void testDeserializeExpansions() throws Exception {
+    // TODO: Figure out why stripe-mock does not expand source when asked
```
remi-stripe: likely worth looking/flagging and then removing the TODO

src/test/java/com/stripe/model/ChargeOutcomeTest.java
```diff
@@ -1,51 +1,22 @@
...
-import com.stripe.exception.StripeException;
 import com.stripe.net.APIResource;
 
-import java.io.IOException;
-
+import org.junit.Ignore;
 import org.junit.Test;
 
+// stripe-mock doesn't return an outcome, disable the test for now
+@Ignore
```
remi-stripe: we should write a test for this

src/test/java/com/stripe/model/DeserializerTest.java
```diff
@@ -1,139 +0,0 @@
-package com.stripe.model;
```
remi-stripe: assuming it's normal that this is entirely removed?

src/test/java/com/stripe/model/EphemeralKeyTest.java
```diff
@@ -1,104 +1,26 @@
...
-    assertEquals(key.getSecret(), "ek_test_hunter2");
-    assertEquals(key.getAssociatedObjects().size(), 1);
-    assertEquals(key.getAssociatedObjects().get(0).getType(), "customer");
-    assertEquals(key.getAssociatedObjects().get(0).getId(), "cus_234");
+  public void testDeserialize() throws Exception {
+    String json = getResourceAsString("/api_fixtures/ephemeral_key.json");
+    EphemeralKey resource = APIResource.GSON.fromJson(json, EphemeralKey.class);
+    assertNotNull(resource);
+    assertEquals("ephkey_123", resource.getId());
+    assertEquals("ephemeral_key", resource.getObject());
```
remi-stripe: why not use stripe-mock?

src/test/java/com/stripe/model/LegalEntityTest.java
```diff
@@ -1,39 +1,19 @@
...
-    assertEquals(new Integer(4), le.getDob().getDay());
-    assertEquals(new Integer(4), le.getDob().getMonth());
-    assertEquals(new Integer(1969), le.getDob().getYear());
-    assertEquals(new LinkedList<Object>(), le.getAdditionalOwners());
-    assertEquals("verified", le.getVerification().getStatus());
-    assertEquals(null, le.getVerification().getDocument());
-    assertEquals(null, le.getVerification().getDetails());
+    assertNotNull(le);
+    // TODO: Figure out how to test various versions of legal_entity
+    // such as with additional_owners, with expanded verification doc, etc.
```
remi-stripe: flagging there's a big/important todo here

src/test/java/com/stripe/model/UsageRecordTest.java
```diff
@@ -1,61 +1,19 @@
...
-    long unixTime = System.currentTimeMillis() / 1000L;
-    params.put("timestamp", unixTime);
-    params.put("livemode", true);
-
-    UsageRecord.create(params, null);
+  public void testDeserialize() throws Exception {
+    String data = getResourceAsString("/api_fixtures/usage_record.json");
+    UsageRecord resource = APIResource.GSON.fromJson(data, UsageRecord.class);
+    assertNotNull(resource);
+    assertNotNull(resource.getId());
```
remi-stripe: why not use stripe-mock?

src/test/resources/api_fixtures/file_upload.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "file_1BrE64KCFFPkgtRhvXK81rmT",
```
remi-stripe: random but let's not put a real id!

src/test/resources/api_fixtures/card.json
```diff
@@ -0,0 +1,26 @@
+{
+  "id": "card_1CaO622eZvKYlo2C22sr7okI",
```
remi-stripe: card_123

src/test/resources/api_fixtures/file_upload_list.json
```diff
@@ -0,0 +1,17 @@
+{
+  "object": "list",
+  "data": [
+    {
+      "id": "file_1BrE64KCFFPkgtRhvXK81rmT",
```
remi-stripe: file_123

src/test/resources/api_fixtures/source_detached.json
```diff
@@ -0,0 +1,37 @@
+{
+  "id": "src_1BoJ2NKrZ43eBVAb7Qj7bnQL",
```
remi-stripe: src_123

src/test/resources/api_fixtures/source_detached.json
```diff
@@ -0,0 +1,37 @@
...
+  "object": "source",
+  "ach_credit_transfer": {
+      "account_number": "test_52796e3294dc",
+      "bank_name": "TEST BANK",
+      "fingerprint": null,
+      "routing_number": "110000000",
+      "swift_code": "TSTEZ122"
+  },
+  "client_secret": "src_client_secret_CCiTVxSUWh1dzjv2qPyYuz55",
```
remi-stripe: src_..._123

src/main/java/com/stripe/Stripe.java
```diff
@@ -10,9 +10,9 @@
   private static final int DEFAULT_CONNECT_TIMEOUT = 30 * 1000;
   private static final int DEFAULT_READ_TIMEOUT = 80 * 1000;
 
-  public static final String UPLOAD_API_BASE = "https://uploads.stripe.com";
   public static final String LIVE_API_BASE = "https://api.stripe.com";
   public static final String CONNECT_API_BASE = "https://connect.stripe.com";
+  public static final String UPLOAD_API_BASE = "https://files.stripe.com";
```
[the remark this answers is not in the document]
ob-stripe: Yes, both `files.stripe.com` and `uploads.stripe.com` point to the same IP.

src/test/java/com/stripe/functional/CardTest.java
ob-stripe: `stubRequest` wants a `Class` and a `String` containing the JSON. `getCardFixture()` returns a `Card` instance. We could pass `getCardFixture().toJson()` but that seems a bit redunant (we'd deserialize the JSON into a `Card` only to reserialize it to JSON).

src/test/java/com/stripe/functional/CardTest.java
ob-stripe: same as above :)

src/test/java/com/stripe/functional/BankAccountTest.java
ob-stripe: Because I'd be mildly annoyed to have a fixture file with a bank account, and a different fixture file with the same bank account wrapped in a list.

Hopefully stripe-mock will be able to handle this case sooner rather than later and those fixtures will become unnecessary anyway.

src/test/java/com/stripe/functional/CardTest.java
ob-stripe: same as above

src/test/java/com/stripe/functional/CouponTest.java
ob-stripe: Probably. The important part of the test is really `verifyRequest`, but might as well be consistent.

src/test/java/com/stripe/functional/FeeRefundTest.java
ob-stripe: Ah yeah, this one should be `refunds`. (I understand that using `resource` / `resources` everywhere makes the initial effort of porting/writing tests easier, but I prefer explicit names.)

src/test/java/com/stripe/functional/LoginLinkTest.java
ob-stripe: No strong opinion on this -- I don't think it matters too much how the fixtures are generated as long as it doesn't impede readability.

src/test/java/com/stripe/model/EphemeralKeyTest.java
ob-stripe: Because ephemeral keys can only be created, not retrieved, and require special parameters for the API version. Since this test is specifically for the deserialization, I didn't want to introduce additional complexity by using stripe-mock here.

src/test/java/com/stripe/model/UsageRecordTest.java
ob-stripe: Like `EphemeralKey`, `UsageRecord` cannot be retrieved.

src/test/java/com/stripe/functional/BankAccountTest.java
ob-stripe: Immutability is good, so ideally all variables that can be declared as `final` _should_ be. I think there's a linter rule that require that variables whose declaration is more than X lines away from their first use be declared as final, which is what prompted me to start using it in some places, and then thought that we might as well use it everywhere.

In practice, I don't think it's worth stressing about too much.

src/test/java/com/stripe/functional/AccountTest.java
```diff
@@ -1,36 +1,145 @@
...
+    resource.reject(params);
+
+    verifyRequest(
+        APIResource.RequestMethod.POST,
+        String.format("/v1/accounts/%s/reject", resource.getId()),
+        params
+    );
+  }
+
+  // TODO: Decide if sub-resources tests should live here.
```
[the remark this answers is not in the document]
ob-stripe: So I kind of decided that already by adding `CardTest` and `BankAccountTest`, though currently they only test the customer sources case. We can add a few more tests specifically for `instanceURL` though.

src/test/java/com/stripe/functional/ApplicationFeeTest.java
remi-stripe: but there are params here right?

src/test/java/com/stripe/functional/BalanceTransactionTest.java
remi-stripe: same as above, we are explicitly passing `limit` but not verifying it.

src/test/java/com/stripe/functional/BankAccountTest.java
remi-stripe: FWIW I did that for other resources though.

src/test/java/com/stripe/functional/FeeRefundTest.java
remi-stripe: > I understand that using resource / resources everywhere makes the initial effort of porting/writing tests easier, but I prefer explicit names.

I disagree, that avoids the copy/paste fails we have in most of our libraries. But would be great to be consistent since we already went through so much trouble to make everything consistent.

src/test/java/com/stripe/functional/SKUTest.java
ob-stripe: We can make an exception for SKU.

brandur-stripe:
Can you maybe kick this back to me once Remi's feedback is satisfied and I'll give it one more pass over? Just looking at the general structure, I suspect that I won't have many qualms — the broad strokes all look totally right. I'm sure we'll find more to fix, but we can easily offload that to other PRs. This one is likely to become stale quickly.

ob-stripe:
Alright, I probably missed more than a few things, but:

- all tests now use `final` whenever possible
- the tests should be a lot more consistent
- params are tested whenever we send params
- I removed unnecessary fixtures
- all fixtures now use fake IDs (`smth_123`)
- I added a `charge_outcome.json` fixture file and enabled the `ChargeOutcome` deserialization test
- I added some janky but working tests for connection and read timeouts in `RequestOptionsTest`

ptal @remi-stripe

src/test/java/com/stripe/functional/RequestOptionsTest.java
```diff
@@ -0,0 +1,77 @@
...
+    assertEquals(idempotencyKey, response.headers().get("Idempotency-Key"));
+  }
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void testConnectTimeout() throws IOException, StripeException {
+    // Kind of terrible, but let's test connection timeouts with a random external IP.
+    Stripe.overrideApiBase("https://254.254.254.254");
```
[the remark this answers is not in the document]
ob-stripe: Yeah, actually, this wasn't a great idea (and also kind of rude to whoever owns this IP). I changed it to `api.stripe.com`. Hopefully Travis' servers are more than 1 millisecond away from Stripe's 😐

src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -1,170 +1,481 @@
...
+      String response) throws StripeException {
+    String url;
+    if (path.startsWith("/")) {
+      url = String.format("%s%s", Stripe.getApiBase(), path);
+    } else {
+      url = path;
+    }
+
+    Mockito
+        .doReturn(APIResource.GSON.fromJson(response, clazz))
```
brandur-stripe: Is this indentation right? Seems weird that the style is two spaces everywhere else, but then four spaces when chaining method calls like this.

brandur-stripe:
Wow, WHAT A DIFF! Thanks for doing so much work.

Okay, admittedly I only took a very high-level pass to see if I could spot anything untoward, but this LGTM. Very much looking forward to getting this in.

ptal @ob-stripe

PR #453, opened 2018-02-20
Add support for /v1/topups/ endpoints

apakulov-stripe:
This change adds support for /v1/topups endpoint.

r? @jkakar-stripe @ccontinanza-stripe @chellman-stripe @kenneth-stripe @miguel-stripe
r? @stripe/api-libraries

I dropped `com.stripe.functional.Topup` because `stripe-java` doesn't use `stripe-mock` for functional testing and relies on publicly exposed endpoints, but at the moment `/v1/topups` is gated.

### DIFF
--- a/src/main/java/com/stripe/model/Topup.java
+++ b/src/main/java/com/stripe/model/Topup.java
@@ -1,0 +1,213 @@
+package com.stripe.model;
+
+import com.stripe.exception.*;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Map;
+
+public class Topup extends APIResource implements MetadataStore<Topup>, HasId {
+    String id;
+    String object;
+    Integer amount;
+    ExpandableField<BalanceTransaction> balanceTransaction;
+    Long created;
+    String currency;
+    String description;
+    Long expectedAvailabilityDate;
+    String failureCode;
+    String failureMessage;
+    Boolean livemode;
+    Map<String, String> metadata;
+    Source source;
+    String statementDescriptor;
+    String status;
+
+    @Override
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public String getObject() {
+        return object;
+    }
+
+    public void setObject(String object) {
+        this.object = object;
+    }
+
+    public Integer getAmount() {
+        return amount;
+    }
+
+    public void setAmount(Integer amount) {
+        this.amount = amount;
+    }
+
+    public String getBalanceTransaction() {
+        if (this.balanceTransaction == null) {
+            return null;
+        }
+        return this.balanceTransaction.getId();
+    }
+
+    public void setBalanceTransaction(String balanceTransactionID) {
+        this.balanceTransaction = setExpandableFieldID(balanceTransactionID, this.balanceTransaction);
+    }
+
+    public BalanceTransaction getBalanceTransactionObject() {
+        if (this.balanceTransaction == null) {
+            return null;
+        }
+        return this.balanceTransaction.getExpanded();
+    }
+
+    public void setBalanceTransactionObject(BalanceTransaction c) {
+        this.balanceTransaction = new ExpandableField<BalanceTransaction>(c.getId(), c);
+    }
+
+    public Long getCreated() {
+        return created;
+    }
+
+    public void setCreated(Long created) {
+        this.created = created;
+    }
+
+    public String getCurrency() {
+        return currency;
+    }
+
+    public void setCurrency(String currency) {
+        this.currency = currency;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    public Long getExpectedAvailabilityDate() {
+        return expectedAvailabilityDate;
+    }
+
+    public void setExpectedAvailabilityDate(Long expectedAvailabilityDate) {
+        this.expectedAvailabilityDate = expectedAvailabilityDate;
+    }
+
+    public String getFailureCode() {
+        return failureCode;
+    }
+
+    public void setFailureCode(String failureCode) {
+        this.failureCode = failureCode;
+    }
+
+    public String getFailureMessage() {
+        return failureMessage;
+    }
+
+    public void setFailureMessage(String failureMessage) {
+        this.failureMessage = failureMessage;
+    }
+
+    public Boolean getLivemode() {
+        return livemode;
+    }
+
+    public void setLivemode(Boolean livemode) {
+        this.livemode = livemode;
+    }
+
+    @Override
+    public Map<String, String> getMetadata() {
+        return metadata;
+    }
+
+    public void setMetadata(Map<String, String> metadata) {
+        this.metadata = metadata;
+    }
+
+    public Source getSource() {
+        return source;
+    }
+
+    public void setSource(Source source) {
+        this.source = source;
+    }
+
+    public String getStatementDescriptor() {
+        return statementDescriptor;
+    }
+
+    public void setStatementDescriptor(String statementDescriptor) {
+        this.statementDescriptor = statementDescriptor;
+    }
+
+    public String getStatus() {
+        return status;
+    }
+
+    public void setStatus(String status) {
+        this.status = status;
+    }
+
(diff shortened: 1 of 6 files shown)


src/main/java/com/stripe/model/Topup.java
```diff
@@ -0,0 +1,199 @@
...
+    public static Topup retrieve(String id) throws AuthenticationException,
+            InvalidRequestException, APIConnectionException, CardException,
+            APIException {
+        return retrieve(id, null);
+    }
+
+    public static Topup retrieve(String id, RequestOptions options)
+            throws AuthenticationException, InvalidRequestException,
+            APIConnectionException, CardException, APIException {
+        return request(RequestMethod.GET, instanceURL(Topup.class, id), null, Topup.class, options);
```
miguel-stripe: Can this one be implemented using the below method:
<pre>
public static Topup retrieve(String id, Map<String, Object> params, RequestOptions options);
</pre>
?

src/main/java/com/stripe/model/Topup.java
```diff
@@ -0,0 +1,199 @@
...
+    String id;
+    String object;
+    Long created;
+    Boolean livemode;
+    Integer amount;
+    String currency;
+    Map<String, String> metadata;
+    Source source;
+    String status;
+    String balanceTransaction;
```
miguel-stripe: Hmm, do you know how this works if a user requests to expand the `BalanceTransaction`?

https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Transfer.java#L20?

src/test/java/com/stripe/model/TopupTest.java
```diff
@@ -0,0 +1,104 @@
...
+        APIResource.setStripeResponseGetter(new LiveStripeResponseGetter());
+    }
+
+    @Test
+    public void testDeserialize() throws IOException {
+        String json = resource("topup.json");
+        Topup topup = APIResource.GSON.fromJson(json, Topup.class);
+        assertEquals("tu_HASH", topup.getId());
+        assertEquals("topup", topup.getObject());
+        assertEquals(1000, topup.getAmount().intValue());
```
miguel-stripe: Hmm, is the `intValue` needed here?  This should automatically resolve right?

src/test/java/com/stripe/model/TopupTest.java
```diff
@@ -0,0 +1,104 @@
...
+        assertEquals("usd", topup.getCurrency());
+        assertEquals("Testing", topup.getDescription());
+        assertEquals(1518134400, topup.getExpectedAvailabilityDate().intValue());
```
miguel-stripe: Why `intValue` here if `getExpectedAvailabilityDate` is a long?
Additionally, could you do
<pre>
assertEquals(1518134400L, topup.getExpectedAvailabilityDate());
</pre>

Same for `getCreated`?

src/test/java/com/stripe/model/TopupTest.java
apakulov-stripe: Yes, we needed it here, since `Assert.assertEquals` has so many overloaded methods and there are only a few that works primitive types:
- http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(long, long)
- http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(double, double, double)
We can use http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(java.lang.Object, java.lang.Object) and do explicit boxing for expected value, e.g.:

```
assertEquals(Long.valueOf(1517619386), topup.getCreated());
```

miguel-stripe: 👍 Ah, tested this, that's sucks.

In any case, should probably use `longValue` for `getCreated` and `getExpectedAvailabilityDate` right?

src/main/java/com/stripe/model/Topup.java
apakulov-stripe: Fixed in 18f6b86

src/main/java/com/stripe/model/Topup.java
```diff
@@ -0,0 +1,199 @@
+package com.stripe.model;
+
+import com.stripe.exception.*;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Map;
+
+public class Topup extends APIResource implements MetadataStore<Topup>, HasId {
+    String id;
```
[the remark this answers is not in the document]
apakulov-stripe: Fixed in 18f6b86

src/main/java/com/stripe/model/Topup.java
apakulov-stripe: Good catch, I didn't know that `BalanceTransaction` could be expanded. Fixed in 18f6b86

src/main/java/com/stripe/model/Topup.java
```diff
@@ -0,0 +1,213 @@
...
+    String object;
+    Integer amount;
+    ExpandableField<BalanceTransaction> balanceTransaction;
+    Long created;
+    String currency;
+    String description;
+    Long expectedAvailabilityDate;
+    String failureCode;
+    String failureMessage;
+    Map metadata;
```
miguel-stripe: Nit: Noticed you dropped the type parameters here.  Intended?

apakulov-stripe: Oops, my bad. Fixed in 1b51c2c

src/test/java/com/stripe/BaseStripeFunctionalTest.java
```diff
@@ -50,6 +51,10 @@ public static String getUniqueCouponId() {
 		return String.format("MY-J-COUPON-%s", UUID.randomUUID().toString().substring(24));
 	}
 
+	public static String getUniqueTopupId() {
```
brandur-stripe: I don't think that this was used in the end?

src/test/java/com/stripe/model/TopupTest.java
```diff
@@ -0,0 +1,121 @@
...
+    public void unmockStripeResponseGetter() {
+        /* This needs to be done because tests aren't isolated in Java */
+        APIResource.setStripeResponseGetter(new LiveStripeResponseGetter());
+    }
+
+    @Test
+    public void testDeserialize() throws IOException {
+        String json = resource("topup.json");
+        Topup topup = APIResource.GSON.fromJson(json, Topup.class);
+        assertEquals("tu_HASH", topup.getId());
```
brandur-stripe: Nit, but can we use `tu_123` instead? It's a convention that's at least starting to be established.

brandur-stripe:
Left a couple minor comments, but LGTM.

src/main/java/com/stripe/model/Topup.java
```diff
@@ -0,0 +1,213 @@
...
+    Integer amount;
+    ExpandableField<BalanceTransaction> balanceTransaction;
+    Long created;
+    String currency;
+    String description;
+    Long expectedAvailabilityDate;
+    String failureCode;
+    String failureMessage;
+    Map<String, String> metadata;
+    Boolean livemode;
```
ob-stripe: `livemode` should be listed before `metadata`.

src/main/java/com/stripe/model/Topup.java
```diff
@@ -0,0 +1,213 @@
...
+        this.metadata = metadata;
+    }
+
+    public Boolean getLivemode() {
+        return livemode;
+    }
+
+    public void setLivemode(Boolean livemode) {
+        this.livemode = livemode;
+    }
```
ob-stripe: Same thing here, the accessors for `livemode` should be before those for `metadata`.

ob-stripe:
Another small nit. Could you also squash everything into a single commit after fixing? Thanks!

apakulov-stripe:
Just FYI, @ob-stripe github provides squash & merge functionality from UI.

brandur-stripe:
> Just FYI, @ob-stripe github provides squash & merge functionality from UI.

I think OB's aware of that. A nice thing about a manual squash is that you (hopefully) end up with a better/cleaner/less noisy commit message in the end. It's particularly nice when you're paging back through `git log` looking at recent changes.

brandur-stripe:
(BTW, OB released this one already as 5.32.0.)

PR #455, opened 2018-02-22
Add support for code attribute on all Stripe exceptions

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

- Move the `code` attribute directly in `StripeException` class so that it is available on all Stripe exceptions
- Bump the `serialVersionUID` value from 1 to 2 on all exception classes
- Clean up / Normalize whitespace
- Mark existing constructors as deprecated with a reminder to remove them in the next major version bump (so we can avoid a breaking change for this, in case users are instantiating Stripe exceptions themselves for some reason)
- Update all existing Stripe exception raises to use the new constructor, with the code returned by the API or `null` for client-side errors

I didn't add any new tests as stripe-java's existing testing infrastructure does not provide an easy way of testing errors, but I can look into adding that if you think it'd be useful.

### DIFF
--- a/src/main/java/com/stripe/exception/APIConnectionException.java
+++ b/src/main/java/com/stripe/exception/APIConnectionException.java
@@ -1,15 +1,13 @@
 package com.stripe.exception;
 
 public class APIConnectionException extends StripeException {
-
-	private static final long serialVersionUID = 1L;
+	private static final long serialVersionUID = 2L;
 
 	public APIConnectionException(String message) {
-		super(message, null, 0);
+		this(message, null);
 	}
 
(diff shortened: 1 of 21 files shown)


brandur-stripe:
LGTM. Great cleanup.

Reading through this I was reminded that it kind of sucks how every time you add a new core field to an object in Java, you end up inflating its constructor by one more parameter and end up with a deprecated version of it for any public API ... I don't think there's a better way to do it, but man, it's messy.

> I didn't add any new tests as stripe-java's existing testing infrastructure does not provide an easy way of testing errors, but I can look into adding that if you think it'd be useful.

Thanks for taking a look at it at least.

PR #456, opened 2018-02-23
Add README section for "per-request configuration"

brandur:
Adds a section to the README that explicitly describes how to configure
options like API key and `Stripe-Account` on a per-request basis. This
is intended to mirror what we already have in the READMEs of sibling
projects like `stripe-ruby`.

r? @ob-stripe

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,13 +61,15 @@ import com.stripe.net.RequestOptions;
 public class StripeExample {
 
     public static void main(String[] args) {
-        RequestOptions requestOptions = (new RequestOptionsBuilder()).setApiKey("YOUR-SECRET-KEY").build();
+        Stripe.apiKey = "sk_test_...";
+
         Map<String, Object> chargeMap = new HashMap<String, Object>();
         chargeMap.put("amount", 100);
         chargeMap.put("currency", "usd");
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,13 +61,12 @@ import com.stripe.net.RequestOptions;
...
 
     public static void main(String[] args) {
-        RequestOptions requestOptions = (new RequestOptionsBuilder()).setApiKey("YOUR-SECRET-KEY").build();
         Map<String, Object> chargeMap = new HashMap<String, Object>();
         chargeMap.put("amount", 100);
         chargeMap.put("currency", "usd");
         chargeMap.put("source", "tok_1234"); // obtained via Stripe.js
         try {
-            Charge charge = Charge.create(chargeMap, requestOptions);
+            Charge charge = Charge.create(chargeMap);
```
brandur-stripe: I also simplified this first example in favor of the new section.

ob-stripe: sgtm, but should we update the example to show how to set the API key globally then?

brandur-stripe: Oops, yep that would be quite helpful wouldn't it? Fixed and commit amended.

PR #457, opened 2018-02-27
Add title and version in JAR manifest

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds the `Implementation-Title`, `Implementation-Version` and `Implementation-Vendor` attributes to the JAR manifest file. This is good practice for distributed JARs and allows for programmatically checking the version.


brandur-stripe:
Slick! Nice how they can re-use the properties file too.

LGTM.

PR #458, opened 2018-02-27
Remove Transfer's sourceTransaction deprecated message

gh-stripe:
We deprecated the sourceTransaction property for a few months when we split Transfers and Payouts but forgot to remove the message when it was reinstated.

I ran the test suite and the build was successful.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released in 5.33.2.

PR #461, opened 2018-03-07
Add UsageRecord creation ability

abouzek:
While working on a project utilizing the new metered/usage based billing, I noticed the library didn't support UsageRecords yet, despite the docs having Java examples (probably due to auto-generated docs based on the template):

https://stripe.com/docs/api/java#usage_record_create

This is not really what I would consider a 'final' or mergeable copy of work, but I wanted to throw it out there for anyone utilizing the new feature.

I had to add `subscriptionItem` ID as a parameter passed to `create` because Stripe API will throw a 400 if you are using a sub-object endpoint such as `subscription_items/<subscription_item_id>/usage_records` and you pass the subscription item ID in both the URL and the body params. I think this problem needs to be solved in a more generic way, but for the intent of moving forward in my current project I am leaving as is and am open for suggestions :)

### DIFF
--- a/src/main/java/com/stripe/model/UsageRecord.java
+++ b/src/main/java/com/stripe/model/UsageRecord.java
@@ -1,0 +1,71 @@
+package com.stripe.model;
+
+import com.stripe.exception.*;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Map;
+
+
+public class UsageRecord extends APIResource implements HasId {
+
+    String id;
+    String object;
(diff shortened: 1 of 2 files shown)


alexander-stripe:
Thank you very much for the pull request! I've just opened a new PR which includes all our flexible billing primitives including the usage record endpoints: https://github.com/stripe/stripe-java/pull/462 .

abouzek:
Awesome, thanks @alexander-stripe. I'll go ahead and close this as it duplicates some of that work.

PR #462, opened 2018-03-08
Flexible billing primitives

alexander-stripe:
Adds flexible billing primitives to the SDK for beta functionality. Tests require an API key that has the functionality enabled.

### DIFF
--- a/src/main/java/com/stripe/model/Plan.java
+++ b/src/main/java/com/stripe/model/Plan.java
@@ -8,12 +8,14 @@ import com.stripe.exception.InvalidRequestException;
 import com.stripe.net.APIResource;
 import com.stripe.net.RequestOptions;
 
+import java.util.List;
 import java.util.Map;
 
 public class Plan extends APIResource implements MetadataStore<Plan>, HasId {
   String id;
   String object;
   Long amount;
+  String billingScheme;
   Long created;
   String currency;
   String interval;
@@ -22,6 +24,10 @@ public class Plan extends APIResource implements MetadataStore<Plan>, HasId {
   Map<String, String> metadata;
   String nickname;
   ExpandableField<Product> product;
+  List<PlanTier> tiers;
+  String tiersMode;
+  PlanTransformUsage transformUsage;
+  String usageType;
 
   @Deprecated
   String statementDescription;
@@ -54,6 +60,14 @@ public class Plan extends APIResource implements MetadataStore<Plan>, HasId {
 
   public void setAmount(Long amount) {
     this.amount = amount;
+  }
+
+  public String getBillingScheme() {
+    return billingScheme;
+  }
+
+  public void setBillingScheme(String billingScheme) {
+    this.billingScheme = billingScheme;
   }
 
   public Long getCreated() {
@@ -190,6 +204,38 @@ public class Plan extends APIResource implements MetadataStore<Plan>, HasId {
   @Deprecated
   public void setStatementDescription(String statementDescription) {
     this.statementDescription = statementDescription;
+  }
+
+  public List<PlanTier> getTiers() {
+    return tiers;
+  }
+
+  public void setTiers(List<PlanTier> tiers) {
+    this.tiers = tiers;
+  }
+
+  public String getTiersMode() {
+    return tiersMode;
+  }
+
+  public void setTiersMode(String tiersMode) {
+    this.tiersMode = tiersMode;
+  }
+
+  public PlanTransformUsage getTransformUsage() {
+    return transformUsage;
+  }
+
+  public void setTransformUsage(PlanTransformUsage transformUsage) {
+    this.transformUsage = transformUsage;
+  }
+
+  public String getUsageType() {
+    return usageType;
+  }
+
+  public void setUsageType(String usageType) {
+    this.usageType = usageType;
   }
 
   public static Plan create(Map<String, Object> params)
--- a/src/main/java/com/stripe/model/TransferReversalCollection.java
+++ b/src/main/java/com/stripe/model/TransferReversalCollection.java
@@ -54,9 +54,16 @@ public class TransferReversalCollection extends StripeCollection<Reversal> {
     return retrieve(id, RequestOptions.builder().setApiKey(apiKey).build());
   }
 
+  /**
+   * Load a transfer reversal.
+   * @param id Id of the reversal
+   * @param options Further request options
+   * @return The transfer reversal
+   */
   public Reversal retrieve(String id, RequestOptions options) throws AuthenticationException,
       InvalidRequestException, APIConnectionException, CardException,
       APIException {
+    // TODO replace with subresourceURL
     String url = String.format("%s%s/%s", Stripe.getApiBase(), this.getURL(), id);
     return APIResource.request(APIResource.RequestMethod.GET, url, null, Reversal.class, options);
   }
--- a/src/main/java/com/stripe/model/UsageRecord.java
+++ b/src/main/java/com/stripe/model/UsageRecord.java
@@ -1,0 +1,98 @@
+package com.stripe.model;
+
+import com.stripe.exception.APIConnectionException;
+import com.stripe.exception.APIException;
+import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class UsageRecord extends APIResource implements HasId {
+  String id;
+  String object;
+  Boolean livemode;
+  Long quantity;
+  String subscriptionItem;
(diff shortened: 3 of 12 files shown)


src/main/java/com/stripe/net/APIResource.java
```diff
@@ -132,6 +134,24 @@ protected static String instanceURL(Class<?> clazz, String id, String apiBase)
...
+			throws InvalidRequestException {
+		try {
+			return String.format("%s/%s/%ss", classURL(clazz, apiBase), urlEncode(id), className(subClazz));
+		} catch (UnsupportedEncodingException e) {
+			throw new InvalidRequestException("Unable to encode parameters to "
+					+ CHARSET
+					+ ". Please contact support@stripe.com for assistance.",
+					null, null, null, 0, e);
+		}
+	}
```
brandur-stripe: Hm, I'm not too crazy about adding a couple new global helpers like this and then using them in just the one place. It feels like we should either just re-use the template found elsewhere (see transfer reversals for example), or convert everything over to use these.

What do you think @ob-stripe?

brandur-stripe: And actually, looking at what's already in here, it seems like we just have subresources in transfer reversals, external accounts, and application fee refunds, so might be plausible to just move everything.

brandur-stripe:
@alexander-stripe Thanks for addressing that other feedback and for answering #461! Left one comment above, but otherwise LGTM. It seems there are at least a few users who want this available, so in your minds is this ready to come into mainline?

src/main/java/com/stripe/net/APIResource.java
ob-stripe: Yeah, I feel the same. I'm fine with adding a new helper method but let's try and make its use consistent.

> transfer reversals, external accounts, and application fee refunds

Also customer sources.

alexander-stripe: So TL;DR; is to move those to the `subresourceURL` helper? If yes I think it's reasonable to do this in another PR. wdyt?

alexander-stripe:
@brandur-stripe : The feature is currently in (private) beta and requires a gate, so I'm not sure if merging it to master is the right thing todo yet.

brandur-stripe:
> The feature is currently in (private) beta and requires a gate, so I'm not sure if merging it to master is the right thing to do yet.

Thanks. Sounds good.

alexander-stripe:
Okay, this PR is now ready to be merged, the API endpoints have been soft-released.

cc @brandur-stripe / @ob-stripe

alexander-stripe:
Thanks! i've addressed the comments.

r? @brandur-stripe @ob-stripe

src/main/java/com/stripe/model/UsageRecord.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * Create a new usage record associated with a subscription item.
+   * @param subscriptionItem Id of the subscription item
+   * @param params The quantity, the timestamp and the conflict behaviour (action)
+   * @param options Request options
+   * @return The created usage record
+   */
+  public static UsageRecord create(String subscriptionItem,
```
brandur-stripe: Hm, I think it's pretty conventional for every `create` to use a `params` object (see `ag 'public.* create\('`). Mind converting this over?

brandur-stripe:
Left a minor comment above, but looks good! Thanks @alexander-stripe.

src/main/java/com/stripe/model/UsageRecord.java
alexander-stripe: In this case the `subscriptionItem` is required to build the URL, so I think it's nice to have this documented in the signature? If we would pass a `params` object then we would either need to check for a `subscription_item` key at runtime and throw an exception if something goes wrong, or it will result in weird errors for the user if forgotten if we default to `""` (=> either `404` or invalid url?)

alexander-stripe:
Answered to your comment above.

r? @brandur-stripe

src/main/java/com/stripe/model/UsageRecord.java
brandur-stripe: @alexander-stripe 

Yeah, I agree it'd be nice to get this check in client-side. Given that that there isn't even a single example of a `create` that doesn't take `params` elsewhere though, I don't think we should break the pattern.

Would it make more sense if the action was on subscription item? We have methods for resource-specific actions elsewhere in the library (see `refund` on `ApplicationFee` for example).

alexander-stripe: Moving it to the subscription item is also not optimal as you'd need to load the subscription item just to report usage. 

I'll remove the `subscriptionItem` param and add a client side check.

alexander-stripe:
Comments addressed :-)

I don't know why the one java build is failing, looks like a download of some dependency is not working?

src/main/java/com/stripe/exception/MissingRequiredParameter.java
```diff
@@ -0,0 +1,16 @@
+package com.stripe.exception;
+
+public class MissingRequiredParameter extends Exception {
```
ob-stripe: I have mixed feelings about introducing a new exception class for this. The existing pattern in the library is to throw `InvalidRequestException` when there is a problem with the arguments. This isn't great because it conflates two different types of issues (client-side or server-side) in a single exception class, but it's pretty consistent right now.

I'd be very much in favor of having distinct exceptions, but only if we're consistent about it, which would require some BC incompatible changes.

@brandur-stripe @remi-stripe Any thoughts on the above?

ob-stripe:
> I don't know why the one java build is failing, looks like a download of some dependency is not working?

Yeah, we need to download an external dependency (BouncyCastle) to work around some missing crypto primitives in OpenJDK 7 that Gradle requires. Ironically, the certificate on https://downloads.bouncycastle.org is expired and causes the download to fail.

Hopefully this will be fixed soon, but in the meantime it should be safe to ignore the OpenJDK 7 build failures as long as the other builds succeed.

src/main/java/com/stripe/exception/MissingRequiredParameter.java
remi-stripe: I agree with you that we should keep the exceptions list as is for now until we make this in all client-side libraries. It feels like a normal "invalid request" to me.

brandur-stripe: +1. It's not ideal, but we should optimize for consistency, especially where exceptions are concerned in a language like Java that has checked exceptions (existing users might have to go and start decorating Stripe-related method signatures with this new exception type).

alexander-stripe: It's not great b/c `InvalidRequestException` has things like `statusCode`,  `requestId` etc. I guess it's fine to set them to `null`, but doesn't feel super clean.

alexander-stripe: I've changed it to use `InvalidRequestException`.

brandur-stripe:
Thanks @alexander-stripe!

@ob-stripe Mind taking the final review pass on this one? I feel like you're catching a lot more mistakes these days than I am.

alexander-stripe:
All checks are passing except for `OpenJDK 7` for the reasons mentioned above :)

src/main/java/com/stripe/model/Plan.java
```diff
@@ -22,6 +24,10 @@
   Map<String, String> metadata;
   String nickname;
   ExpandableField<Product> product;
+  List<PlanTier> tiers;
+  PlanTransformUsage transformUsage;
+  String tiersMode;
```
ob-stripe: `tiersMode` should be before `transformUsage`.

src/main/java/com/stripe/model/TransferReversalCollection.java
```diff
@@ -54,9 +54,16 @@ public Reversal retrieve(String id, String apiKey) throws AuthenticationExceptio
     return retrieve(id, RequestOptions.builder().setApiKey(apiKey).build());
   }
 
+  /**
+   * Load a transfer reversal.
+   * @param id Id of the reversal
+   * @param options Further request options
+   * @return The transfer reversal
+   */
```
ob-stripe: So Checkstyle has this rule that any method with more than 2 lines of code has to have Javadoc. I guess comments count towards this... Thanks for adding the Javadoc :)

src/main/java/com/stripe/model/Plan.java
alexander-stripe: Would be cool if a linter could catch this :-)

alexander-stripe:
Alright, nit fixed. Thanks 👍

ob-stripe:
@alexander-stripe One final request before merging: can you squash into a single commit with a clear message?

ob-stripe:
Thanks a lot for bearing with us @alexander-stripe :)

Released in 5.35.0.

PR #466, opened 2018-03-22
Fix serialization of null expandable attributes

### LINKED ISSUE #465: .toString returns NullPointerException when called for an Order
Hi all,

I have successfully created some test orders. However, when I try to call the method `toString` for an `Order`,  I get the following error:
```
NullPointerException   com.google.gson.JsonPrimitive.isPrimitiveOrString (JsonPrimitive.java:282)
```

Note that if I call all the other `Order` methods (such as `getId`), everything works fine. 

Thanks.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @remi-stripe @elenacanovi

Fixes the serialization of expandable attributes when both the ID and the expanded model are `null`.

Fixes #465.


brandur-stripe:
LGTM. (Thanks for adding the new test too!)

PR #472, opened 2018-03-26
Fix null-or-empty checks

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fix a couple of buggy null-or-empty string checks.


brandur-stripe:
LGTM.

PR #473, opened 2018-03-28
Fix encoding of arrays

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

While working on #452, I noticed that the library can encode `List`s, but not native arrays.

- fix encoding of arrays
- no longer URL-encode `[` and `]`. This is consistent with stripe-ruby (https://github.com/stripe/stripe-ruby/blob/529b9ec4176c1c3c1b0d65587c88cc008f8652c5/lib/stripe/util.rb#L198-L202) and makes the encoded query strings easier to read

### DIFF
--- a/src/main/java/com/stripe/net/APIResource.java
+++ b/src/main/java/com/stripe/net/APIResource.java
@@ -7,6 +7,7 @@ import java.util.Map;
 import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+
 import com.stripe.Stripe;
 import com.stripe.exception.APIConnectionException;
 import com.stripe.exception.APIException;
@@ -148,7 +149,12 @@ public abstract class APIResource extends StripeObject {
 		if (str == null) {
 			return null;
 		} else {
-			return URLEncoder.encode(str, CHARSET);
+			// Don't use strict form encoding by changing the square bracket control
+			// characters back to their literals. This is fine by the server, and
+			// makes these parameter strings easier to read.
+			return URLEncoder.encode(str, CHARSET)
+				.replaceAll("%5B", "[")
+				.replaceAll("%5D", "]");
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
```diff
@@ -139,7 +170,7 @@ public void testAppInfo() {
 
 		Stripe.setAppInfo("MyAwesomePlugin", "1.2.34", "https://myawesomeplugin.info");
 
-		Map<String, String> headers = srg.getHeaders(options);
+		Map<String, String> headers = LiveStripeResponseGetter.getHeaders(options);
```
ob-stripe: Opportunistic warning fix: `LiveStripeResponseGetter.getHeaders` is a static method and so should not be accessed from an instance.

src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
```diff
@@ -148,10 +179,12 @@ public void testAppInfo() {
 
 		Gson gson = new Gson();
 
-		Map<String, String> uaMap = gson.fromJson(headers.get("X-Stripe-Client-User-Agent"), Map.class);
+		Map<String, String> uaMap = gson.fromJson(headers.get("X-Stripe-Client-User-Agent"),
+			new TypeToken<Map<String, String>>() {}.getType());
```
ob-stripe: Another opportunistic warning fix: use `TypeToken` to work around the type erasure.

src/main/java/com/stripe/net/APIResource.java
```diff
@@ -148,7 +149,12 @@ public static String urlEncode(String str) throws UnsupportedEncodingException {
 		if (str == null) {
 			return null;
 		} else {
-			return URLEncoder.encode(str, CHARSET);
+			// Don't use strict form encoding by changing the square bracket control
+			// characters back to their literals. This is fine by the server, and
+			// makes these parameter strings easier to read.
+			return URLEncoder.encode(str, CHARSET)
+				.replaceAll("%5B", "[")
+				.replaceAll("%5D", "]");
```
brandur-stripe: Nice.

src/test/java/com/stripe/net/LiveStripeResponseGetterTest.java
brandur-stripe: Yep. Thanks for the comment.

PR #474, opened 2018-03-28
Set indentation settings in VSCode

ob-stripe:
cc @stripe/api-libraries 

Set indentation settings in VSCode:
- always render whitespace
- use tabs in `.java` files and spaces in `.json` files
- display a ruler at 100 characters

(At some point in the future, I'd like us to start using Google's Java style guide, which uses 2 spaces for indents and 100 characters per line.)


ob-stripe:
Self-approving this since it's just editor settings

PR #475, opened 2018-03-28
Use checkstyle with Google checks

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

As mentioned, I'd like to start using [Google's Java style guide](https://google.github.io/styleguide/javaguide.html) in stripe-java.

This first PR simply adds `checkstyleMain` and `checkstyleTest` tasks in Gradle to generate checkstyle reports. For now, the tasks can only be called explicitly and are not part of the build process. Once we've fixed all issues, we can enable the tasks as part of the build process to ensure the style is actually enforced going forward.

The choice of using Google's guidelines is a bit arbitrary, and I expect that we will customize some of the rules. E.g. Google says acronyms should be camel-cased (XML -> `Xml`) while we keep them in all caps (e.g. `SKU` or `RedirectURL`).

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,7 @@
 {
     "recommendations": [
         "vscjava.vscode-java-pack",
+        "shengchen.vscode-checkstyle",
         "naco-siren.gradle-language"
     ]
 }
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Wow, coding conventions for `stripe-java`? Never thought I'd see the day :)

LGTM.

PR #479, opened 2018-03-29
Limit line length to 100 characters

ob-stripe:
cc @stripe/api-libraries 

Break lines so that no line is longer than 100 characters.

### DIFF
--- a/src/main/java/com/stripe/exception/APIException.java
+++ b/src/main/java/com/stripe/exception/APIException.java
@@ -12,7 +12,8 @@ public class APIException extends StripeException {
     this(message, requestId, null, statusCode, e);
   }
 
-  public APIException(String message, String requestId, String code, Integer statusCode, Throwable e) {
+  public APIException(String message, String requestId, String code, Integer statusCode,
+      Throwable e) {
     super(message, requestId, code, statusCode, e);
   }
 }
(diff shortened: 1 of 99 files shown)


src/main/java/com/stripe/exception/APIException.java
```diff
@@ -12,7 +12,8 @@ public APIException(String message, String requestId, Integer statusCode, Throwa
     this(message, requestId, null, statusCode, e);
   }
 
-  public APIException(String message, String requestId, String code, Integer statusCode, Throwable e) {
+  public APIException(String message, String requestId, String code, Integer statusCode,
+      Throwable e) {
```
remi-stripe: are we sure this is better that one param per line? I'm not familiar with Java styling so wanted to ask

ob-stripe: The [style guide](https://google.github.io/styleguide/javaguide.html) doesn't enforce a specific style for how to line-wrap long declarations, but taking a look at Gson's source code, they use the same style of wrapping only when necessary: https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/reflect/TypeToken.java#L185-L186.

brandur-stripe:
+1000.

ob-stripe:
Self-approving since it's just a stylistic change.

(It's interesting that coveralls thinks the coverage decreased -- I didn't realize coveralls was actually counting which _lines_ get hit by tests. So for instance if we had `a ? b : c` and were only testing `a == true`, then changing the above to:
```
a
? b
: c
```
will result in a lower coverage % because now the `: c` line is counted as not covered.)

PR #480, opened 2018-03-29
Fix more linting errors

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fix more linting errors:
- break lines before symbols (e.g.
```
foo
&& bar
```
not
```
foo &&
bar
```
- `static` immediately after visibility modifier
- declare each variable separately

Asking for a r? as there's always a minor chance I broke something with these changes.


brandur-stripe:
I'm +1 on all of this.

LGTM.

PR #481, opened 2018-03-29
Rule for consecutive capital letters

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Google's style rules state names should not have more than 2 consecutive capital letters / acronyms should be camel-case like regular words: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case.

In the library, we keep acronyms in all caps. As much as I'd like to stick to the standard, I don't think we should introduce a breaking change just for the sake of style.

This PR changes the rule to allow for up to 5 consecutive capital letters (to allow the worst case of `populateMapFromJSONObject`), and fixes a couple instances of final variables that were styled as constants.

### DIFF
--- a/src/main/java/com/stripe/model/ExternalAccountTypeAdapterFactory.java
+++ b/src/main/java/com/stripe/model/ExternalAccountTypeAdapterFactory.java
@@ -17,7 +17,7 @@ public class ExternalAccountTypeAdapterFactory implements TypeAdapterFactory {
       return null; // this class only serializes 'ExternalAccount' and its subtypes
     }
 
-    final String SOURCE_OBJECT_PROP = "object";
+    final String sourceObjectProp = "object";
 
     final TypeAdapter<JsonElement> elementAdapter = gson.getAdapter(JsonElement.class);
     final TypeAdapter<ExternalAccount> externalAccountAdapter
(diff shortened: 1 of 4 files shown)


src/main/java/com/stripe/model/ExternalAccountTypeAdapterFactory.java
```diff
@@ -17,7 +17,7 @@
       return null; // this class only serializes 'ExternalAccount' and its subtypes
     }
 
-    final String SOURCE_OBJECT_PROP = "object";
+    final String sourceObjectProp = "object";
```
remi-stripe: don't we consider this a constant and then should follow the constant style of all caps? https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

brandur-stripe: Just OOC, I guess these were recased becaused they're `final` but not `static`?

brandur-stripe:
> In the library, we keep acronyms in all caps. As much as I'd like to stick to the standard, I don't think we should introduce a breaking change just for the sake of style.

Yep, sounds logical.

LGTM.

src/main/java/com/stripe/model/ExternalAccountTypeAdapterFactory.java
ob-stripe: Yes, the rules define constants as "static final fields whose contents are deeply immutable and whose methods have no detectable side effects": https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names.

cc @remi-stripe as this answer your question as well.

brandur-stripe: Makes sense. Thanks!

PR #482, opened 2018-03-29
Fix distance linting issues

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes distance linting issues. The linter doesn't like it when a variable is declared too far away from its first usage. This can be fixed in one of two ways:
- move the declaration closer to the first usage
- make the variable `final`

I also fixed a few other unrelated linting issues (mostly some variables in tests starting with a capital letter).


brandur-stripe:
Wow, this is one discerning linter :) Thanks for the detailed PR description + info.

LGTM.

PR #483, opened 2018-03-29
Group overload methods

ob-stripe:
cc @stripe/api-libraries 

Regroup overloaded methods.

This is just moving some method declarations around so should be fairly safe.


ob-stripe:
Tests pass, self-approving.

PR #485, opened 2018-03-29
Fix variable and method names

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Under the new style rules, method and variable names must start with two lowercase letters, and not include underscores.


brandur-stripe:
> Under the new style rules, method and variable names must start with two lowercase letters, and not include underscores.

I love this so much, haha — just looking at this diff, it's amazing of all the kind of junk that's worked it's way in over the years. I can't even imagine the mental gymnastics that were required to justify methods that start with underscores in a language that has visibility rules and where this is not a convention.

Just OOC, do you know if the "methods must start with two lowercase numbers" rule there to prevent people from trying to use Hungarian notation?

LGTM.

PR #486, opened 2018-03-29
Fix linting issue about having multiple top-level classes in same file

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

The rules enforce having one dedicated file per top-level class. The only offense was in `PagingIteratorTest` which defines a couple of classes for testing. I turned the classes into nested static classes to fix the issue.


brandur-stripe:
LGTM.

PR #487, opened 2018-03-30
Fix missing Javadoc linting issues

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes the missing Javadoc linting issues.

The style rules requires that all public methods with a body longer than 2 lines have a Javadoc summary.

I fixed this in a few different ways:
- reworking bodies so that they're under 2 lines :P (mostly for the expandable field getters)
- changing the method visibilities (there were a few public methods that really should have been protected or private)
- occasionally, adding Javadoc :)

This PR brings the number of Checkstyle issues to zero! So once this is merged, we can start enforcing the rules as part of the build process.

### DIFF
--- a/src/main/java/com/stripe/exception/StripeException.java
+++ b/src/main/java/com/stripe/exception/StripeException.java
@@ -8,6 +8,8 @@ public abstract class StripeException extends Exception {
   private Integer statusCode;
 
   /**
+   * Constructs a new Stripe exception with the specified details.
+   *
    * @deprecated Use new constructor with `code` argument instead.
    */
   @Deprecated
@@ -17,6 +19,8 @@ public abstract class StripeException extends Exception {
   }
 
   /**
+   * Constructs a new Stripe exception with the specified details.
+   *
    * @deprecated Use new constructor with `code` argument instead.
    */
   @Deprecated
@@ -29,6 +33,9 @@ public abstract class StripeException extends Exception {
     this(message, requestId, code, statusCode, null);
   }
 
+  /**
+   * Constructs a new Stripe exception with the specified details.
+   */
   public StripeException(String message, String requestId, String code, Integer statusCode,
       Throwable e) {
     super(message, e);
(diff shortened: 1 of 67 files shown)


src/main/java/com/stripe/exception/StripeException.java
```diff
@@ -49,6 +56,12 @@ public Integer getStatusCode() {
     return statusCode;
   }
 
+  /**
+   * Returns a description of the exception, including the HTTP status code and request ID (if
+   *     applicable).
```
brandur-stripe: Is this line supposed to be indented?

src/main/java/com/stripe/model/ChargeOutcome.java
```diff
@@ -30,26 +30,26 @@ public String getType() {
     return type;
   }
 
+  /**
+   * Returns the {@code rule} object, if expanded. If not expanded, use
+   *     {@link #getRuleId()} to get the ID.
```
brandur-stripe: And same question here.

brandur-stripe:
> The style rules requires that all public methods with a body longer than 2 lines have a Javadoc summary.

Interesting that they have a "2 line rule" — I haven't really seen that anywhere else.

Nice! This is obviously a huge patch so I'm not sure that I would've caught everything, but given that it's mostly changes to comments, I think it's pretty safe to fix anything we missed later. LGTM.

> This PR brings the number of Checkstyle issues to zero! So once this is merged, we can start enforcing the rules as part of the build process.

OMG.

src/main/java/com/stripe/exception/StripeException.java
ob-stripe: Oops! No, indentation in Javadoc is not necessary in the summary, only when using e.g. `@param` or `@return`. Fixed!

ob-stripe:
> Interesting that they have a "2 line rule" — I haven't really seen that anywhere else.

Same -- I only found out about that rule empirically.

The state of Javadoc in the library is still pretty sad, tbh, but we can tackle that later (maybe by using the OpenAPI spec to document the resource models and attributes).

PR #489, opened 2018-03-30
Enforce no checkstyle issues during build

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Enforce no checkstyle issues during builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,9 @@ notifications:
     on_success: never
 
 script:
-  - ./gradlew clean cobertura coveralls
+  - ./gradlew clean
+  - if [ "$TRAVIS_JDK_VERSION" == "oraclejdk9" ]; then ./gradlew check; fi
+  - ./gradlew cobertura coveralls
 
 before_cache:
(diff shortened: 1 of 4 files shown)


brandur-stripe:
The day of reckoning!

LGTM, but it looks like the build failed under OpenJDK 7:

```
* What went wrong:
Execution failed for task ':checkstyleMain'.
> com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask : Unsupported major.minor version 52.0
```

Maybe it's too old a version to support this task?

ptal @ob-stripe

ob-stripe:
Sigh. Of course the most recent version of Checkstyle requires Java 8. I'll look for a workaround.

ob-stripe:
Okay, it's not the most elegant solution, but I modified `.travis.yml` to only run the `check` task with Java 9.

brandur-stripe:
> Okay, it's not the most elegant solution, but I modified .travis.yml to only run the check task with Java 9.

Yeah, it's not perfect, but that seems like the right move.

PR #492, opened 2018-04-03
added apikey configuration options

japzio:


japzio:
ability to set api key on the ff locations :
* ~/stripe/credentials file
* stripe.apiKey system properties
* STRIPE_API_KEY env var

brandur:
Hi @japzio, thanks a lot for the effort here!

I just wanted to give you an early heads up that although I'm quite support of at least a few of the changes (I like the idea of getting a `STRIPE_API_KEY` from environment convention in place across all our libraries — it's already in `stripe-node`), getting something like this in might require some fairly significant changes. A few thoughts that I had reading this over:

1. Does this PR work right now? It wasn't immediately obvious to me where the new credentials providers were being read from.
2. I don't want to support a special file at `~/.stripe/credentials`. It doesn't seem to me to be very necessary or conventional, and storing secrets on the file system like that is generally not considered good practice. If users really want to do this, I think we can enable similar functionality by allowing `STRIPE_API_KEY` combined with a solution like [direnv](https://github.com/direnv/direnv).
3. I'm not sure that this should necessitate changes that go directly into users code (i.e., like writing something like `StripeCredentials credentials = new SharedCredentialsProvider().getCredentials();`. Instead we should just prefer `Stripe.apiKey` if available, then fall back to `STRIPE_API_KEY` and then system properties. This could be supported through the addition of some new utility classes internally, but I don't think that we should expose them as new APIs.

There's also a few more minor things with the code in the patch and some conventions, but we can take a look at those on a future review pass.

Anyway, if you're willing to work through some of the above, we could get an updated version of this merged.

japzio:
Yeah sure. your suggestion is much more cleaner. i'll try to work on this during my free time.

ob-stripe:
Closing this due to age, but feel free to reply if you'd like to resume work on this and I'll reopen.

PR #493, opened 2018-04-05
Update Bouncy Castle to 1.59

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Looks like the cert issues are fixed, so let's update the dependency!


brandur-stripe:
Nice!

LGTM.

PR #494, opened 2018-04-05
Implement Google's Error Prone in the build

codylerum:
This PR implements Googles Error Prone code analysis http://errorprone.info/

I’ve gone through and corrected the raised errors.

Error Prone cannot run on Java 7 and, as I am not a Gradle person, I don’t know how to conditionally run it. Thus the travis step for Java 7 is currently REMOVED. I would imagine that, like Checkstyle, you can configure this to only run for the Java 9 (or current latest) build.

http://errorprone.info/docs/installation

Per Google’s recommendation this is the Gradle plugin used:

https://github.com/tbroyer/Gradle-errorprone-plugin 

I’ve also added an Eclipse Java style file that is based off Google’s Java style and tweaked to conform to your Checkstyle. It works for everything I’ve run into, so it’s probably a good starting point.

I would also recommend configuring error prone to fail on warning so that small issues don't creep in over time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ out/*
 # compiler output
 bin/
 
+/.gradle/
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ addons:
   hostname: short-hostname
 
 jdk:
-  - openjdk7
   - openjdk8
   - oraclejdk8
   - oraclejdk9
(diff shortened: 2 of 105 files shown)


ob-stripe:
Hi @codylerum, thanks for the PR! This looks super interesting.

We won't be able to merge this PR as is because it includes many different changes, some of which I suspect may be backwards-incompatible. Also, we're in the process of rewriting the test suite (cf. #452) and I'd rather avoid introducing merging more PRs that that affect the test codebase until we're done to avoid painful rebases. Once that's done though, I'm very much in favor of using this!

I'm going to close this PR, but I'll open an issue to keep track of this.

codylerum:
No problem. Let me do a second PR that fixes just the bugs that were identified.

PR #503, opened 2018-05-07
Fix testSourceCreateRead

rattrayalex-stripe:
Bitcoin is no longer supported, so use ach_credit_transfer as the example source in this functional test.

r? @brandur-stripe


brandur-stripe:
LGTM.

Go for the merge on both these PRs and let's just check Travis to make sure `master` comes out green.

PR #509, opened 2018-05-28
Fix invoice test

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

The test suite has been broken for the past 17 days because of an upstream change. The ["list invoice lines"](https://stripe.com/docs/api#invoice_lines) endpoint (`/v1/invoices/<INVOICE_ID>/lines`) now only accepts pagination parameters and nothing else. 

For context: previously, the endpoint accepted (and ignored) more parameters due to the implementation being shared with the ["list upcoming invoice lines"](https://stripe.com/docs/api#upcoming_invoice_lines) endpoint (`/v1/invoices/upcoming/lines`).

This caused one of the tests in the suite to fail because it was sending a `subscription` parameter that is now rejected by the API. Removing the parameter fixes the test.


ob-stripe:
Actually, going to self-approve this one as it's pretty trivial.

brandur-stripe:
Nice detective work on that on OB.

For what it's worth, LGTM.

PR #511, opened 2018-05-30
Test more Java versions

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Java is moving to a faster development cycle where a new major version will be released every 6 months (and one LTS version every 3 years). Java 10 was released a few weeks back.

Travis has been fairly slow in adding support for these new versions, so someone wrote a script to install the missing JDKs. This PR uses the script in order to test the following versions that are not natively supported by Travis:
- OpenJDK 9
- Oracle JDK 10
- OpenJDK 10

(The script also supports the "early access" versions of Java 11, but then Gradle doesn't yet support Java 11 so there's not much point in adding those versions for now.)

References:
- https://adtmag.com/articles/2018/03/21/java-10.aspx
- https://github.com/travis-ci/travis-ci/issues/9368
- https://sormuras.github.io/blog/2017-12-08-install-jdk-on-travis.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,9 @@ addons:
     - short-hostname
   hostname: short-hostname
 
-jdk:
-  - openjdk7
-  - openjdk8
-  - oraclejdk8
-  - oraclejdk9
-
 before_install:
+  # Install script for missing JDKs
+  - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Neat. These shell scripts that modify the environment in arbitrary ways scare me a little bit, but this one seems to be pretty well maintained, and we can always go back.

LGTM.

PR #512, opened 2018-06-06
Fix update() methods for FeeRefund and Reversal

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Currently, trying to update a `FeeRefund` or `Reversal` instance will fail because the URL is formed with `this.fee` / `this.transfer` to find the ID of the owner. Because these fields are expandable, accessing the attributes directly results in URLs like `/v1/application_fees/com.stripe.model.ExpandableField@792bbc74/refunds/fr_123`.

The fix is simply to use the getter, which will return the proper ID of the owner resource.

No tests in this PR, but there are tests for this in #452 and I've verified that this fix makes the tests pass.


remi-stripe:
LGTM

PR #513, opened 2018-06-06
Add support for on_behalf_of attribute on charge objects

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add support for [`on_behalf_of`](https://stripe.com/docs/api#charge_object-on_behalf_of) attribute on charge objects.

### DIFF
--- a/src/main/java/com/stripe/model/Charge.java
+++ b/src/main/java/com/stripe/model/Charge.java
@@ -45,8 +45,9 @@ public class Charge extends APIResource implements MetadataStore<Charge>, HasId 
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Invoice> invoice;
   Boolean livemode;
   Map<String, String> metadata;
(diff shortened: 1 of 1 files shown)


src/main/java/com/stripe/model/Charge.java
```diff
@@ -218,6 +219,24 @@ public void setInvoiceObject(Invoice c) {
   }
   // </editor-fold>
 
+  // <editor-fold desc="onBehalfOf">
+  public String getOnBehalfOf() {
+    return (this.onBehalfOf != null) ? this.onBehalfOf.getId() : null;
+  }
+
+  public void setOnBehalfOf(String onBehalfOfID) {
+    this.onBehalfOf = APIResource.setExpandableFieldID(onBehalfOfID, this.destination);
```
remi-stripe: I don't understand why `this.destination` is used here, can you clarify?

ob-stripe: Copy/paste fail -_- Thanks!

Issue #514, opened 2018-06-06
Missing "hosted page url" in "Invoice" object.


afzalex:
According to current api version of stripe, an Invoice object has "hosted_page_url " which could be used to directly pay invoice using Stripe hosted page.

brandur-stripe:
Just to be clear, you meant `hosted_invoice_url` right?

afzalex:
yeah ,, my typo.. its `hosted_invoice_url`

remi-stripe:
those were added just 2 days ago but I can push a PR in a bit!

remi-stripe:
Pushed a PR in https://github.com/stripe/stripe-java/pull/515. Will fix the other languages too while I'm at it.

ob-stripe:
Released in 5.39.0.

PR #515, opened 2018-06-06
Add invoice's url and pdf properties

remi-stripe:
Fixes https://github.com/stripe/stripe-java/issues/514

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Invoice.java
+++ b/src/main/java/com/stripe/model/Invoice.java
@@ -1,5 +1,7 @@
 package com.stripe.model;
 
+import com.google.gson.annotations.SerializedName;
+
 import com.stripe.exception.APIConnectionException;
 import com.stripe.exception.APIException;
 import com.stripe.exception.AuthenticationException;
@@ -37,6 +39,8 @@ public class Invoice extends APIResource implements MetadataStore<Invoice>, HasI
   Long dueDate;
   Long endingBalance;
   Boolean forgiven;
+  @SerializedName("hosted_invoice_url") String hostedInvoiceURL;
+  @SerializedName("invoice_pdf") String invoicePDF;
   InvoiceLineItemCollection lines;
   Boolean livemode;
   Map<String, String> metadata;


brandur-stripe:
Haha, beat me to it Remi!

This looks good to me. @ob-stripe would you mind taking a quick look at this as well? Looking around the code, I couldn't figure out whether we're trying to move towards `hostedInvoicePDF` or `hostedInvoicePdf`. It looks to me like what Remi's put here is right (based on lots of `Id` properties elsewhere), but we've got a pretty crazy mix throughout.

ob-stripe:
In [version 2.0.0](https://github.com/stripe/stripe-java/blob/master/CHANGELOG.md#200---2016-04-08) we changed a bunch of accessors from `Url` to `URL`, so I think we should stick to all uppercase for acronyms, at least for accessors.

It looks like we're not handling this very consistently through the library though:
- either the attribute itself uses all uppercase, in which case we need a `@SerializedName` annotation, e.g.: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/ThreeDSecure.java#L33
- or the attribute only capitalizes the first letter of the acronym, in which case we need to manually create the accessors (otherwise Lombok would use the same capitalization as the attribute), e.g.: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/SourceRedirectFlow.java#L13

I have a slight preference for the first, as it's less verbose and more consistent.

remi-stripe:
yeah I looked at a random one and likely chose the wrong thing. I will fix.

remi-stripe:
fixed, PTAL

remi-stripe:
Okay so I pushed the change to keep the capitalized version. But I think that if we have to add custom logic to serialize those, we likely should default to the other approach and keep `Url` and `Pdf` instead.

ob-stripe:
> But I think that if we have to add custom logic to serialize those, we likely should default to the other approach and keep `Url` and `Pdf` instead.

Switching to capitalizing only the first letter would also make us 100% compliant with Google's Java style rules[0], so I'd be in favor of that in the next major release.

For now, pulling this in.

[0] https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

PR #519, opened 2018-06-08
Fix all warnings

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This PR fixes all warnings, and enables the `-Werror` compiler flag to treat warnings as errors to ensure we don't introduce new warnings going forward.

Some of the changes are a bit more sensitive than others, so I split the PR in smaller commits to make it easier to review.

### DIFF
--- a/src/main/java/com/stripe/net/APIResource.java
+++ b/src/main/java/com/stripe/net/APIResource.java
@@ -195,7 +195,7 @@ public abstract class APIResource extends StripeObject {
    * options and params through so that we can iterate to the next page if
    * necessary.
    */
-  public static <T extends StripeCollectionInterface> T requestCollection(
+  public static <T extends StripeCollectionInterface<?>> T requestCollection(
       String url, Map<String, Object> params, Class<T> clazz,
       RequestOptions options)
       throws AuthenticationException, InvalidRequestException,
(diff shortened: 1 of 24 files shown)


src/main/java/com/stripe/net/APIResource.java
```diff
@@ -72,38 +72,20 @@ public static void setStripeResponseGetter(StripeResponseGetter srg) {
...
-      return "three_d_secure";
-    } else if (className.equals("usagerecord")) {
-      return "usage_record";
-    } else {
-      return className;
+    // Convert CamelCase to snake_case
+    final String className = clazz.getSimpleName()
+        .replaceAll("(.)([A-Z][a-z]+)", "$1_$2")
+        .replaceAll("([a-z0-9])([A-Z])", "$1_$2")
+        .toLowerCase();
```
brandur-stripe: Just to verify: we're pretty confident the test suite would catch a problem here (if one happened to be introduced)?

ob-stripe: Yes, as long as all resources are tested and the asserted URLs are correct (which should be the case, save for some deprecated resources e.g. bitcoin receivers), then any issue in this code would be detected as it is used to format endpoint URLs. I'm relatively confident that this is the case -- the first version of my code had issues with consecutive capital letters and transformed `SKU` into `s_ku`, which caused the SKU functional tests to fail.

PR #523, opened 2018-06-14
Better timeout tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Instead of trying to connect to `api.stripe.com`, the timeout tests now create and use their own server socket. This should hopefully make them more reliable.

### DIFF
--- a/src/test/java/com/stripe/functional/TimeoutTest.java
+++ b/src/test/java/com/stripe/functional/TimeoutTest.java
@@ -1,0 +1,49 @@
+package com.stripe.functional;
+
+import com.stripe.BaseStripeTest;
+import com.stripe.Stripe;
+import com.stripe.exception.APIConnectionException;
+import com.stripe.exception.StripeException;
+import com.stripe.model.Balance;
+import com.stripe.net.RequestOptions;
+
+import java.io.IOException;
+import java.net.Inet4Address;
+import java.net.ServerSocket;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class TimeoutTest extends BaseStripeTest {
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void testConnectTimeout() throws IOException, StripeException {
+    // Kind of a hack, but we use the non-routable address 10.255.255.0 to trigger a connection
+    // timeout
+    Stripe.overrideApiBase(String.format("http://10.255.255.0"));
+
+    thrown.expect(APIConnectionException.class);
+    thrown.expectMessage("connect timed out");
+
+    final RequestOptions options = RequestOptions.builder().setConnectTimeout(1).build();
+    Balance.retrieve(options);
+  }
+
+  @Test
+  public void testReadTimeout() throws IOException, StripeException {
+    // Create a local server that does nothing to trigger a read timeout
+    try (final ServerSocket serverSocket =
+        new ServerSocket(0, 1, Inet4Address.getByName("localhost"))) {
+      Stripe.overrideApiBase(String.format("http://localhost:%d", serverSocket.getLocalPort()));
+
+      thrown.expect(APIConnectionException.class);
+      thrown.expectMessage("Read timed out");
+
+      final RequestOptions options = RequestOptions.builder().setReadTimeout(1).build();
+      Balance.retrieve(options);
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


src/test/java/com/stripe/functional/TimeoutTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Test
+  public void testConnectTimeout() throws IOException, StripeException {
+    Stripe.overrideApiBase(String.format("http://localhost:%d", port));
+
+    thrown.expect(APIConnectionException.class);
+    thrown.expectMessage("connect timed out");
+
+    try (final Socket socket = new Socket()) {
+      // Fill backlog queue with this request so consequent requests will be blocked
```
brandur-stripe: *consecutive

brandur-stripe:
Excellent call rewriting these! Calling the previous implementation just "a hack" is far too generous haha.

Okay, two possible question/ideas that may or may help in understanding the failure here:

1. First off, I'm a little surprised that this approach works. When talking about backlog size, [the documentation seems to state quite explicitly](https://docs.oracle.com/javase/7/docs/api/java/net/ServerSocket.html#ServerSocket(int,%20int)) that connections are refused when the backlog becomes to deep (i.e., they shouldn't time out):

    > The maximum queue length for incoming connection indications (a request to connect) is set to the backlog parameter. If a connection indication arrives when the queue is full, the connection is refused.

2. Secondly, I'm not sure about this, but I suspect that maybe this could be blocking because you're never accepting the initial connection on the server side. I think the way the class is supposed to work is a little like this:

    ``` java
    // Initialize the server socket. Non-blocking.
    ServerSocket serverSocket = new ServerSocket(portNumber);

    // Block until we get a client connection
    Socket clientSocket = serverSocket.accept();

    // Do work
    ```

    So it's _possible_ that you're when you initialize an initial connection with ` socket.connect(serverSocket.getLocalSocketAddress())`, `ServerSocket` is blocking expecting something to call `accept` and take that connection of its hands. Only I'm not sure why this would work for you locally — it is possible though that there are some subtle platform-specific quirks at work here.

Anyway, if none of the above is ringing true ^^, I'd probably suggest just putting some print statements into the test case's body so that you can at least see after which line in hangs.

ptal @ob-stripe

ob-stripe:
So I think this comes down to how TCP backlog works on Linux and on BSD systems (such as macOS).

On macOS, the backlogged connection is in the `SYN_SENT` state, which does trigger the connection timeout. On Linux, the backlogged connection is in the `ESTABLISHED` state, which does not trigger the connection timeout (but does trigger the read timeout).

Long story short, this trick is not portable :(

Some googling reveals that a popular way to test connection timeouts with Java is to try and reach a non-routable address, e.g. 10.255.255.0. I gave it a try and it does seem to work, but it feels like another hack... wdyt @brandur-stripe ?

brandur-stripe:
> On macOS, the backlogged connection is in the SYN_SENT state, which does trigger the connection timeout. On Linux, the backlogged connection is in the ESTABLISHED state, which does not trigger the connection timeout (but does trigger the read timeout).

Ah, interesting.

> Some googling reveals that a popular way to test connection timeouts with Java is to try and reach a non-routable address, e.g. 10.255.255.0. I gave it a try and it does seem to work, but it feels like another hack... wdyt @brandur-stripe ?

Still seems better than what we have, so I'd be okay with it. I think I found the same Stack Overflow answer as you, and this is way better than using `www.google.com:81`.

Just OOC though, do you know if there is a special significance to `10.255.255.0` particularly? It's still theoretically routable if a machine lives there right? I guess the idea is just that it's a high enough number that it's unlikely to have been assigned on a local intranet.

remi-stripe:
Just restating that we should consider adding this feature to stripe-mock no? That seems useful across all of our libraries.

ob-stripe:
> Just restating that we should consider adding this feature to stripe-mock no? That seems useful across all of our libraries.

stripe-mock would have the same issue. Simulating a read timeout is fairly easy (just accept the connection then do nothing), but simulating a connection timeout is more difficult as the above shows.

> Just OOC though, do you know if there is a special significance to 10.255.255.0 particularly? It's still theoretically routable if a machine lives there right?

I think you're not supposed to have addresses ending with zeroes, so it's very unlikely that the address would be assigned.

ob-stripe:
Alright, I updated the PR. ptal @brandur-stripe

brandur-stripe:
LGTM.

PR #524, opened 2018-06-14
Fix issues detected by ErrorProne

ob-stripe:
r? @brandur-stripe 

A few weeks back, @codylerum opened a PR to enable the use of the [ErrorProne](http://errorprone.info/) compiler and fix a number of issues it detected.

Unfortunately, I don't think we can enable ErrorProne by default because it only works with Java 8 and 9, and we still need to support 7 for the foreseeable future. (There's probably a way to only enable it for the supported versions but it's beyond my meagre Gradle skills.)

I did run the compiler locally though, and fixed a number of issues it detected.

For the [TypeParameterUnusedInFormals](http://errorprone.info/bugpattern/TypeParameterUnusedInFormals) (see last commit), I wasn't able to simply fix the issue because it would be a breaking change, but I added new signatures that don't have the issue and flagged the existing signatures as deprecated so we can remove those in the next major version.

I've also ignored the [MissingOverride](http://errorprone.info/bugpattern/MissingOverride) warnings for now -- I'll tackle that one in a separate PR since it's going to be pretty chunderous :/

### DIFF
--- a/src/test/java/com/stripe/BaseStripeTest.java
+++ b/src/test/java/com/stripe/BaseStripeTest.java
@@ -71,7 +71,7 @@ public class BaseStripeTest {
     }
 
     String version = conn.getHeaderField("Stripe-Mock-Version");
-    if ((version != "master") && (compareVersions(version, MOCK_MINIMUM_VERSION) > 0)) {
+    if (!(version.equals("master")) && (compareVersions(version, MOCK_MINIMUM_VERSION) > 0)) {
       throw new RuntimeException(String.format(
         "Your version of stripe-mock (%s) is too old. The minimum "
         + "version to run this test suite is %s. Please see its "
--- a/src/test/java/com/stripe/BaseStripeTest.java
+++ b/src/test/java/com/stripe/BaseStripeTest.java
@@ -362,18 +362,18 @@ public class BaseStripeTest {
       IOException, MalformedURLException, ProtocolException {
     int status;
 
-    StringBuffer urlStringBuffer = new StringBuffer();
-    urlStringBuffer.append("http://localhost:" + port + path);
+    StringBuilder urlStringBuilder = new StringBuilder();
+    urlStringBuilder.append("http://localhost:" + port + path);
 
     if (expansions != null) {
-      urlStringBuffer.append("?");
+      urlStringBuilder.append("?");
       for (String expansion : expansions) {
-        urlStringBuffer.append("expand[]=");
-        urlStringBuffer.append(expansion);
-        urlStringBuffer.append("&");
+        urlStringBuilder.append("expand[]=");
+        urlStringBuilder.append(expansion);
+        urlStringBuilder.append("&");
       }
     }
-    URL url = new URL(urlStringBuffer.toString());
+    URL url = new URL(urlStringBuilder.toString());
 
     HttpURLConnection conn = (HttpURLConnection)url.openConnection();
     conn.setRequestMethod("GET");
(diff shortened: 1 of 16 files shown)


src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -71,7 +71,7 @@ public static void checkStripeMock() throws Exception {
     }
 
     String version = conn.getHeaderField("Stripe-Mock-Version");
-    if ((version != "master") && (compareVersions(version, MOCK_MINIMUM_VERSION) > 0)) {
+    if (!(version.equals("master")) && (compareVersions(version, MOCK_MINIMUM_VERSION) > 0)) {
```
brandur-stripe: Nice that it caught this! I'd feel a lot safer with this on because this is the type of mistake I make in Java all the time as an infrequent user.

src/test/java/com/stripe/net/StripeResponseTest.java
```diff
@@ -74,6 +74,6 @@ public void testGetIdempotencyKey() {
   public void testRequestId() {
     final Map<String, List<String>> headerMap = generateHeaderMap();
     final StripeResponse stripeResponse = new StripeResponse(200, chargeBody, headerMap);
-    assertEquals(stripeResponse.requestId(), "req_12345");
+    assertEquals("req_12345", stripeResponse.requestId());
```
brandur-stripe: Wow, I love that there's a check rule for this given how common of a mistake this is.

src/main/java/com/stripe/model/BalanceTransaction.java
```diff
@@ -62,10 +62,20 @@ public void setSourceObject(HasId o) {
...
+   *
+   * @deprecated Use {@link #getSourceObjectAs(Class)} instead.
+   */
+  @Deprecated
   @SuppressWarnings("unchecked")
   public <O extends HasId> O getSourceObjectAs() {
     return (this.source != null) ? (O) this.source.getExpanded() : null;
   }
+
+  public <O extends HasId> O getSourceObjectAs(Class<O> clazz) {
```
brandur-stripe: Doesn't this reduce the usefulness of this method? As in you used to be able to invoke it with just a type parameter, but now you need another parameter as well?

ob-stripe: I guess -- tbh I have a hard time wrapping my head around Java's generics so I'm not sure if this new method actually offers additional safety, or it just pleases ErrorProne for no benefit...

I'll remove this commit from the PR so that we can merge the rest of the fixes, we can always revisit this particular issue later.

codylerum:
@ob-stripe & @brandur-stripe Glad you were able to at least loop through the current warnings.

src/main/java/com/stripe/model/BalanceTransaction.java
brandur-stripe: Sounds good. I was just [reading through the documentation on this lint](https://github.com/google/error-prone/blob/master/docs/bugpattern/TypeParameterUnusedInFormals.md), but it doesn't seem to quite adequately explain all ramifications — they talk about why it's not type safe, which is useful, but they don't explain the obvious hit on usability. Looking at a method like this one, I would always have assumed that there is some runtime risk of the cast failing.

brandur-stripe:
LGTM.

> @ob-stripe & @brandur-stripe Glad you were able to at least loop through the current warnings.

@codylerum Yes! Seems like a pretty definitive win for sure.

PR #525, opened 2018-06-15
Add missing `@Override` annotations

ob-stripe:
r? @brandur-stripe 

Adds missing `@Override` annotations detected by Error Prone.

There are still a few missing annotations: `@EqualsAndHashCode` automatically annotates the `equals` and `hashCode` methods it generates with `@Override`, but it also generates a `canEqual` method that is not annotated (cf. https://github.com/rzwitserloot/lombok/issues/1343#issuecomment-374416985). So all `ExternalAccount` subclasses are missing the `@Override` annotation on their `canEqual` method that overrides `ExternalAccount`'s own method.

There doesn't seem to be a good solution for this issue, but it's not that big a deal so ignoring for now.

I also fixed `deploy.gradle` -- our custom `makeJavadocs` task did not use the delombok'd sources.

### DIFF
--- a/deploy.gradle
+++ b/deploy.gradle
@@ -97,8 +97,8 @@ afterEvaluate { project ->
         sign configurations.archives
     }
 
-    task makeJavadocs(type: Javadoc) {
-        source = sourceSets.main.allJava
+    task makeJavadocs(type: Javadoc, dependsOn: delombok) {
+        source = delombok.outputDir
         classpath = configurations.compile
         failOnError = true
     }
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -21,7 +21,7 @@ import lombok.Setter;
 @Setter
 @EqualsAndHashCode(callSuper = false)
 public class Account extends APIResource implements HasId, MetadataStore<Account> {
-  String id;
+  @Getter(onMethod = @__({@Override})) String id;
   String object;
   String businessLogo;
   String businessName;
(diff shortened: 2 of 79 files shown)


src/main/java/com/stripe/model/Account.java
```diff
@@ -21,7 +21,7 @@
 @Setter
 @EqualsAndHashCode(callSuper = false)
 public class Account extends APIResource implements HasId, MetadataStore<Account> {
-  String id;
+  @Getter(onMethod = @__({@Override})) String id;
```
ob-stripe: There is a slightly better syntax for this: `@Getter(onMethod_ = {@Override})`, but it requires Java 8+ :/

src/main/java/com/stripe/model/Order.java
```diff
@@ -110,13 +110,14 @@ public static Order retrieve(String id, Map<String, Object> params, RequestOptio
...
 
+  @Override
   public Order update(Map<String, Object> params)
       throws AuthenticationException, InvalidRequestException,
       APIConnectionException, CardException, APIException {
     return update(params, null);
   }
 
-  @Deprecated
+  @Override
```
ob-stripe: This method was flagged as `@Deprecated`, but it shouldn't have been.

src/main/java/com/stripe/model/Account.java
brandur-stripe: Ah, too bad :/

deploy.gradle
```diff
@@ -97,8 +97,8 @@ afterEvaluate { project ->
         sign configurations.archives
     }
 
-    task makeJavadocs(type: Javadoc) {
-        source = sourceSets.main.allJava
+    task makeJavadocs(type: Javadoc, dependsOn: delombok) {
+        source = delombok.outputDir
```
brandur-stripe: Nice.

brandur-stripe:
LGTM.

PR #526, opened 2018-06-15
Alphabetize methods

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Alphabetize methods for API requests in resource classes.

### DIFF
--- a/src/main/java/com/stripe/model/Plan.java
+++ b/src/main/java/com/stripe/model/Plan.java
@@ -98,42 +98,37 @@ public class Plan extends APIResource implements MetadataStore<Plan>, HasId {
   }
   // </editor-fold>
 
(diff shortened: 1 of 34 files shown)


src/main/java/com/stripe/model/Plan.java
```diff
@@ -98,96 +98,44 @@ public void setProductObject(Product product) {
...
-    this.transformUsage = transformUsage;
-  }
-
-  public String getUsageType() {
-    return usageType;
-  }
-
-  public void setUsageType(String usageType) {
-    this.usageType = usageType;
-  }
```
ob-stripe: All these accessors were unnecessary as they can be generated by Lombok. This is why the PR has a negative net line count.

PR #530, opened 2018-06-18
Added steps in readme for setting up IDEA for development.

drake-stripe:
Add steps in readme for setting up IDEA for development.

 ### Reviewers
r? @tmaxwell-stripe 
cc @stripe-internal/

 ### Summary

Adds in the couple of steps required to setup a fresh IDEA installation for working with and indexing code for `stripe-java`

 ### Motivation

I found this to be an underdocumented process that I had to go through.

 ### Safe to revert?
Yes. Please cc me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,21 @@ You can run particular tests by passing `--tests Class#method`. Make sure you us
 
 The library uses [Project Lombok][lombok]. While it is not a requirement, you might want to install a [plugin][lombok-plugins] for your favorite IDE to facilitate development.
 
+### Instructions to Install Dependencies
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Ah, thanks for the effort here Drake!

+1 in probably leaving this out for now — it's a little more maintenance-friendly to let our tooling install these dependencies wherever possible. Also, tying instructions to specific versions and menu items in IDEA will tend to increase the probability that when upgrades are made later, we'll have to remember to change the README as well. We should probably be aiming to stay useful, but not _too_ specific.

PR #531, opened 2018-06-19
Drop connection timeout test

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

We're _still_ seeing transient failures with this test (e.g. https://travis-ci.org/stripe/stripe-java/jobs/393784427#L866-L869). Let's just drop it :(


brandur-stripe:
Haha, that's terrible. Yeah, this is way more trouble than it's worth.

LGTM.

PR #533, opened 2018-06-20
stripe-java 6.0.0

### LINKED ISSUE #409: Deserializing resources should work even if the resource is deleted

ob-stripe:
- [x] Remove deprecated methods flagged for removal (#532)
- [x] Only capitalize the first letter of acronyms per Google's style rules (#534)
- [x] Change signatures to throw `StripeException` (#535)
- [x] Use `isX()` instead of `getX()` for Boolean accessors (#536)
- [x] Change all integer types to `Long` and all floating point types to `Double` (#538)
- [x] Add support for idempotency errors (#540)
- [x] Minor fixes (#541)
- [x] Move nested objects into inner static classes (#547)
- [x] Remove `Deleted` models and add `deleted` attribute instead (#553)
- [x] Use `BigDecimal` for floating point values (#558)

Fixes #409 (via #553).


codylerum:
@ob-stripe RE: https://github.com/stripe/stripe-java/pull/536#issuecomment-399557031 

If that effects the `Boolean` object as well you may want to change those to `boolean` to conform with the Java Beans spec 8.3.2 (http://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/) 

`is` is allowed for `boolean` but not `Boolean`

codylerum:
Any chance of using BigDecimal instead of doubles?

ob-stripe:
> Any chance of using BigDecimal instead of doubles?

Hey @codylerum, sorry, I'm only seeing your comment now! Using `BigDecimal` is less convenient than `Double` due to arithmetic operators not being overloaded. That said, when dealing with money, it's definitely preferable to use `BigDecimal` to avoid rounding errors.

@brandur-stripe @remi-stripe wdyt? I'm personally in favor of it. It will make the library less convenient to use where floating-point values are concerned (e.g. exchange rates), but the added precision is worth it IMO.

remi-stripe:
@ob-stripe Would you be able to give some examples of code that would be different if we moved to it? Like what would be less convenient and what can't you do as easily?

ob-stripe:
Here's an example:

```java
Long amountInEur = 499; // 4.99 EUR
Double exchangeRate = ExchangeRate.retrieve("eur").getRates().get("usd");
Long amountInUsd = Long.valueOf(Math.round(amountInEur * exchangeRate));
```

vs.

```java
Long amountInEur = 499; // 4.99 EUR
BigDecimal exchangeRate = ExchangeRate.retrieve("eur").getRates().get("usd");
Long amountInUsd = Long.valueOf(exchangeRate.multiply(new BigDecimal(amountInEur)).setScale(0, BigDecimal.ROUND_HALF_UP).intValue());
```

Basically, you have to use `BigDecimal`'s methods like `multiply()` to perform arithmetics, which means that all other factors involved must also be converted to `BigDecimal`s.

remi-stripe:
Oh wow, that is quite wordy and painful. I would hate to have to do that if I was using the library though I am not a Java developer so I don't know how common this library is in the wild. Do you know of other SDKs like ours that would be using BigDecimal?
Also is there a way to improve this by overloading arithmetic operators ourselves?

ob-stripe:
> Also is there a way to improve this by overloading arithmetic operators ourselves?

No, Java doesn't support operator overloading at all.

One thing I didn't mention is that for users who don't really care for the added precision, they can simply convert the value to a regular floating-point by calling `.doubleValue()` on the `BigDecimal` instance provided by the library.

codylerum:
@remi-stripe @ob-stripe Yes it is wordy (slightly simpler than @ob-stripe example) as BigDecimal can go direct to long.

`Long amountInUsd = exchangeRate.multiply(new BigDecimal(amountInEur)).setScale(0, BigDecimal.ROUND_HALF_UP).longValue();`

But anyone working with floating point in Java with money is/should be using BigDecimal

https://blogs.oracle.com/corejavatechtips/the-need-for-bigdecimal

codylerum:
There is definitely an argument for leaving it as the primitive Double and letting the user convert to BigDecimal on their own.

Any Java developer should definitely be converting it to a BigDecimal doing their math with BigDecimal though if it involves math involves money.

ob-stripe:
cc @stripe/api-libraries 

I think this is ready for release. Each of the individual PRs has already been reviewed, so don't feel like you have to review each of the 253 modified files :)

PR #534, opened 2018-06-21
Only capitalize the first letter of acronym per Google's style rules

ob-stripe:
cc @stripe/api-libraries 

- Updates the `checkstyle.xml` file to be an exact copy of https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml (with the exception that the severity is raised from `WARNING` to `ERROR`)
- Rename a ton of variables, methods and classes to be compliant with Google's style rules, i.e. only the first letter must be capitalized in acronyms


src/main/java/com/stripe/model/Invoice.java
```diff
@@ -42,8 +42,8 @@
   Long dueDate;
   Long endingBalance;
   Boolean forgiven;
-  @SerializedName("hosted_invoice_url") String hostedInvoiceURL;
-  @SerializedName("invoice_pdf") String invoicePDF;
+  String hostedInvoiceUrl;
+  String invoicePdf;
```
brandur-stripe: Nice to see the `SerializedName`s going away.

brandur-stripe:
LGTM. I made it through the diff until roughly halfway through the tests, and no problems found. I also looked for acronyms that you might have missed, but didn't find any.

src/main/java/com/stripe/model/Invoice.java
ob-stripe: Yes! I was also able to remove a few manually defined accessors :)

ob-stripe:
> I also looked for acronyms that you might have missed, but didn't find any.

I'm pretty confident that the checker would have failed the build if there were any left.

brandur-stripe:
> I'm pretty confident that the checker would have failed the build if there were any left.

Oh, true! Haha. This is why it's good to have tooling.

PR #535, opened 2018-06-22
Change signatures to throw StripeException

ob-stripe:
cc @stripe/api-libraries 

Change method signatures to throw `StripeException` instead of listing every possible exception class.

This makes it way easier to introduce new exceptions (e.g. `IdempotentException`).

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -1,10 +1,6 @@
 package com.stripe.model;
 
-import com.stripe.exception.ApiConnectionException;
-import com.stripe.exception.ApiException;
-import com.stripe.exception.AuthenticationException;
-import com.stripe.exception.CardException;
-import com.stripe.exception.InvalidRequestException;
+import com.stripe.exception.StripeException;
 import com.stripe.net.ApiResource;
 import com.stripe.net.RequestOptions;
 
@@ -75,9 +71,7 @@ public class Account extends ApiResource implements HasId, MetadataStore<Account
   /**
    * Create an account.
    */
-  public static Account create(Map<String, Object> params)
-      throws AuthenticationException, InvalidRequestException,
-      ApiConnectionException, CardException, ApiException {
+  public static Account create(Map<String, Object> params) throws StripeException {
     return create(params, null);
   }
 
@@ -85,8 +79,7 @@ public class Account extends ApiResource implements HasId, MetadataStore<Account
    * Create an account.
    */
   public static Account create(Map<String, Object> params, RequestOptions options)
-      throws AuthenticationException, InvalidRequestException,
-      ApiConnectionException, CardException, ApiException {
+      throws StripeException {
     return request(RequestMethod.POST, classUrl(Account.class), params, Account.class, options);
   }
   // </editor-fold>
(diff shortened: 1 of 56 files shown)


remi-stripe:
I'm excited by this change as a _maintainer_ but I wonder if we don't make the experience worse for the developer(s) integrating as they now have to dig into the code to know which ones are really being raised.

LGTM though and you can self-approve for this one I thin.

brandur-stripe:
Yeah, I'd echo Remi's comment in that it's not a clear user win, although some of those exception classes (e.g. `ApiException`) were fairly broad already, so it might not be _that_ much bigger of a problem. Maintenance and cleanliness-wise though, this seems like a good idea.

LGTM as well.

ob-stripe:
I agree with the sentiment but I think that if it turns out to be an issue, it's something that can be solved with better documentation (both in Javadoc and on stripe.com).

Pulling this in!

Flo354:
**Please, revert the change!**

Seriously guys, the previous method of throwing every exceptions even when that was not necessary was very uggly, but this one is worse!

I don't know which exceptions may be thrown without digging really deep in your code (that is not as simple as that to understand), like said @remi-stripe. It's clearly horrible, I don't want to have a generic exception since I can not customise the error message.

@ob-stripe you said that it's `something that can be solved with better documentation (both in Javadoc and on stripe.com)`, but it's completely wrong. Even if I loose hours of my life reading all the documentation and by miracle develop the ability to remember every exceptions for every possible method, how can I deal with changes over the time? At each release, should I inspect deeply the code to check if no other exceptions are thrown?

This job is the job of the IDE. So please, consider another solution! Like "throws ExceptionA ExceptionC" when and not ExceptionB when not needed.

It's really painful.

ob-stripe:
Hi @Flo354, thanks for the feedback.

I think your request is based on an incorrect assumption:
> the previous method of throwing every exceptions even when that was not necessary was very uggly

The previous signatures did _not_ include any unnecessary exceptions. All methods in the library's public API behave in the same way: they encode the request's parameters, send the request to Stripe's HTTP API, and either decode the JSON response into a model class or raise an exception if an error happened. Thus, all methods can raise the _same_ set of exceptions, and this set is exactly "all exceptions classes that derive from `StripeException`". I wouldn't have made this change otherwise.

(In practice, methods that don't trigger a payment action will never raise a `CardError`, but that's only because Stripe's HTTP API will never return an error with `type: "card_error"` for these requests. The Java compiler has no way of knowing this, so we would still have to add `CardError` to the list of possible exceptions for every method.)

I still think this is something that can be solved with better documentation. The fact that you had to dig into the library's own source code to understand which exceptions may be thrown is clearly a failure on our part to point you to the correct reference (https://stripe.com/docs/api/java#errors + https://stripe.com/docs/api/java#error_handling).

> At each release, should I inspect deeply the code to check if no other exceptions are thrown?

We use [semantic versioning](https://semver.org/) for all our libraries. That means you should always be able to safely upgrade to a new minor or patch version. Breaking changes are always released as new major versions (and throwing a new exception type would of course be considered a breaking change).

When upgrading to a new major version, you should check the [CHANGELOG](https://github.com/stripe/stripe-java/blob/master/CHANGELOG.md) to see which changes were introduced and what might need updating in your own code.

Flo354:
Ohhhh ok, I understand much better now!
Thanks for the time you took to explain this. It makes sense now.

PR #536, opened 2018-06-22
Use `isX()` instead of `getX()` for Boolean accessors

ob-stripe:
cc @stripe/api-libraries 

Change Lombok's configuration to generate `isX()` instead of `getX()` methods for `Boolean` getters. This is (I think) more idiomatic Java.

### DIFF
--- a/lombok.config
+++ b/lombok.config
@@ -1,2 +1,1 @@
-lombok.getter.noIsPrefix = true
 lombok.addLombokGeneratedAnnotation = true


remi-stripe:
I'm fine with the change if it's more idiomatic

brandur-stripe:
LGTM as well.

I guess this just flipped a lot of generated accessors though — we'll definitely want to write some helper guides for people to get upgraded to V6.

ob-stripe:
> I guess this just flipped a lot of generated accessors though — we'll definitely want to write some helper guides for people to get upgraded to V6.

Yep, absolutely 👍

codylerum:
I believe the traditional java bean way is that a that for a field such as refund

`Boolean getRefund()`

while

`boolean isRefund()`

Trying to track down the exact language on that.

ob-stripe:
@codylerum Oops, you're absolutely right! This setting only applies to `boolean` and not `Boolean` (https://projectlombok.org/features/GetterSetter). I'm not sure why I turned this on in the first place then 🤦‍♂️

codylerum:
@ob-stripe Glad to help. Keep up the good work all.

PR #540, opened 2018-06-27
Add support for idempotency errors

ob-stripe:
cc @stripe/api-libraries 

Add a new `IdempotencyException` exception class that is raised when the API returns an `idempotency_error`.

No tests because stripe-mock doesn't support this scenario and it's ~impossible to stub this right now, but the logic is fairly trivial and is exactly the same as https://github.com/stripe/stripe-ruby/blob/ec66c3f0f44274f885de8d13de5dce2657932121/lib/stripe/stripe_client.rb#L320-L329

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -5,6 +5,7 @@ import com.stripe.exception.ApiConnectionException;
 import com.stripe.exception.ApiException;
 import com.stripe.exception.AuthenticationException;
 import com.stripe.exception.CardException;
+import com.stripe.exception.IdempotencyException;
 import com.stripe.exception.InvalidRequestException;
 import com.stripe.exception.PermissionException;
 import com.stripe.exception.RateLimitException;
@@ -372,7 +373,6 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
   }
 
   private static class StripeError {
-    @SuppressWarnings("unused")
     String type;
 
     String message;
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -684,17 +684,19 @@ private static StripeResponse getMultipartStripeResponse(
...
     LiveStripeResponseGetter.StripeError error = ApiResource.GSON.fromJson(responseBody,
         LiveStripeResponseGetter.StripeErrorContainer.class).error;
     switch (responseCode) {
       case 400:
-        throw new InvalidRequestException(error.message, error.param, requestId, error.code,
-            responseCode, null);
       case 404:
-        throw new InvalidRequestException(error.message, error.param, requestId, error.code,
-            responseCode, null);
+        if (error.type.equals("idempotency_error")) {
```
remi-stripe: Why is this in `case 404`?

ob-stripe: `switch` in Java is *fall through*, i.e. all `case` blocks are executed until a `break` statement appears. So this code applies to both `case 400:` and `case 404:`, i.e. exactly what stripe-ruby does with `when 400, 404`.

Server-side, I think that `idempotency_error`s will always have a 400 status code. We should revamp the error handling logic in all libraries to prefer using the error's `type` over the status code IMHO, but that's out of scope here.

remi-stripe: I've never been a fan for using this case hack tbh, especially for something as key as this. It's really easy to miss when changing in the future. But up to you in that case, feel free to self-approve/merge

ob-stripe: I wouldn't call this a hack, it's a well documented behavior (https://docs.oracle.com/javase/tutorial/java/nutsandbolts/switch.html), though I agree it's not the most readable syntax ever. In this case it lets up slightly simplify the logic (since `invalid_request_error` can be either a 400 or a 404).

Going to merge this one, and will try to revamp the error handling logic later.

PR #542, opened 2018-06-28
Fix account.application.* events data deserialization issue 

spastorelli-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

When trying to get the data of an `account.application.*` event like so:
```java
Event event = APIResource.GSON.fromJson(data, Event.class);
Application application = (Application) event.getData().getObject();
```

casting the result of `getObject()` to `Application` throws a `java.lang.ClassCastException`:
```
java.lang.ClassCastException: com.stripe.model.StripeRawJsonObject cannot be cast to com.stripe.model.Application
``` 

This seems to be happening because an entry for `Application` model is missing from the `EventDataDeserializer` [`objectMap`](https://github.com/stripe/stripe-java/blob/58df26a289b2719898a33a86068d78233a043ab5/src/main/java/com/stripe/model/EventDataDeserializer.java#L22 ).

As a result, the deserializer seems to [fallback to the `StripeRawJsonObject​` type](https://github.com/stripe/stripe-java/blob/58df26a289b2719898a33a86068d78233a043ab5/src/main/java/com/stripe/model/EventDataDeserializer.java#L140) when trying to deserialize the object data of an `account.application.*` event.

This attempts to fix this issue and also add specific tests for the EventDataDeserializer.

### DIFF
--- a/src/test/java/com/stripe/model/EventDataDeserializerTest.java
+++ b/src/test/java/com/stripe/model/EventDataDeserializerTest.java
@@ -1,0 +1,26 @@
+package com.stripe.model;
+
+import static org.junit.Assert.assertNotNull;
+
+import com.stripe.BaseStripeTest;
+import com.stripe.net.APIResource;
+
+import org.junit.Test;
+
+public class EventDataDeserializerTest extends BaseStripeTest {
+
+  @Test
+  public void testEventAccountApplicationDeauthorized() throws Exception {
+    final String data = getResourceAsString("/api_fixtures/account_application_deauthorized.json");
+    final Event event = APIResource.GSON.fromJson(data, Event.class);
+    assertNotNull(event);
+    assertNotNull(event.getId());
+    assertNotNull(event.getData());
+    assertNotNull(event.getData().getObject());
+
+    final Application application = (Application) event.getData().getObject();
+    assertNotNull(application);
+    assertNotNull(application.getId());
+  }
+
+}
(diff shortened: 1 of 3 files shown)


src/test/java/com/stripe/model/EventDataDeserializerTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.stripe.BaseStripeTest;
+import com.stripe.net.APIResource;
+
+import org.junit.Test;
+
+public class EventDataDeserializerTest extends BaseStripeTest {
+
+  @Test
+  public void testApplicationEventData() throws Exception {
```
ob-stripe: Can you rename this to something like `testEventAccountApplicationDeauthorized` to make it clear that the test is only for a specific event type? (In the future, we could add more tests for different event types.)

src/test/java/com/stripe/model/EventDataDeserializerTest.java
```diff
@@ -0,0 +1,24 @@
...
+    final String data = getResourceAsString("/api_fixtures/account_application_deauthorized.json");
+    final Event event = APIResource.GSON.fromJson(data, Event.class);
+    assertNotNull(event);
+    assertNotNull(event.getId());
+
```
ob-stripe: Can you add a `assertNotNull(event.getData());` and `assertNotNull(event.getData().getObject());` here? That way if either of these is `null` for some reason we'll get a clear assertion error rather than an NPE or cast error in the following line.

spastorelli-stripe: Sounds good. Done.

spastorelli-stripe:
Thanks @ob-stripe for the suggestions and help. I made the suggested changes. PTAL

ob-stripe:
Awesome, thanks @spastorelli-stripe. Mind squashing the commits and removing the WIP tag?

spastorelli-stripe:
Done. Thanks

ob-stripe:
Released in 5.45.0.

PR #543, opened 2018-06-28
Add support for the payment_intent resource

remi-stripe:
This resource is gated so for now tests will be stubbed.

cc @michelle-stripe @jenan-stripe @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/PaymentIntent.java
+++ b/src/main/java/com/stripe/model/PaymentIntent.java
@@ -1,0 +1,311 @@
+package com.stripe.model;
+
+import com.stripe.exception.APIConnectionException;
+import com.stripe.exception.APIException;
+import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.CardException;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.net.APIResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.List;
+import java.util.Map;
+
+import lombok.AccessLevel;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class PaymentIntent extends APIResource implements MetadataStore<PaymentIntent>, HasId {
+  @Getter(onMethod = @__({@Override})) String id;
+  String object;
+  List<String> allowedSourceTypes;
+  Long amount;
+  Long amountCapturable;
+  Long amountReceived;
+  @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Application> application;
+  Long applicationFee;
+  Long canceledAt;
+  String captureMethod;
+  ChargeCollection charges;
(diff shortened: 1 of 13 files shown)


src/main/java/com/stripe/model/PaymentIntentCollection.java
```diff
@@ -0,0 +1,4 @@
+package com.stripe.model;
+
+public class PaymentIntentCollection extends StripeCollection<Payout> {
```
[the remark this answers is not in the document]
remi-stripe: yes it's just a WIP branch to start discussion on naming

src/main/java/com/stripe/model/PaymentIntent.java
```diff
@@ -0,0 +1,285 @@
...
+  String confirmationMethod;
+  Long created;
+  String currency;
+  @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Customer> customer;
+  Boolean livemode;
+  @Getter(onMethod = @__({@Override})) Map<String, String> metadata;
+  PaymentIntentSourceAction nextSourceAction;
+  @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Account> onBehalfOf;
+  String receiptEmail;
+  String returnUrl;
```
remi-stripe: Can you confirm it's the right way to write url and that it does not need anything custom?

src/main/java/com/stripe/model/PaymentIntentSourceActionDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public PaymentIntentSourceAction deserialize(JsonElement json, Type typeOfT,
+      JsonDeserializationContext context)
+      throws JsonParseException {
+    PaymentIntentSourceAction sourceAction = new PaymentIntentSourceAction();
+    JsonObject jsonObject = json.getAsJsonObject();
+    String type = jsonObject.get("type").getAsString();
+    sourceAction.setType(type);
+    Class<? extends PaymentIntentSourceActionValue> cl = objectMap.get(type);
+    if (cl != null) {
```
remi-stripe: added this and confirmed that if I add a wrong `type` the deserializer does not crash.

src/main/java/com/stripe/model/ShippingDetails.java
```diff
@@ -9,6 +9,8 @@
 @EqualsAndHashCode(callSuper = false)
 public final class ShippingDetails extends StripeObject {
   protected Address address;
+  protected String carrier;
   protected String name;
   protected String phone;
+  protected String trackingNumber;
```
remi-stripe: This is a bit unrelated but we did not support those properties even though they are valid on `Order`. They are not valid on `Charge`. We likely should revisit `shipping` as a whole in the API though.

remi-stripe:
@ob-stripe Thank you for the careful review! Damn I missed a few things! Can you have a look?

ob-stripe:
Released in 5.46.0.

Issue #299, opened 2016-06-09
Fileupload from Google Appengine


sowdri:
File objects cannot be created on Google App Engine. So in order to upload documents for id_verification it will be great if FileUpload can support InputStreams such that files from GCS (Google Cloud Storage) or direct multi-part file uploads could be pushed to stripe.

brandur:
+1.

This is very similar to stripe/stripe-python#227, and as noted there, we have a bit of an issue in that backend (thanks to Rack) in that incoming multipart file uploads without a filename are not currently handled. One of my colleagues who is coming back after finishing up school has expressed interesting in fixing this problem though.

I'd suggest we wait for that fix; once it's in, it should be relatively easy to change stripe-java's implementation to take advantage of it.

praboud:
@sowdri as was surfaced in the other issue, you can upload files to Stripe directly from the browser with your publishable API key. [1] IMHO, this is overall better for most use cases than forwarding the file upload through your server, and would also get around the proximate issue ie: that GAE doesn't support creating files.

[1] https://stripe.com/docs/file-upload

sowdri:
Thank you. I'll upload directly from the browser for now.

sowdri:
Basically the problem is that, first the file has to be uploaded to stripe directly via. browser, then I've to get the file_id and send it to the server and then associate it as id_verification document. Doable nonetheless. Otherwise I can do it in a single shot. And since we are a platform we are doing id_verification for multiple clients, it will just be more convenient if we have the support for upload from the server. 

Thank you guys.

praboud:
It's always two requests to Stripe, though - one to upload the file to Stripe, and one to set `account.legal_entity.verification.document` to the file token you got in the first step. This doesn't change whether the upload step is done by the client or the server.

To make explicit what I glossed over previously, doing the file upload from the browser keeps sensitive files off your servers, which makes your security story here easier. Even after stripe-java does support uploading from an `InputStream`, I'd argue that doing the upload from the client is still strictly better than proxying the request through your server.

brandur:
Thanks @praboud!

musby:
Hi I've been searching for ages about how to upload directly from the browser, as is suggested in this case of using Stripe with Google App Engine, so I was hoping I could continue the conversation here on how to accomplish that.

[...]

PR #544, opened 2018-06-29
Allow uploading files from InputStreams and Files

drake-stripe:
This allows users to create a FileUpload with either a `File` or a `InputStream` under the `file` parameter. It defaults to calling the uploaded file "blob".

stripe/stripe-java#299

### DIFF
--- a/src/main/java/com/stripe/net/MultipartProcessor.java
+++ b/src/main/java/com/stripe/net/MultipartProcessor.java
@@ -3,6 +3,7 @@ package com.stripe.net;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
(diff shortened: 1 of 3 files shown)


brandur-stripe:
@drake-stripe Can you take a look at the failed build before we find you a reviewer? It looks like checkstyle is failing:

```
> Task :checkstyleMain
[ant:checkstyle] [ERROR] /home/travis/build/stripe/stripe-java/src/main/java/com/stripe/net/MultipartProcessor.java:75: Line is longer than 100 characters (found 102). [LineLength]
> Task :checkstyleMain FAILED
```

Also, this is a bit of a nit, but those of us who work on these external projects tend to try to keep slightly better commit hygiene than what you might see internally. It might be slightly better to get into a workflow where you pull minor changes/fixes into a single commit with `git commit --amend`, or use `git rebase -i` to fix a series of them. It's not the end of the world either way because we can squash them down before merging here, but it does save us from having to strip all the "fix test", "fix error", etc. cruft from commit messages.

Thanks for taking a crack at this!

ob-stripe:
@drake-stripe Looks good overall, but you'll need to fix the style errors as Brandur mentioned above. You can run both the test suite and the linter locally with `./gradlew check` (or `./gradlew checkstyleMain checkstyleTest` if you just want to run the linter).

Also, could you add a test case in the functional test for `FileUpload`? You can copy the `testCreate` function into a new `testCreateWithStream` case.

drake-stripe:
PR now passes linter, and there is a functional test implemented for InputStream uploading. I also squashed it into one commit with a more descriptive commit message.

src/main/java/com/stripe/net/MultipartProcessor.java
```diff
@@ -59,20 +60,42 @@ public void addFormField(String name, String value) {
...
    */
   public void addFileField(String name, File file) throws IOException {
-    String fileName = file.getName();
-    writer.append("--" + boundary).append(LINE_BREAK);
-    writer.append(
-        "Content-Disposition: form-data; name=\"" + name
-            + "\"; filename=\"" + fileName + "\"").append(
-        LINE_BREAK);
+    addFileField(name, file.getName(), new FileInputStream(file));
+  }
```
brandur-stripe: Minor, but since these methods are meant to be internal anyway, can we just get rid of the `File` form of `addFileField` since it's only used in one place anyway? If other places in the library need to do this, they can instantiate their own `FileInputStream`.

brandur-stripe:
Left a minor comment, but thanks a lot for fixing all that stuff Drake!

(Olivier, you're a lot better learned in Java than I am, so I'll leave a final review to you.)

src/main/java/com/stripe/net/MultipartProcessor.java
drake-stripe: Done.

ob-stripe:
Released in 5.48.0.

PR #546, opened 2018-07-02
Add setClientId() method to RequestOptionsBuilder

### LINKED ISSUE #545: Missing setClientId in RequestOptionsBuilder
Hi,

I am trying to use the class `RequestOptionsBuilder` to build my environments with differents apiKey and clientId.

The problem is that RequestOptionsBuilder always use `clientId = Stripe.clientId`. There is no setter to modify it.

Why?

ob-stripe:
Fixes #545.

Not sure why the setter was missing in the first place...

cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -131,10 +131,6 @@ public class RequestOptions {
       return apiKey;
     }
 
-    public String getClientId() {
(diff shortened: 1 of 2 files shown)


brandur-stripe:
> Not sure why the setter was missing in the first place...

lol, so weird that the getter was present, but not the setter.

LGTM.

PR #547, opened 2018-07-02
[wip] Move nested objects into inner static classes

ob-stripe:

For nested objects ("object" being defined as a map, not an API resource with an `object` attribute) that belong to a single API resource, declare them as inner static classes in the owner resource rather than a dedicated model class.

Also, if the class name's was prefixed with the resource's name, remove the prefix (e.g. `AccountDeclineChargeOn` is now `Account.DeclineChargeOn`).

Nested objects that are used by more than one resource are unaffected (e.g. `Address`).

Tagged as WIP because I'm mostly looking for comments for now. What do you think? It does declutter the `model` package at the expense of cluttering the resource classes.


src/main/java/com/stripe/model/Charge.java
```diff
@@ -493,4 +493,104 @@ public Charge update(Map<String, Object> params, RequestOptions options) throws
...
+
+    public void setRuleObject(Rule rule) {
+      this.rule = new ExpandableField<Rule>(rule.getId(), rule);
+    }
+    // </editor-fold>
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class Rule extends StripeObject implements HasId {
```
ob-stripe: Note that this class is doubly nested: `ChargeOutcomeRule` is now `Charge.Outcome.Rule`.

brandur-stripe:
> Tagged as WIP because I'm mostly looking for comments for now. What do you think? It does declutter the model package at the expense of cluttering the resource classes.

Personally I like it!

I'm always really disliked how a lot of subresources that are very obviously related to a parent (like many of the examples already in this PR) are not colocated with their parent very well in that you need to open another file and sometimes the subresource is not alphabetically ordered with its parent.

There might be some complications around subresources that are shared like you mentioned with `Address` — occasionally these might appear as linked to only one class originally, but then be introduced in another class, so we'd have to promote them to the top level. I don't think it happens often enough to be a major problem though.

ob-stripe:
@remi-stripe Any feels on this? Should I go ahead and merge?

src/main/java/com/stripe/model/Charge.java
```diff
@@ -493,4 +493,104 @@ public Charge update(Map<String, Object> params, RequestOptions options) throws
...
+    List<LineItem> lineItems;
+    String merchantReference;
+    String shippingAddressZip;
+    String shippingFromZip;
+    Long shippingAmount;
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class LineItem extends StripeObject {
```
remi-stripe: I wonder if we might end up with classes way too deep because of this and if this hurts readability. Just flagging while reviewing.

src/main/java/com/stripe/model/Customer.java
```diff
@@ -215,4 +215,12 @@ public Customer update(Map<String, Object> params, RequestOptions options)
         Customer.class, options);
   }
   // </editor-fold>
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class NextRecurringCharge extends StripeObject {
```
remi-stripe: I feel like this is the kind of property we should stop supporting fully. I did not even know this existed

src/main/java/com/stripe/model/BalanceTransaction.java
```diff
@@ -100,4 +100,15 @@ public static BalanceTransaction retrieve(String id, RequestOptions options)
     return request(RequestMethod.GET, url, null, BalanceTransaction.class, options);
   }
   // </editor-fold>
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class Fee extends ApiResource {
```
remi-stripe: This one is an `ApiResource`, is it expected that it was moved?

src/main/java/com/stripe/model/Transfer.java
```diff
@@ -303,4 +303,21 @@ public Transfer update(Map<String, Object> params, RequestOptions options)
         Transfer.class, options);
   }
   // </editor-fold>
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class Summary extends StripeObject {
```
remi-stripe: Similar as the other one, it's likely time to remove support for this one entirely too. This is unrelated to this PR for sure but it's the best opportunity I found to flag some of those

src/main/java/com/stripe/model/Account.java
```diff
@@ -236,4 +234,43 @@ public Account update(Map<String, Object> params, RequestOptions options) throws
...
+  public static class TosAcceptance extends StripeObject {
+    Long date;
+    String ip;
+    String userAgent;
+  }
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class TransferSchedule extends StripeObject {
```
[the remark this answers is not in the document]
ob-stripe: Java doesn't support type aliasing, so I can't think of a good solution here unfortunately. Besides, these two sub-resources are declared separately in the OpenAPI spec, so I think having separate classes is the right call.

src/main/java/com/stripe/model/Address.java
```diff
@@ -3,17 +3,15 @@
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
-import lombok.experimental.Accessors;
 
 @Getter
 @Setter
-@Accessors(chain = true)
```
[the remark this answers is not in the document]
ob-stripe: Oops, yeah, this is not really related to this PR. Before we switched to Lombok, the setters on this class returned `this` which allowed for chaining (`address.setFoo("foo).setBar("bar")`). I used the `@Accessors` annotation to maintain this behavior and guarantee 100% compatibility.

This isn't the convention used in the rest of the library though, so I did some cleanup here and standardized the behavior with that of all other object classes.

Technically a breaking change, but this only affects setters which users are not supposed to use anyway.

src/main/java/com/stripe/model/Address.java
```diff
@@ -3,17 +3,15 @@
...
-  protected String city;
-  protected String country;
-  protected String line1;
-  protected String line2;
-  protected String postalCode;
-  protected String state;
+  String city;
+  String country;
+  String line1;
+  String line2;
```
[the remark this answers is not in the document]
ob-stripe: More standardization across objects. All API attributes are declared without a visibility modifier so they are "package-protected" (like protected, except that the attributes are not visible to subclasses in different packages).

src/main/java/com/stripe/model/Balance.java
```diff
@@ -34,4 +34,23 @@ public static Balance retrieve(RequestOptions options) throws StripeException {
     return request(RequestMethod.GET, singleClassUrl(Balance.class), null, Balance.class, options);
   }
   // </editor-fold>
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class Money extends StripeObject {
```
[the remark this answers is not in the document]
ob-stripe: Not sure. We had to call it something I guess :)

src/main/java/com/stripe/model/Charge.java
ob-stripe: Yeah, that's my main concern with this approach, and is why I left `LegalEntity` as a standalone class.

src/main/java/com/stripe/model/BalanceTransaction.java
ob-stripe: It shouldn't be -- nested objects should all be `StripeObject`s, not `ApiResource`s.

src/main/java/com/stripe/model/Transfer.java
ob-stripe: It does mean the library would become partially incompatible with those very old API versions though.

remi-stripe: IMO it already is in the sense that there are so many old cases we have no idea exist and might have broken or never supported though.
I'm fine leaving as it, but I do worry that we don't have any tests for things like this and end up thinking we support it when we might not be doing it properly

ob-stripe:
Okay, this is not perfect but I think we all agree that it's going in the right direction, so pulling this in.

codelance:
I know this thread is a bit stale... I am just now starting to go through and migrate through 6.0.0+ on this SDK. Might I suggest that as you create these pull request with breaking changes it might make it easy to also document quick migration hints that way when the next release comes, you can just roll up all the hints for any one to easily grab. I've found myself having to inspect the commits to see what and how it changed in order to efficiently update my code. Just my thoughts. Thanks for all the improvements/enhancements...  I'm just a friendly consumer :)

ob-stripe:
Hi @codelance. We use [semantic versioning](https://semver.org/) for our libraries, meaning breaking changes are always released with major version bumps.

When upgrading to a new major version, you should always check the [CHANGELOG](https://github.com/stripe/stripe-java/blob/master/CHANGELOG.md) to see what changes you should be looking out for and what might need updating in your own code.

The last major version included a larger number of changes than usual, so we pulled together a quick migration guide that you can find [here](https://github.com/stripe/stripe-java/wiki/Migration-guide-for-v6).

PR #552, opened 2018-07-11
Rename application_fee to application_fee_amount on PaymentIntent

remi-stripe:
We are renaming `application_fee` to `application_fee_amount` as a parameter and property. This is cleaner as `application_fee` usually is an Application Fee id and can be expanded.

We have confirmed no Java users have used this parameter yet so while a breaking change we can merge and release as the change is being deployed.

r? @brandur-stripe @ob-stripe 
cc @stripe/api-libraries 
cc @michelle-stripe @ln-stripe

### DIFF
--- a/src/main/java/com/stripe/model/PaymentIntent.java
+++ b/src/main/java/com/stripe/model/PaymentIntent.java
@@ -27,7 +27,7 @@ public class PaymentIntent extends APIResource implements MetadataStore<PaymentI
   Long amountCapturable;
   Long amountReceived;
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Application> application;
-  Long applicationFee;
+  Long applicationFeeAmount;
   Long canceledAt;
   String captureMethod;
   ChargeCollection charges;


brandur-stripe:
LGTM.

brandur-stripe:
> We have confirmed no Java users have used this parameter yet so while a breaking change we can merge and release as the change is being deployed.

Just for clarity: are you okay with it if I just release this now? (It seems unlikely that this feature will pick up any new users between now and when the change is deployed to the API.)

remi-stripe:
@brandur-stripe yeah for once I would just merge early since the change is being deployed to prod soon! Thank you for the quick review.
Also, at _worst_ the bug is that they can't see the app fee on the resource and since it's a beta, they would just ask why and we could explain quickly.

brandur-stripe:
> Also, at worst the bug is that they can't see the app fee on the resource and since it's a beta, they would just ask why and we could explain quickly.

Yep, agreed. Okay, thanks!

brandur-stripe:
Released as 5.50.0.

PR #553, opened 2018-07-23
Remove StripeDeleted abstraction

spastorelli-stripe:
Remove `Deleted*` models and add a Boolean deleted property to corresponding models.
This attempts to fix https://github.com/stripe/stripe-java/issues/409 

r? @ob-stripe 
cc @stripe/api-libraries


src/test/java/com/stripe/functional/BankAccountTest.java
```diff
@@ -135,9 +135,14 @@ public void testDelete() throws StripeException {
...
 
-    final DeletedBankAccount deletedBankAccount = bankAccount.delete();
+    final BankAccount deletedBankAccount = bankAccount.delete();
 
     assertNotNull(deletedBankAccount);
+    /**
+     * TODO(spastorelli): stripe-mock currently doesn't include deleted:true in the response of a
+     * DELETE  /v1/customers/cus_1234/sources/* call. Uncomment below assertion when it does.
+     */
+    //assertTrue(deletedBankAccount.getDeleted());
```
ob-stripe: @brandur-stripe Do you have a sense of how difficult it would be to handle this correctly in stripe-mock?

src/test/java/com/stripe/functional/CardTest.java
```diff
@@ -136,14 +137,23 @@ public void testList() throws IOException, StripeException {
     assertNotNull(card);
   }
 
+  /**
+   * TODO(spastorelli): stripe-mock returns a bank account object when calling "Delete Card".
+   *  i.e DELETE  /v1/customers/cus_1234/sources/card_123
+   *  So object is deserialized into a BankAccount and since we now return a Card for delete() as
+   *  a result the test fails with a
+   *  ClassCastException (Cannot cast com.stripe.model.BankAccount to com.stripe.model.Card).
```
ob-stripe: Right. It's probably going to be a while until stripe-mock handles this one correctly unfortunately, so I would simply do something similar to the list test just above and mock the fixture.

Since the JSON is so simple, you don't even need an external file, you could do something like this:
```java
     stubRequest(
         ApiResource.RequestMethod.DELETE,
         String.format("/v1/customers/%s/sources/%s", customer.getId(), card.getId()),
         null,
         Card.class,
         String.format("{\"id\": \"%s\", \"object\": \"card\", \"deleted\": true}", card.getId())
     );
```

(untested, but you get the idea)

src/test/java/com/stripe/functional/BankAccountTest.java
brandur-stripe: So the basic problem here is that the endpoint sometimes returns non-deleted objects (I believe deleting a source "detaches" it, but it's still a live object):

```
      responses:
        '200':
          content:
            application/json:
              schema:
                anyOf:
                - "$ref": "#/components/schemas/customer_source"
                - "$ref": "#/components/schemas/deleted_bank_account"
                - "$ref": "#/components/schemas/deleted_object"
                - "$ref": "#/components/schemas/deleted_card"
```

So stripe-mock is just doing the dumbest thing: picking the first branch, and returning that. If it had picked any of the last three, there would be a `deleted` field.

The only good fix I can think of is to try and recognize the resource type based on the input ID, and that'll require a little bit of work.

ob-stripe: Ack. @spastorelli-stripe I recommend that you stub this request as well for now.

spastorelli-stripe: Thanks @ob-stripe and @brandur-stripe for your input and suggestions. I've stubbed the request.

src/test/java/com/stripe/functional/CardTest.java
spastorelli-stripe: Thanks @ob-stripe. I've stubbed the request as suggested.

ob-stripe:
@spastorelli-stripe Apologies, I rebased the `integration-v6` branch on `master` to include all of the issuing stuff released yesterday. Some of the commits were modified, which means now rebasing your branch on `integration-v6` will be a pain :/ It might be simpler to start a new branch and cherry-pick your commit.

That said, the PR looks good to me. Thanks for adding a test specifically for the expandable + deleted case!

cc @brandur-stripe @tmaxwell-stripe @drake-stripe as we recently talked about the shape of deleted objects in our libraries

spastorelli-stripe:
Thanks @ob-stripe. PTAL

PR #556, opened 2018-07-25
Add risk score property to ChargeOutcome resource

spastorelli-stripe:
Add the risk score property to ChargeOutcome.

r? 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/ChargeOutcome.java
+++ b/src/main/java/com/stripe/model/ChargeOutcome.java
@@ -14,6 +14,7 @@ public class ChargeOutcome extends APIResource {
   protected String networkStatus;
   protected String reason;
   protected String riskLevel;
+  protected Long riskScore;
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE)
       protected ExpandableField<ChargeOutcomeRule> rule;
   protected String sellerMessage;
--- a/src/test/resources/api_fixtures/charge_outcome.json
+++ b/src/test/resources/api_fixtures/charge_outcome.json
@@ -2,6 +2,7 @@
   "network_status": "approved_by_network",
   "reason": null,
   "risk_level": "normal",
+  "risk_score": 50,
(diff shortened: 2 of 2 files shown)


spastorelli-stripe:
👋@ob-stripe, I've used for now  `Integer` as the type for `riskScore`. What do you think would be best to use for this: `Long`, `Double` or `Integer`?

ob-stripe:
@spastorelli-stripe Please use `Long` instead. Currently we use a mix of both `Integer` and `Long`, but in the upcoming v6 we've updated all numeric types to `Long` for integers and `Double` for decimals.

remi-stripe:
@ob-stripe  The issue is that if we ever make this a double in the API (does not sound farfetched IMO) then it's broken and we have the conundrum we just had with `percent_off`. Should we play it safe and just use a `Double`?

ob-stripe:
> @ob-stripe The issue is that if we ever make this a double in the API (does not sound farfetched IMO) then it's broken and we have the conundrum we just had with percent_off. Should we play it safe and just use a Double?

I'd say no. It's not the client libraries' place to try and guess what the API will look like in the future. Also, floating point math comes with its own can of worms:
```
jshell> Double a = 0.1;
a ==> 0.1

jshell> Double b = 0.2;
b ==> 0.2

jshell> Double c = 0.3;
c ==> 0.3

jshell> (a + b) == c;
$4 ==> false
```
so let's not start storing integer values into floats just because we think there's a chance we might start returning decimals in the future.

remi-stripe:
I disagree but it's fine, I've confirmed that this value will stay an integer for the foreseeable future anyway!

spastorelli-stripe:
Thanks @ob-stripe and @remi-stripe. I've updated the type.

remi-stripe:
LGTM cc @ob-stripe as we should release that with Issuing!

ob-stripe:
Released as 5.52.0.

PR #558, opened 2018-07-27
Use `BigDecimal` for floating point values

ob-stripe:

Use `BigDecimal` for floating-point values instead of `Double`.

See discussion here for context: https://github.com/stripe/stripe-java/pull/533#issuecomment-408380799.


brandur-stripe:
I'd echo Remi saying that basic operations seem to become pretty painful with `BigDecimal`, but the evidence that this is the right thing for money types seems to be pretty comprehensive.

No strong opinion from me! LGTM if you want to do it.

remi-stripe:
Also seems fairly painful to me but I still assume most people would not be hurt by it since only limited fields are with BigDecimal.

We'll have to remember to fix the coupon one when it's released

PR #564, opened 2018-07-31
Properly support object on list resources

remi-stripe:
This ensures that the `object` property is kept when serializing a list. This is important when you serialize a Customer object for example so that it gets deserialized in the mobile SDKs where they expect the list resource to have a valid object key.

Not sure if it's a breaking change on that one so flagging just in case.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/StripeCollection.java
+++ b/src/main/java/com/stripe/model/StripeCollection.java
@@ -33,6 +33,7 @@ import java.util.Map;
  */
 public abstract class StripeCollection<T extends HasId> extends StripeObject
     implements StripeCollectionInterface<T> {
+  String object;
   List<T> data;
   Long totalCount;
   Boolean hasMore;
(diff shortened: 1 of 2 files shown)


anelder-stripe:
Tested an old library against this new one to be sure that this was actually working:

### Older Version
```json
  "sources": {
    "data": [...],
    "total_count": 5,
    "has_more": false,
    "request_options": null,
    "request_params": null,
    "url": "/v1/customers/cus_Ci752i07jWGYLP/sources",
    "count": null
  },
```

### New Version
```
  "sources": {
    "object": "list",
    "data": [],
    "total_count": 5,
    "has_more": false,
    "request_options": null,
    "request_params": null,
    "url": "/v1/customers/cus_Ci752i07jWGYLP/sources",
    "count": null
  },
```

This change looks good to me!  👍  Will yield the floor to @ob-stripe for final approval though.

Thanks for doing this @remi-stripe, it should solve our larger issue.

PR #565, opened 2018-08-01
Add a retrieve() signature with params to all resources

### LINKED ISSUE #559: [Feature Request] Plan.retrieve implementation that supports expanding fields.
The Stripe docs demonstrate how to do expansion:

```java
Stripe.apiKey = "{key}";

List<String> expandList = new LinkedList<String>();
expandList.add("customer");

params.put("expand", expandList);

Charge.retrieve("{charge_id}", params, null);
```

And Plans have `product` as an expandable field. But, there is no method signature for retrieving a Plan with params like there is for Charge.

At the moment, I'm working around it by subclassing, but it would be nice if I didn't have to:

```java
import com.stripe.exception.*;
import com.stripe.model.Plan;
import com.stripe.net.RequestOptions;

import java.util.Map;

public class StripePlan extends Plan {
    /**
     * Retrieve a plan.
     */
    public static Plan retrieve(String id, Map<String, Object> params, RequestOptions options)
        return request(RequestMethod.GET, instanceURL(Plan.class, id), params, Plan.class, options);
    }
}
```

ob-stripe:
cc @stripe/api-libraries 

Adds a `retrieve()` signature with params to all resources.

Fixes #559.

### DIFF
--- a/src/main/java/com/stripe/model/Charge.java
+++ b/src/main/java/com/stripe/model/Charge.java
@@ -463,7 +463,7 @@ public class Charge extends ApiResource implements MetadataStore<Charge>, HasId 
    * Retrieve a charge.
    */
   public static Charge retrieve(String id, RequestOptions options) throws StripeException {
-    return request(RequestMethod.GET, instanceUrl(Charge.class, id), null, Charge.class, options);
+    return retrieve(id, null, options);
   }
 
   /**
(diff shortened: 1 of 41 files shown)


src/main/java/com/stripe/model/Charge.java
```diff
@@ -463,7 +463,7 @@ public static Charge retrieve(String id) throws StripeException {
    * Retrieve a charge.
    */
   public static Charge retrieve(String id, RequestOptions options) throws StripeException {
-    return request(RequestMethod.GET, instanceUrl(Charge.class, id), null, Charge.class, options);
+    return retrieve(id, null, options);
```
remi-stripe: Minor but you end up piping through 3 methods instead of 2 because the one with just the id will call that one which then calls the one with 3 parameters. Definitely not a big deal but wanted to mention it

ob-stripe:
Released as 6.2.0.

PR #567, opened 2018-08-02
Use delomboked sources when creating sources JAR

### LINKED ISSUE #555: Sources do not match byte code
Hi
The sources provided through maven for version 5.49+ do not match the byte code. This is most likely the result of lombok generations. Project maintainer may want to use [lombok maven plugin](http://awhitford.github.io/lombok.maven/lombok-maven-plugin/faq.html) 
Thank you.

ob-stripe:
cc @stripe/api-libraries @lure

Use delomboked sources instead of original files when creating sources JAR.

I've tested that this works by manually creating the JAR (with `gradle clean sourcesJar`) then uncompressing it (JARs are actually just ZIP files) and checking the sources.

Fixes #555.

### DIFF
--- a/deploy.gradle
+++ b/deploy.gradle
@@ -108,9 +108,9 @@ afterEvaluate { project ->
         from makeJavadocs.destinationDir
     }
 
-    task sourcesJar(type: Jar) {
+    task sourcesJar(type: Jar, dependsOn: delombok) {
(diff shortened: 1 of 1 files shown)


ob-stripe:
Released as 6.2.1.

PR #579, opened 2018-08-28
Add support for authorization_code on the Charge resource

remi-stripe:
Note that this property is internal and not usually returned in the API.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as 6.8.0.

PR #580, opened 2018-08-29
Update attributes for Account and LegalEntity

ob-stripe:
cc @stripe/api-libraries @scherr-stripe 

Updates attributes for the `Account` and `LegalEntity` classes. A lot of attributes were missing, particularly the Japan-specific ones.

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -27,16 +27,18 @@ public class Account extends ApiResource implements HasId, MetadataStore<Account
   Boolean debitNegativeBalances;
   DeclineChargeOn declineChargeOn;
   String defaultCurrency;
+  Boolean deleted;
   Boolean detailsSubmitted;
   String displayName;
-  LoginLinkCollection loginLinks;
   String email;
   ExternalAccountCollection externalAccounts;
   Keys keys;
   LegalEntity legalEntity;
+  LoginLinkCollection loginLinks;
   @Getter(onMethod = @__({@Override})) Map<String, String> metadata;
-  Boolean payoutsEnabled;
   PayoutSchedule payoutSchedule;
+  String payoutStatementDescriptor;
+  Boolean payoutsEnabled;
   String productDescription;
   String statementDescriptor;
   String supportEmail;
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/model/Account.java
```diff
@@ -17,6 +17,10 @@
 public class Account extends ApiResource implements HasId, MetadataStore<Account> {
   @Getter(onMethod = @__({@Override})) String id;
   String object;
+  String applicationIcon;
```
remi-stripe: This is not in the docs for me. None of the `application*` are and it's not something we surface in the other bindings. We should remove

src/main/java/com/stripe/model/LegalEntity.java
```diff
@@ -37,12 +52,14 @@
   @Getter
   @Setter
   @EqualsAndHashCode(callSuper = false)
-  public static class Verification extends StripeObject {
-    String details;
-    String detailsCode;
-    String document;
-    String documentBack;
-    String status;
+  public static class JapanAddress extends StripeObject {
```
remi-stripe: why not extend `Address` and just defined `town` as extra?

src/main/java/com/stripe/model/Account.java
ob-stripe: The `application_*` attributes are in the OpenAPI spec. I can remove them as they don't seem all that useful today, but on the other hand it doesn't cost us anything to keep them and they'll be available if someone ever needs to use them.

src/main/java/com/stripe/model/LegalEntity.java
ob-stripe: Internally and in the OpenAPI spec, they're defined as two different schemas.

Also, at the moment there is no concrete resource class that derives from another concrete class, and I'd rather keep that invariant in place. If we had `JapanAddress` extends `Address`, and ever needed to add a field to `Address`, chances are whoever makes the change will not notice that the field would also be added to `JapanAddress`.

src/main/java/com/stripe/model/Account.java
remi-stripe: I'm in favour of removing. I have never heard of those properties before. They seem more for the dashboard than for the API to be honest. If we added them here we'd have to add them to all the other ones and keep maintaining them.

src/main/java/com/stripe/model/LegalEntity.java
remi-stripe: Gotcha, I'm fine with that in that case.

src/main/java/com/stripe/model/Account.java
ob-stripe: Actually, these attributes can be useful for Connect platforms that use destination charges and their users. Retrieving the `py_` charge object on the destination account returns these attributes in the `source` attribute:
```
  "id": "py_1D4SzzJBZTufeQX6VjXfw9sr",
  "object": "charge",
  ...
  "source": {
    "application_name": "Planet Express",
    "application_url": "http://localhost:3000",
    "id": "acct_16bewcKCFFPkgtRh",
    "object": "account"
  },
  ...
```

Unlike most other attributes, they're only returned if they have non-null values, which is why `application_icon` and `application_logo` are missing in the example above.

Are you still in favor of removing them?

remi-stripe: huh why did we do something like this, that does not make sense. The charge already has `application` as a hash for that. This information should live *there* in that case. I still think we should at least discuss this internally and do this in a separate PR

ob-stripe: 👍

ob-stripe:
ptal @remi-stripe

ob-stripe:
Released as 6.9.0.

PR #591, opened 2018-09-17
Cleanup

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Minor cleanup:
- removed unused imports and variables
- fixed a case where a stream could potentially not be closed with Lombok's `@Cleanup` annotation


brandur-stripe:
LGTM!

> fixed a case where a stream could potentially not be closed with Lombok's `@Cleanup` annotation

Nice. It's cool that Lombok can do this, although it's a little unfortunate that there's no better mechanism built into Java!

ob-stripe:
Thanks Brandur!

> It's cool that Lombok can do this, although it's a little unfortunate that there's no better mechanism built into Java!

Yep. The Java way is to use a `try/finally` block (which is exactly what `@Cleanup` does under the hood). Very verbose, though that's to be expected with Java :)

PR #593, opened 2018-09-18
Replace `FileUpload` class with `File` class

ob-stripe:
cc @stripe/api-libraries 

Replaces the `FileUpload` class with the new `File` class, which is mostly identical to the old one except that:
- it has an additional `title` attribute
- requests for the `retrieve` and `list` methods are sent to `api.stripe.com` instead of `files.stripe.com`

Unfortunately, this is a breaking change. I tried creating a `FileUpload` subclass to act as a backwards compatibility layer, but I ran into an issue with the `list` methods: in Java, a `SubClass` is a `Parent`, but a `List<SubClass>` is not a `List<Parent>`. Even when forcing an unsafe cast (which would be more or less safe here, considering that the returned lists are not supposed to be mutable), there isn't (AFAIK) any way to make the `list` methods in the subclass covariant with that of the superclass.


src/test/java/com/stripe/functional/FileTest.java
```diff
@@ -4,8 +4,8 @@
 
 import com.stripe.BaseStripeTest;
 import com.stripe.exception.StripeException;
-import com.stripe.model.FileUpload;
-import com.stripe.model.FileUploadCollection;
+//import com.stripe.model.File;
```
brandur-stripe: It looks like you can into problems with a naming conflict on `File` or something? Maybe just removed this commented out import if it's the easiest thing.

ob-stripe:
Released as 7.0.0.

PR #598, opened 2018-09-22
Compatibility with latest API version

ob-stripe:
cc @stripe/api-libraries 

The library is incompatible with the latest API version, because `percent_off` on coupons is now a float.

Fix this by changing the type of the `percentOff` attribute from `Long` to `BigDecimal`. For users on older API versions, they will simply get an integer value in the `BigDecimal` instance. They can access the exact value via the `percentOffPrecise` attribute.

Also, it looks like the `transaction` attribute on issuing dispute objects was renamed to `disputed_transaction`, so fix that as well.

Unfortunately, changing the type of `percentOff` is a breaking change, but I was going to release #593 which also breaks BC anyway. I set the target branch for this PR to `integration-v7` so that we can release both changes at once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
 
 env:
   global:
-  - STRIPE_MOCK_VERSION=0.30.0
+  - STRIPE_MOCK_VERSION=0.32.0
 
 matrix:
   include:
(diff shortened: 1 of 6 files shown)


ob-stripe:
Released as 7.0.0.

PR #600, opened 2018-09-25
Start testing Java 11

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Java 11 was released today. It's a LTS version too, so people are likely to adopt it. Let's start testing it right now :)

Travis now supports all JDKs, so I was able to remove the custom script we used to install missing JDKs.

I also added EA (Early Access) JDKs to the build matrix, with failures allowed, to help add support for future versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ addons:
   hostname: short-hostname
 
 before_install:
-  # Install script for missing JDKs
-  - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh
   # Work around missing crypto in openjdk7
   - |
     if [ "$TRAVIS_JDK_VERSION" == "openjdk7" ]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -61,7 +55,7 @@ script:
   - ./gradlew --version
   - ./gradlew clean
   - if [ "$TRAVIS_JDK_VERSION" == "oraclejdk9" ]; then ./gradlew check; fi
-  - ./gradlew cobertura coveralls
+  - ./gradlew -Djdk.tls.client.protocols="TLSv1.2" cobertura coveralls
```
ob-stripe: I had to add this option, otherwise uploading the report to coveralls.io fails with an SSL error with Java 11. My best guess is that Java 11 tries to use TLS 1.3 and doesn't fall back to 1.2.

brandur-stripe: Wow, Java 11 is just too advanced!

.travis.yml
```diff
@@ -34,24 +32,20 @@ env:
...
+    - jdk: openjdk9
+    - jdk: oraclejdk10
+    - jdk: openjdk10
+    - jdk: oraclejdk11
+    - jdk: openjdk11
+    - jdk: oraclejdk-ea
+    - jdk: openjdk-ea
+  allow_failures:
+    - jdk: oraclejdk-ea
+    - jdk: openjdk-ea
```
brandur-stripe: So much better!

brandur-stripe:
LGTM.

PR #601, opened 2018-09-28
logging friendly client-user-agent header

mickjermsurawong-stripe:
- Format client-user-agent key to be friendly for logging analysis
- This is consistent with other language bindings
  - dotnet https://github.com/stripe/stripe-dotnet/blob/5b6a010515e94cd72f6bef4659350922f023d2cf/src/Stripe.net/Infrastructure/Client.cs#L38
  - ruby https://github.com/stripe/stripe-ruby/blob/156145b96b3f3196c9c14a5c0401fea3458413df/lib/stripe/stripe_client.rb#L586

r? @remi-stripe @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -125,29 +125,7 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
 
     headers.put("Authorization", String.format("Bearer %s", options.getApiKey()));
 
-    // debug headers
-    String[] propertyNames = {"os.name", "os.version", "os.arch",
-        "java.version", "java.vendor", "java.vm.version",
-        "java.vm.vendor"};
-    Map<String, String> propertyMap = new HashMap<String, String>();
-    for (String propertyName : propertyNames) {
-      propertyMap.put(propertyName, System.getProperty(propertyName));
-    }
-    propertyMap.put("bindings.version", Stripe.VERSION);
-    propertyMap.put("lang", "Java");
-    propertyMap.put("publisher", "Stripe");
-
-    Set<String> propertyKeySet = new HashSet<>(propertyMap.keySet());
-    for (String propertyKey: propertyKeySet) {
-      String property = propertyMap.get(propertyKey);
-      propertyMap.put(propertyKey.replace('.', '_'), property);
-      propertyMap.remove(propertyKey);
-    }
-
-    if (Stripe.getAppInfo() != null) {
-      propertyMap.put("application", ApiResource.GSON.toJson(Stripe.getAppInfo()));
-    }
-    headers.put("X-Stripe-Client-User-Agent", ApiResource.GSON.toJson(propertyMap));
+    headers.put("X-Stripe-Client-User-Agent", ApiResource.GSON.toJson(buildClientUserAgentMap()));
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -134,6 +136,14 @@ static String formatAppInfo(Map<String, String> info) {
...
     propertyMap.put("lang", "Java");
     propertyMap.put("publisher", "Stripe");
+
+    Set<String> propertyKeySet = new HashSet<>(propertyMap.keySet());
+    for (String propertyKey: propertyKeySet) {
+      String property = propertyMap.get(propertyKey);
+      propertyMap.put(propertyKey.replace('.', '_'), property);
+      propertyMap.remove(propertyKey);
+    }
+
```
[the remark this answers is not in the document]
brandur-stripe: +1, and a comment please! Even after reading this PR's description, the code, the test, and following the relevant links, I still don't think I 100% understand what we're doing here exactly.

If the intent is just to transform `bindings.version` to `bindings_version`, should we just explicitly set that one property instead?

mickjermsurawong-stripe:
hi @fred-stripe @brandur-stripe, I made the following changes.
- Added more explicit comment in the implementation
- I did not simply change one value for `binding_versions` because besides making it consistent with other libraries, the other intended change here is to remove "." in all keys because in splunk treats them like nested key. I also didn't remove checking for "." because I want to catch unexpected implication of nested structure.
- Added more assertions

mickjermsurawong-stripe:
ptal @fred-stripe

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -150,6 +138,34 @@ static String formatAppInfo(Map<String, String> info) {
...
+    // Format keys to be friendly for log analysis and consistent with other client libraries
+    // Key containing '.' is considered nested in Splunk, but that's not the case here,
+    // so we are replacing it with '_'
+    Set<String> keySet = new HashSet<>(clientUserAgentMap.keySet());
+    for (String key: keySet) {
+      String value = clientUserAgentMap.get(key);
+      clientUserAgentMap.remove(key);
+      String formattedKey = key.replace('.', '_');
```
fred-stripe: This library's programming style dictates that this variable definition should be at the top of this block, after the line that declares `value`. Can you either move this line up omit the variable and put the `key.replace()` call directly onto the `clientUserAgentMap.put()` call?

mickjermsurawong-stripe: Moved the line up to the top of this block.
Thanks Fred for the careful review!

mickjermsurawong-stripe:
Will come back to merge this and handle the release, so I'm leaving it here first.
Thank you Brandur and Fred!

mickjermsurawong-stripe:
This should not be changed until the consumer of headers is adjusted here..
I had thought it's only for splunk analysis. The inconsistency that this PR aims to address is still a problem expressed by comment in the consumer here. https://git.corp.stripe.com/stripe-internal/pay-server/blob/HEAD/api/lib/middleware/api_request_logger.rb#L434
I will close this PR first, and will revisit it again  -- creating a ticket for this work  https://jira.corp.stripe.com/browse/DX-2171

PR #602, opened 2018-10-10
Adds support for 'partner_id' to 'setAppInfo'

zwick-stripe:
This adds support for the `partner_id` parameter in the `setAppInfo` API for use by participants in the Stripe Partner Program. Similar patches landed in the other Stripe API client libraries, but I didn't add support for Java at that time because the Stripe docs seemed to indicate that SetAppInfo wasn't supported in Java.
I will also be updating the Stripe docs at https://stripe.com/docs/building-plugins#setappinfo to show how to use this part of the Java bindings.


remi-stripe:
cc @stripe/api-libraries

PR #603, opened 2018-10-12
Add support for the Person resource

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -144,6 +144,33 @@ public class Account extends ApiResource implements HasId, MetadataStore<Account
   }
   // </editor-fold>
 
+  // <editor-fold desc="persons">
+  /**
+   * List an account's persons.
+   */
+  public PersonCollection persons()
+      throws StripeException {
+    return persons(null, null);
+  }
+
+  /**
+   * List an account's persons.
+   */
+  public PersonCollection persons(Map<String, Object> params)
+      throws StripeException {
+    return persons(params, null);
+  }
+
+  /**
+   * List an account's persons.
+   */
+  public PersonCollection persons(Map<String, Object> params,
+      RequestOptions options) throws StripeException {
+    String url = instanceUrl(Account.class, this.getId()) + "/persons";
+    return requestCollection(url, params, PersonCollection.class, options);
+  }
+  // </editor-fold>
+
   // <editor-fold desc="reject">
   /**
    * Reject an account.
--- a/src/main/java/com/stripe/model/Person.java
+++ b/src/main/java/com/stripe/model/Person.java
@@ -1,0 +1,148 @@
+package com.stripe.model;
+
+import com.google.gson.annotations.SerializedName;
+
+import com.stripe.exception.StripeException;
+import com.stripe.net.ApiResource;
+import com.stripe.net.RequestOptions;
+
+import java.math.BigDecimal;
+import java.util.List;
+import java.util.Map;
+
+import lombok.AccessLevel;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class Person extends ApiResource implements MetadataStore<Person>, HasId {
+  @Getter(onMethod = @__({@Override})) String id;
+  String object;
+  String account;
+  Address address;
+  JapanAddress addressKana;
+  JapanAddress addressKanji;
+  Long created;
+  Boolean deleted;
+  DateOfBirth dob;
+  String email;
+  String firstName;
+  String firstNameKana;
+  String firstNameKanji;
+  String gender;
+  Boolean idNumberProvided;
+  String lastName;
+  String lastNameKana;
+  String lastNameKanji;
+  String maidenName;
+  @Getter(onMethod = @__({@Override})) Map<String, String> metadata;
+  String phone;
+  Requirements requirements;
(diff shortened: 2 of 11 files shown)


src/main/java/com/stripe/model/Person.java
```diff
@@ -0,0 +1,109 @@
...
+    Boolean owner;
+    BigDecimal percentOwnership;
+    Boolean representative;
+    String title;
+  }
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class Requirements extends StripeObject {
```
ob-stripe: You mentioned that this class might also be used directly on accounts in the future. Maybe we should not make it an inner class then, to avoid a breaking change later.

src/main/java/com/stripe/model/Account.java
```diff
@@ -144,6 +144,33 @@ public static AccountCollection list(Map<String, Object> params, RequestOptions
   }
   // </editor-fold>
 
+  // <editor-fold desc="persons">
+  /**
+   * Retrieve an account's persons.
```
ob-stripe: This is a list call, so maybe change this (and the other 2 below) to:
```suggestion
   * List an account's persons.
```

src/main/java/com/stripe/model/PersonCollection.java
```diff
@@ -0,0 +1,63 @@
...
+import com.stripe.exception.StripeException;
+import com.stripe.net.ApiResource;
+import com.stripe.net.RequestOptions;
+
+import java.util.Map;
+
+public class PersonCollection extends StripeCollection<Person> {
+  // <editor-fold desc="all">
+  @Deprecated
+  public PersonCollection all(Map<String, Object> params) throws StripeException {
```
ob-stripe: `all()` methods are deprecated. You've added `list()` methods below, no need to have `all()` methods too. We removed most of these in the big v6 overhaul.

src/test/java/com/stripe/functional/PersonTest.java
```diff
@@ -0,0 +1,120 @@
...
+    PersonCollection persons = account.persons();
+    resetNetworkSpy();
+
+    return persons;
+  }
+
+  @Test
+  public void testCreate() throws IOException, StripeException {
+    final Account account = Account.retrieve(ACCOUNT_ID, null);
+    PersonCollection personsTmp = getPersonCollectionFixture(account);
```
ob-stripe: You might as well move the call to `Account.retrieve()` inside `getPersonCollectionFixture()` and remove the argument.

ob-stripe: Also you can declare `personsTmp` as `final`.

remi-stripe:
> Also, making Verification and JapanAddress top-level classes unfortunately means this is a breaking change, but I don't see any way around it.

Can you clarify why it's a breaking change? Is the solution to duplicate them to avoid the breaking change? This will definitely happen more and more over time as we ship products with features that get re-used by others later.

src/test/java/com/stripe/functional/PersonTest.java
remi-stripe: > You might as well move the call to Account.retrieve() inside getPersonCollectionFixture() and remove the argument.

That won't work since I need the account id to verify the path. I could hardcode it but we've had issues before where the resource id in the path is not the one you asked for so it felt cleaner. Do you disagree?

> Also you can declare personsTmp as final.

ack

src/main/java/com/stripe/model/PersonCollection.java
remi-stripe: Did we miss some? If not why are they still here? I mostly copy/pasted another resource when I added this

src/main/java/com/stripe/model/Account.java
remi-stripe: Ack, though I did what you did in `Source.java` for source transactions

src/main/java/com/stripe/model/Person.java
remi-stripe: Ack. I feel like it was a real mistake to make those inner classes if moving them later is a breaking change. I think we should actively consider undoing that so that classes can just be re-used if they end up being shared.

ob-stripe:
> Can you clarify why it's a breaking change?

If someone has code like this:
```java
Account.Verification verification = account.getVerification();
```

then their code wouldn't compile anymore since `Account.Verification` is now just `Verification`.

> Is the solution to duplicate them to avoid the breaking change?

That would work, but I'm not thrilled about keeping dupe code around :/

remi-stripe:
@ob-stripe Fixed it by changing to inner classes again (And realized that I used Account.Verification instead of LegalEntity.Verification on Person which was wrong).

PTAL

src/main/java/com/stripe/model/PersonCollection.java
ob-stripe: `BitcoinReceiverCollection`, `BitcoinTransactionCollection` and `ExternalAccountCollection` still have `all()` methods. We should probably remove all of those in the next major version.

src/main/java/com/stripe/model/Person.java
ob-stripe: Yeah. We briefly discussed it in #547. @brandur-stripe said:
> There might be some complications around subresources that are shared like you mentioned with `Address` — occasionally these might appear as linked to only one class originally, but then be introduced in another class, so we'd have to promote them to the top level. I don't think it happens often enough to be a major problem though.

We might have been overly optimistic here :/

PR #615, opened 2018-11-14
Add support for `last_payment_error` for PaymentIntent

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @kelvin-stripe


ob-stripe:
LGTM.

It's not great that we're more or less duplicating the error object here, but I can't think of a better solution given the current state of the library.

PR #623, opened 2018-11-27
Add support for Review resource

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Review.java
+++ b/src/main/java/com/stripe/model/Review.java
@@ -1,5 +1,14 @@
 package com.stripe.model;
 
+import com.stripe.exception.StripeException;
+import com.stripe.model.ExpandableField;
+import com.stripe.model.HasId;
+import com.stripe.net.ApiResource;
+import com.stripe.net.RequestOptions;
+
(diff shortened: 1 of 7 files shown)


src/main/java/com/stripe/model/Review.java
```diff
@@ -1,18 +1,134 @@
...
+import lombok.AccessLevel;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 
 @Getter
 @Setter
 @EqualsAndHashCode(callSuper = false)
-public class Review extends StripeObject implements HasId {
+public class Review extends ApiResource implements HasId {
```
zanker-stripe: My impression from reading the code, is `StripeObject` is a basic object wrapper, while `ApiResource` is a basic object that can be extended to support other APIs like `approve` in this case. Is that the right way of looking at it?

ob-stripe: Basically, `StripeObject` == a k/v hash returned by Stripe's API and `ApiResource` == an actual resource that has API methods (including standard CRUDL methods).

PR #626, opened 2018-12-08
[WIP] Add support for the Account Link resource

remi-stripe:
This is marked as [WIP] because it won't be merged for now. We should wait for the API to be 100% stable before releasing it in .net.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ before_install:
 
 env:
   global:
-  - STRIPE_MOCK_VERSION=0.38.0
+  - STRIPE_MOCK_VERSION=0.39.0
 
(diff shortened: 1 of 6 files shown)


mickjermsurawong-stripe:
Verified with the that model has complete information!
Test request parameters has all the required values, except for optional `collect`
LGTM!

ob-stripe:
Thanks @mickjermsurawong-stripe! LGTM as well.

remi-stripe:
re-assigning to @ob-stripe. You already checked, but since every time I try to release a java version it fails, it seems safer to just have you re-review and release.

PR #628, opened 2018-12-27
Apply code improvements on tests

seratch:
Please forgive me for sending a bit large patch. The changes are quite simple but it may take long to review. This pull request includes the following changes:

* Remove unnecessary imports, thrown exceptions
* Use diamond operators as much as possible
* Fix a few typo/misspelling in naming

~In regards to organizing unnecessary imports, I relied on IntelliJ IDEA's feature. It was inevitable to avoid reordering imports and static imports. If you don't like those changes, I can come up with only manual changes (it's a bit hard for me, though..)~

Update: `./gradlew checkstyleTest` validates the above. I manually fixed unnecessary imports.

PR #630, opened 2018-12-27
Use diamond operators as much as possible

seratch:
This pull request modifies the code to use diamond operators more. Similar to #628


ob-stripe:
Nice, thanks again @seratch. I wish there were a Checkstyle rule we could add to enforce this going forwards, but it doesn't look like one exists.

brandur-stripe:
Woo! Yeah, that's way nicer.

PR #648, opened 2019-01-11
Add support for new properties and params on Invoice and Subscription

remi-stripe:
* Add `transfer_data[destination]` to both
* Add `application_fee_amount` to Invoice

cc @stripe/api-libraries 

NOT asking for review yet as I want to double check that we don't need `on_behalf_of` for now.

### DIFF
--- a/src/main/java/com/stripe/model/Invoice.java
+++ b/src/main/java/com/stripe/model/Invoice.java
@@ -22,7 +22,7 @@ public class Invoice extends ApiResource implements MetadataStore<Invoice>, HasI
   Long amountDue;
   Long amountPaid;
   Long amountRemaining;
-  Long applicationFee;
+  Long applicationFeeAmount;
   Long attemptCount;
   Boolean attempted;
   Boolean autoAdvance;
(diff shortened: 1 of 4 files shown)


remi-stripe:
This is now ready for review. It's not yet public but it's something we want to support in the libraries for a while before we ship it publicly.

r? @mickjermsurawong-stripe

src/main/java/com/stripe/model/Invoice.java
```diff
@@ -450,4 +459,8 @@ public Invoice voidInvoice(Map<String, Object> params, RequestOptions options)
     List<String> lineItemIds;
     Long usageGte;
   }
+
+  public static class TransferData extends StripeObject {
+    String destination;
```
mickjermsurawong-stripe: whoops i think we might need these magic to make the fields accessible.. 
for setter just for consistency. Same for the `Subscription#TransferData` too.

```
  @Getter
  @Setter
```

mickjermsurawong-stripe:
ptal @remi-stripe

remi-stripe:
@mickjermsurawong-stripe Nice catch! I went to add a test but since it's gated stripe-mock doesn't support it yet so I put a TODO for now. PTAL

mickjermsurawong-stripe:
LGTM

remi-stripe:
I'm going to go ahead and merge + release even though the approved label did not unblock.

PR #651, opened 2019-01-11
Introduce version pinning for autogen

mickjermsurawong-stripe:
- This should not be merged to master but to `mickjermsurawong/autogen-pre-master` branch, but i'm basing here on master to show compatibility and to pass master tests first.
- The only complication is in `EphemeralKey` that does not use the static version. (This class is together with File, Event, are not autogenerated).
  - Introduces `stripeVersionOnBehalfOf` for request options. This is different from both [.NET](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/EphemeralKeys/EphemeralKeyService.cs#L39-L46) and [Go](https://github.com/stripe/stripe-go/blob/master/ephemeralkey/client.go#L24-L26). where it takes version as the method params. I think that it would be better continue the pattern of header-related values to request option, and not in the parameters map.
r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -1,26 +1,35 @@
 package com.stripe.net;
 
 import com.stripe.Stripe;
+import java.util.Map;
 
 public class RequestOptions {
   public static RequestOptions getDefault() {
-    return new RequestOptions(Stripe.apiKey, Stripe.clientId, Stripe.apiVersion, null, null,
+    return new RequestOptions(Stripe.apiKey, Stripe.clientId, null, null, null,
         Stripe.getConnectTimeout(), Stripe.getReadTimeout());
   }
 
   private final String apiKey;
   private final String clientId;
-  private final String stripeVersion;
+  /**
+   * Stripe version always set at {@link Stripe#API_VERSION}.
+   */
+  private final String stripeVersion = Stripe.API_VERSION;
+  /**
+   * Stripe version when made on behalf of others. This can be used when the returned response
+   * will not be deserialized into the current classes pinned to {@link Stripe#VERSION}.
+   */
+  private final String stripeVersionOnBehalfOf;
   private final String idempotencyKey;
   private final String stripeAccount;
   private final int connectTimeout;
   private final int readTimeout;
 
-  private RequestOptions(String apiKey, String clientId, String stripeVersion,
+  private RequestOptions(String apiKey, String clientId, String stripeVersionOnBehalfOf,
       String idempotencyKey, String stripeAccount, int connectTimeout, int readTimeout) {
     this.apiKey = apiKey;
     this.clientId = clientId;
-    this.stripeVersion = stripeVersion;
+    this.stripeVersionOnBehalfOf = stripeVersionOnBehalfOf;
     this.idempotencyKey = idempotencyKey;
     this.stripeAccount = stripeAccount;
     this.connectTimeout = connectTimeout;
@@ -33,6 +42,10 @@ public class RequestOptions {
 
   public String getClientId() {
     return clientId;
+  }
+
+  public String getStripeVersionOnBehalfOf() {
+    return stripeVersionOnBehalfOf;
   }
 
   public String getStripeVersion() {
@@ -93,9 +106,12 @@ public class RequestOptions {
     int result = apiKey != null ? apiKey.hashCode() : 0;
     result = 31 * result + (clientId != null ? clientId.hashCode() : 0);
     result = 31 * result + (stripeVersion != null ? stripeVersion.hashCode() : 0);
+    result = 31 * result + (stripeVersionOnBehalfOf != null ? stripeVersionOnBehalfOf.hashCode()
+        : 0);
     result = 31 * result + (idempotencyKey != null ? idempotencyKey.hashCode() : 0);
-    result = 31 * result + readTimeout;
+    result = 31 * result + (stripeAccount != null ? stripeAccount.hashCode() : 0);
     result = 31 * result + connectTimeout;
+    result = 31 * result + readTimeout;
     return result;
   }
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -1,26 +1,35 @@
...
-  private final String stripeVersion;
+  /**
+   * Stripe version always set at {@link Stripe#API_VERSION}.
+   */
+  private final String stripeVersion = Stripe.API_VERSION;
+  /**
+   * Stripe version when made on behalf of others. This can be used when the returned response
+   * will not be deserialized into the current classes pinned to {@link Stripe#VERSION}.
+   */
+  private final String stripeVersionOnBehalfOf;
```
mickjermsurawong-stripe: Generalize the stripe version of mobile client required by ephemeral key to `stripeVersionOnBehalfOf`.
Putting in request option here avoid considering stripe version here has method params.

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -93,9 +106,12 @@ public int hashCode() {
     int result = apiKey != null ? apiKey.hashCode() : 0;
     result = 31 * result + (clientId != null ? clientId.hashCode() : 0);
     result = 31 * result + (stripeVersion != null ? stripeVersion.hashCode() : 0);
+    result = 31 * result + (stripeVersionOnBehalfOf != null ? stripeVersionOnBehalfOf.hashCode()
+        : 0);
     result = 31 * result + (idempotencyKey != null ? idempotencyKey.hashCode() : 0);
-    result = 31 * result + readTimeout;
+    result = 31 * result + (stripeAccount != null ? stripeAccount.hashCode() : 0);
```
mickjermsurawong-stripe: `stripeAccount` is added, although not related to this PR. I think it was unintentionally left out

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -155,13 +174,19 @@ public RequestOptionsBuilder clearClientId() {
...
+     * to explicit Stripe resources -- it intends to make request on behalf of others, makes
+     * request with their API version, and simply passes the raw values along. Example for this is
+     * {@link com.stripe.model.EphemeralKey#create(Map, RequestOptions)}.
+     * Setting this value in typical scenario will result in deserialization error
+     * as the model classes has schema according to the pinned {@link Stripe#API_VERSION} and not
+     * the {@code stripeVersionOnBehalfOf}
+     * @param stripeVersionOnBehalfOf stripe version of the client to make request on behalf of
+     * @return option builder
+     */
+    public RequestOptionsBuilder setStripeVersionOnBehalfOf(String stripeVersionOnBehalfOf) {
```
brandur-stripe: Not too sure about the naming here — can we just call it `setStripeVersion`? We can specify in the comments that it'll override the built-in version.

mickjermsurawong-stripe: @brandur, I'm wondering if i should make the name different here. I'm afraid that merchants who are using pre-pinning version still have `setStripeVersion` invocations, and once they migrate to pinned-version will not realize the semantic difference here. 
Making it explicitly different here, and remove the old method will at least ensure that they set up the integration correctly.

Would naming like `overrideStripeVersion` would be a better alternative?

brandur-stripe: Oh oops — I didn't read carefully enough and didn't realize that `setStripeVersion` could already be called for `RequestOptions` before.

Yeah, I was actually going to suggest using "override" in the name as an alternative before, so totally okay with that. I'd suggest `stripeVersionOverride` instead of `overrideStripeVersion` though since it sorts more nicely.

mickjermsurawong-stripe: acked!

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -72,12 +87,21 @@ public boolean equals(Object o) {
...
+    if (idempotencyKey != null ? !idempotencyKey.equals(that.idempotencyKey) :
+        that.idempotencyKey != null) {
+      return false;
+    }
+    if (stripeAccount != null ? !stripeAccount.equals(that.stripeAccount) :
+        that.stripeAccount != null) {
       return false;
     }
-    if (stripeVersion != null ? !stripeVersion.equals(that.stripeVersion)
-        : that.stripeVersion != null) {
```
mickjermsurawong-stripe: Reordering to match the alphabetically sorted instance variables, plus adding `stripeAccount`

mickjermsurawong-stripe:
thank you @brandur! I cleaned up variable alphabetical order, and renamed the the `stripeOnBehalfOf` to `stripeVersionOverride`
ptal @ob-stripe

src/main/java/com/stripe/net/RequestOptions.java
ob-stripe: We could probably replace these manual implementations with Lombok's `@EqualsAndHashCode`.

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -92,28 +116,35 @@ public boolean equals(Object o) {
...
 
+  /**
+   * Convert request options to builder, retaining invariant values for the integration.
+   * @return option builder.
+   */
   public RequestOptionsBuilder toBuilder() {
-    return new RequestOptionsBuilder().setApiKey(this.apiKey).setStripeVersion(this.stripeVersion)
-        .setStripeAccount(this.stripeAccount);
+    return new RequestOptionsBuilder()
+        .setApiKey(this.apiKey).setStripeAccount(this.stripeAccount);
```
ob-stripe: Shouldn't you call `.setStripeVersionOverride()` here?

ob-stripe:
LGTM!

mickjermsurawong-stripe:
thank you both! will rebase this on my pre-autogen branch

mickjermsurawong-stripe:
self-approving after rebase on autogen branch
LGTM

src/main/java/com/stripe/net/RequestOptions.java
mickjermsurawong-stripe: whoops, sorry i missed this and merged to my branch without it. Will update in subsequent PR

PR #654, opened 2019-01-16
Best-attempt event read on version pinning

mickjermsurawong-stripe:
- This PR again should not be merged to master but to pre-autogen branch, similar to https://github.com/stripe/stripe-java/pull/651 (because pre-autogen branch without autogenerated code on top of it doesn't compile)
- This PR aims to make interface on reading `EventData` more explicit, having 3 the read mode: safe, best attempt, and raw read on failure.
- The proposed abstraction here has non-breaking interface `getData()` to existing integration, and has the same behavior except on failure case.
- This implementation here now still uses non-pinned version, but the intention here is to release with pinned version; sorry for the inconsistency here.
r? @ob-stripe @brandur-stripe 
cc @stripe/api-libraries 

The build fail now is due to jdk 11 failing to download.

### DIFF
--- a/src/main/java/com/stripe/model/Event.java
+++ b/src/main/java/com/stripe/model/Event.java
@@ -1,5 +1,8 @@
 package com.stripe.model;
 
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.stripe.Stripe;
 import com.stripe.exception.StripeException;
 import com.stripe.net.ApiResource;
 import com.stripe.net.RequestOptions;
@@ -19,7 +22,7 @@ public class Event extends ApiResource implements HasId {
   String account;
   String apiVersion;
   Long created;
-  EventData data;
+  TolerantRead<EventData> data;
   Boolean livemode;
   Long pendingWebhooks;
   EventRequest request;
@@ -69,6 +72,90 @@ public class Event extends ApiResource implements HasId {
   public static Event retrieve(String id, Map<String, Object> params, RequestOptions options)
       throws StripeException {
     return request(RequestMethod.GET, instanceUrl(Event.class, id), params, Event.class, options);
+  }
+
+  /**
+   * Deprecated in favor of more specific mode of reading data based on API version configured,
+   * and the failure during deserialization. When deserialization fails, result is null, and
+   * failure information should be checked. See {@link Event#rawReadDataOnFailure()} and
+   * {@link Event#getReadDataException()}.
+   * @return nullable event data.
+   */
+  @Deprecated
+  public EventData getData() {
+    if (data == null) {
+      return null;
+    }
+    if (canSafeReadData()) {
+      return data.getSafeData();
+    } else {
+      return data.getBestAttemptData();
+    }
+  }
+
+  /**
+   * Checks whether API version of the event matches that {@link Stripe#apiVersion} of this
+   * integration.
+   * @return true when the versions match.
+   */
+  public boolean canSafeReadData() {
+    return apiVersion != null && apiVersion.equals(Stripe.apiVersion);
+  }
+
+  /**
+   * When non-null, the deserialized data preserves high integrity because
+   * of strong correspondence between schema of Stripe response and the Java model class.
+   * @return Safe event data with high fidelity.
+   */
+  public EventData safeReadData() {
+    if (data == null) {
+      return null;
+    }
+    return data.getSafeData();
+  }
+
+  /**
+   * When non-null, the deserialized data is not guaranteed to to contain all the fields, because
+   * there can be mismatch between schema of Stripe response and the Java model class.
+   *
+   * @return Best attempt deserialized data.
+   */
+  public EventData bestAttemptReadData() {
+    if (data == null) {
+      return null;
+    }
+    return data.getBestAttemptData();
+  }
+
+  /**
+   * When non-null, gets raw JSON data that couldn't be successfully deserialized.
+   * The returned response can be mutated and will not affect the original data. This allows for
+   * evolving your JSON object to be compatible with that of current model class.
+   * @return raw JSON event data
+   */
+  public JsonObject rawReadDataOnFailure() {
+    if (data == null) {
+      return null;
+    }
+    return data.getRawDataOnFailure().deepCopy();
+  }
+
+  /**
+   * Flag whether read fails due to deserialization failure.
+   */
+  public boolean isReadDataFailure() {
+    return data != null && data.getRawDataOnFailure() != null && data.getReadException() != null;
+  }
+
+  /**
+   * Parsing exception during data deserialization. Other exceptions are not specifically handled.
+   * @return parsing exception.
+   */
+  public JsonParseException getReadDataException() {
+    if (data == null) {
+      return null;
+    }
+    return data.getReadException();
   }
   // </editor-fold>
 }
--- a/src/main/java/com/stripe/model/EventTypeAdapterFactory.java
+++ b/src/main/java/com/stripe/model/EventTypeAdapterFactory.java
@@ -1,0 +1,113 @@
+package com.stripe.model;
+
+import com.google.gson.ExclusionStrategy;
+import com.google.gson.FieldAttributes;
+import com.google.gson.FieldNamingPolicy;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+import java.lang.reflect.Type;
+
+public class EventTypeAdapterFactory implements TypeAdapterFactory {
+
+  private static final Gson GSON_EVENT = new GsonBuilder()
+      .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+      // excludes `data` field during automatic deserialization because we have custom handling
+      .setExclusionStrategies(new ExcludeEventDataFieldStrategy())
+      .create();
+
+  private static class ExcludeEventDataFieldStrategy implements ExclusionStrategy {
+    public boolean shouldSkipClass(Class<?> arg0) {
+      return false;
+    }
+
+    public boolean shouldSkipField(FieldAttributes f) {
+      return (f.getDeclaringClass() == Event.class && f.getName().equals("data"));
+    }
+  }
+
+  public static class EventSerializer implements JsonSerializer<Event> {
+    @Override
+    public JsonElement serialize(Event event, Type typeOfSrc, JsonSerializationContext context) {
+
+      // get event data from different read modes
+      JsonElement rawEventData;
+      if (event.isReadDataFailure()) {
+        rawEventData = event.rawReadDataOnFailure();
+      } else if (event.canSafeReadData()) {
+        rawEventData = context.serialize(event.safeReadData());
+      } else {
+        rawEventData = context.serialize(event.bestAttemptReadData());
+      }
+
+      // serialized content here has no `data` field due to the exclusion strategy
+      JsonObject serialized =  GSON_EVENT.toJsonTree(event, typeOfSrc).getAsJsonObject();
+      serialized.add("data", rawEventData);
+      return serialized;
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+    if (!Event.class.isAssignableFrom(type.getRawType())) {
+      return null;
+    }
+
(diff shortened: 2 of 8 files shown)


brandur-stripe:
Nice @mickjermsurawong-stripe! Thanks for taking a stab at this one.

I think this looks like a super plausible approach. The one thing I'd consider changing though is just the public interface that we expose — to my eye, the one in here presents the user with too many options to leaf through.

If we boil everything right down, I think what we want to do is give the user two paths for accessing event data:

1. A "safe" option that returns either data or nothing if decoding failed. 
2. An unsafe option that returns easy access to data or throws an exception if decoding fails.

The way I'd love to see the interface work is something like this:

``` java
maybeData = event.getDataMaybe();
if (maybeData.deserialize()) {
    maybeData.getData().doSomething();
} else {
    maybeData.getRawJson().doSomething();
}
```

Or for the unsafe variant:

``` java
try {
    data = event.getDataMaybe().unwrap();
} catch (EventDataDeserializationException e) {
    e.getRawJson().doSomething();
}
```

I don't think it's too dissimilar to what you have, but IMO the current interface might be shoehorning too many methods onto `Event` when creating a proxy like in the above might allow us to expose a cleaner interface. Also, I think that the "helper" methods like `canSafeReadData` and `isReadDataFailure` should not be exposed to the user to help streamline the interface a bit. Thoughts?

mickjermsurawong-stripe:
Thank you @brandur-stripe! 
I like your approach of making explicit the notion deserialization using the interface that asks users on how to deserialize it instead of mine trying to conceal with vague notions like "TolerantRead", or "safe/bestAttemptRead" and to deserialize `EventData` on behalf of user, and to maintain/present the state of that work to the user later.

src/main/java/com/stripe/exception/EventDataDeserializationException.java
```diff
@@ -0,0 +1,18 @@
+package com.stripe.exception;
+
+import com.google.gson.JsonElement;
+import lombok.Getter;
+
+public class EventDataDeserializationException extends StripeException {
```
mickjermsurawong-stripe: New exception class to be handled when forcing deserialization

src/main/java/com/stripe/model/EventTypeAdapterFactory.java
```diff
@@ -0,0 +1,104 @@
...
+            // remove data from root, and we handle deserialization separately
+            rootEventJsonObject.remove("data");
+
+            Event event = eventTypeAdapter.fromJsonTree(rootEventJsonObject);
+
+            if (eventDataJsonObject != null) {
+              if (event.getApiVersion() != null) {
+                // motivation for deserializing event data from its root event object:
+                // we want the event data to know which api version it is tagged with.
+                eventDataJsonObject.add("api_version", new JsonPrimitive(event.getApiVersion()));
```
mickjermsurawong-stripe: How `EventData` now has the information about its version from the root object `Event`.

src/main/java/com/stripe/model/EventData.java
```diff
@@ -10,6 +15,137 @@
...
+  // Original properties
   StripeObject object;
   Map<String, Object> previousAttributes;
+
+  // Augmented properties during deserialization
+  @Setter(AccessLevel.NONE)
+  JsonElement rawJson;
+  @Getter(AccessLevel.NONE)
+  @Setter(AccessLevel.NONE)
+  String apiVersion;
```
mickjermsurawong-stripe: When event data stands by itself, not tied to the root-level object, it should have information about its tagged api version to be useful for guiding the "safe" deserialization.
An alternative is to make `EventData` non-static inner class of `Event`, but it will have reference to all its parent attribute.

src/main/java/com/stripe/model/EventData.java
```diff
@@ -10,6 +15,137 @@
...
+   * user instead. See {@link EventData#safeDeserialize()} and
+   * {@link EventData#forceDeserialize()}.
+   * @param rawJson             raw JSON for {@code object} field that can be deserialized into
+   *     StripeObject.
+   * @param previousAttributes  previous attributes of event data (has no further
+   *     deserialization concerns)
+   * @param apiVersion          api version of the event data, used to determine whether the
+   *     deserialization to {@code object} field is safe.
+   */
+  EventData(JsonElement rawJson,
```
mickjermsurawong-stripe: default access here, so it's not available to the user, only available within our package and specifically for deserialization

src/main/java/com/stripe/model/EventData.java
```diff
@@ -10,6 +15,137 @@
...
+
+  /**
+   * Safe deserialize raw JSON into StripeObject. This operation mutates the state, and the
+   * successful result can be accessed via {@link EventData#getObject()}. Matching API
+   * version between the current integration and the event is necessary to guarantee safe
+   * deserialization.
+   * @return whether deserialization has been successful.
+   */
+  public boolean safeDeserialize() {
+    if (!apiVersionMatch()) {
```
mickjermsurawong-stripe: it will also return nothing when there's api version mismatch between the "pinned" version and the event version

src/main/java/com/stripe/model/EventDataDeserializer.java
```diff
@@ -69,29 +67,29 @@ private void populateMapFromJsonObject(Map<String, Object> objMap, JsonObject js
...
-            entry.getValue(), cl != null ? cl : StripeRawJsonObject.class);
-        eventData.setObject(object);
+        // deserialization to `StripeObject` is done in the event data itself.
+        rawJson = entry.getValue();
+      } else if ("api_version".equals(key)) {
+        apiVersion =  entry.getValue().getAsString();
       }
     }
-    return eventData;
+    return new EventData(rawJson, previousAttributes, apiVersion);
```
mickjermsurawong-stripe: Constructor instead of setter. Builder pattern is over-kill for now.

src/test/java/com/stripe/model/EventTest.java
```diff
@@ -39,5 +61,40 @@ public void testReserialize() throws Exception {
...
+
+    // comparing deserialized stripe object
+    StripeObject reserializedStripeObject = stubIntegrationApiVersion(
+        reserializedEventData, "2018-09-24").getObject();
+    StripeObject stripeObject = stubIntegrationApiVersion(
+        eventData, "2018-09-24").getObject();
+    assertEquals(reserializedStripeObject, stripeObject);
+  }
+
+  private EventData stubIntegrationApiVersion(EventData data,
```
mickjermsurawong-stripe: I didn't use the pattern of saving old Stripe version before the test and restore afterwards because they can be flakey when running parallel tests. Making a stub here instead.

mickjermsurawong-stripe:
ptal @brandur-stripe, the revised interface now will work like this. I try to make the naming for the two options a little consistent.
```
   StripeObject stripeObject;
    if (eventData.safeDeserialize()) {
      stripeObject = eventData.getObject();
    } else {
      eventData.getObject(); // gives null
      try {
        stripeObject = eventData.forceDeserialize();
      } catch (EventDataDeserializationException e) {
        JsonElement rawJson = e.getRawJson();
        // should we offer functionality below?
        // JsonElement patchedRawJson = patch(rawJson);
        // stripeObject = eventData.deserializeStripeObject(patchedRawJson); 
      }
    }
```
- What i'm not sure is if we should expose a way to deserialized the patched raw JSON in the failure mode. 
- I also didn't create a new interface for this new mode -- it is reusing the same `.getData()` and `getObject`. But that can be unexpected to user suddenly getting `getObject` returning null on version mismatch. I'm thinking of going with this, so making all the changes in this PR to the new class, and retain the old `EventData` object.
```
public class Event extends ApiResource implements HasId {
  ... 
  String apiVersion;
  Long created;
  @Deprecated
  EventData data;
  EventVersionedData versionedData;
  ...
```

src/main/java/com/stripe/model/Event.java
```diff
@@ -31,6 +31,14 @@
   @Deprecated
   String userId;
 
+  /**
+   * Get event data with explicit deserialization support. This object contains superset
+   * information of {@link EventData}.
+   */
+  public EventVersionedData getVersionedData() {
```
mickjermsurawong-stripe: Entry point to the new proxy object that is preferred over the original `EventData`

src/main/java/com/stripe/model/EventData.java
```diff
@@ -10,6 +10,22 @@
 @Setter
 @EqualsAndHashCode(callSuper = false)
 public class EventData extends StripeObject {
-  StripeObject object;
+  /**
+   * Raw JSON object intended to be deserialized as {@code StripeObject}. The deserialization
+   * should be deferred to the user. See the now deprecated method {@link EventData#getObject()}.
+   */
+  JsonObject object;
```
mickjermsurawong-stripe: Having JSON makes sure that class-specific deserialization failure is deferred (not during deserializing the outer class `Event`) and can be handled in the new API `EventVersionedData`.

Although there is type change here, but the exposed API to client getting StripeObject is still the same via `getObject` deprecated method below.

src/main/java/com/stripe/model/EventDataDeserializer.java
```diff
@@ -85,13 +85,24 @@ public EventData deserialize(JsonElement json, Type typeOfT, JsonDeserialization
           eventData.setPreviousAttributes(previousAttributes);
         }
       } else if ("object".equals(key)) {
-        String type = element.getAsJsonObject().get("object").getAsString();
-        Class<? extends StripeObject> cl = EventDataClassLookup.findClass(type);
-        StripeObject object = ApiResource.GSON.fromJson(
-            entry.getValue(), cl != null ? cl : StripeRawJsonObject.class);
-        eventData.setObject(object);
+        eventData.setObject(entry.getValue().getAsJsonObject());
```
mickjermsurawong-stripe: `EventData` no longer tries to deserialize risky class-specific JSON, and it is exposed in `EventVersionedData` instead making using of the method below `deserializeStripeObject`

src/main/java/com/stripe/model/EventVersionedData.java
```diff
@@ -0,0 +1,182 @@
...
+ * } else {
+ *   try {
+ *     stripeObject = versionedData.forceDeserialize();
+ *   } catch (EventDataDeserializationException e) {
+ *     JsonElement compatibleJson = compatibilityMapper(
+ *       e.getRawJsonObject(),
+ *       event.getApiVersion());
+ *     stripeObject = versionedData.retryDeserialize(compatibleJson);
+ *   }
+ * }
```
mickjermsurawong-stripe: The flow of how client may use this new object.

brandur-stripe:
Nice @mickjermsurawong-stripe!

LGTM. The only thing I'd argue for are some name changes:

* `getVersionedData`: IMO, this should start with `getData*` just so that it sorts nicely with the existing method `getData` (which makes it more discoverable). `getDataVersioned` is a little awkward, but maybe something like `getDataDeserializer`, `getDataWrapper`, or `getDataProxy`?
* `EventVersionedData`: Keep this in line if anything changes above. I'm not sure that "versioned" is important enough to be in the name although I definitely see where you're coming from. (I'd probably go with `EventDataDeserializer` because this feels more like a deserializer than a data object.)
* `safeDeserialize` + `forceDeserialize`: As above, I think these should start with the same suffix so that they lie next to each other when someone's looking at them in an autocomplete dialog. I'd probably go with `deserialize` and `deserialzeUnsafe`, or something along those lines.

brandur-stripe:
Woot. Thanks @mickjermsurawong-stripe!

@ob-stripe Mind taking a look at this one as well? You probably have more informed ideas than I do.

mickjermsurawong-stripe:
Thank you @ob-stripe for the discussion on Slack.
- I've addressed your concern of exposing GSON-based JSON for user to transform.
Returns the raw JSON as string for more custom usage.
For allowing user to transform JSON and retry deserialization, I want to get the benefit of having GSON `JsonObject` for easy manipulation but also want to hide it away from user for general usage.
Here, I implemented `CompatibilityTransformer` essentially a "lambda" that provides the raw JSON and some metadata for user to decide how to transform the JSON.
I know this is essentially pushing users to do their own forward compatibility, but I think it's the fullest support we can provide when we move to pinned-version, given the current state of our backed.
```
  Event event = Event.retrieve("evt_123");
  EventDataObjectDeserializer deserializer = event.getDataObjectDeserializer();
  StripeObject stripeObject;
  if (deserializer.deserialize()) {
    stripeObject = deserializer.getObject();
  } else {
    try {
      stripeObject = deserializer.deserializeUnsafe();
    } catch (EventDataDeserializationException e) {
      stripeObject = deserializer.deserializeUnsafeWith(
          new EventDataObjectDeserializer.CompatibilityTransformer() {
            @Override
            public JsonObject transform(JsonObject rawJsonObject,
                                        String apiVersion,
                                        String eventType) {
              if (apiVersion.equals("2018-05-01")) {
              	String fooVal = rawJsonObject.remove("foo");
              	rawJsonObject.add("foo_v2", fooVal);
              }
              if (apiVersion.equals("2018-08-01")) {
              	anotherTransform(rawJsonObject);
              }
              return rawJsonObject;
            }
          });
      );
    }
  }
```
ptal @ob-stripe

mickjermsurawong-stripe:
Addressed the pending concern, and from discussion on slack @ob-stripe agrees with this as a temporary solution for event endpoint integration.
From the team discussion on Monday, this is a valid approach for now. Self-approving here
LGTM

mickjermsurawong-stripe:
Rebasing to `autogen/8.0`

PR #653, opened 2019-01-15
Add support for `custom_fields` and `footer` on `Invoice`

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @mickjermsurawong-stripe 

https://stripe.com/docs/api/invoices/create#create_invoice-custom_fields
https://stripe.com/docs/api/invoices/create#create_invoice-footer

PR #655, opened 2019-01-17
Add support for billing thresholds

remi-stripe:
r? @mickjermsurawong-stripe @ob-stripe 
cc @stripe/api-libraries 

cc @clintonb-stripe Could you double check that I added the right set of properties too and I'm not missing any? (Java only lists properties, not parameters).

Docs: https://stripe.com/docs/billing/subscriptions/metered-billing/thresholds
Internal PR with gate removal: #126799

### DIFF
--- a/src/main/java/com/stripe/model/Invoice.java
+++ b/src/main/java/com/stripe/model/Invoice.java
@@ -62,6 +62,7 @@ public class Invoice extends ApiResource implements MetadataStore<Invoice>, HasI
   Long subtotal;
   Long tax;
   BigDecimal taxPercent;
+  ThresholdReason thresholdReason;
   Long total;
   Long webhooksDeliveredAt;
 
@@ -414,5 +415,21 @@ public class Invoice extends ApiResource implements MetadataStore<Invoice>, HasI
   public static class CustomField extends StripeObject {
     String name;
     String value;
+  }
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class ThresholdReason extends StripeObject {
+    Long amountGte;
+    List<ThresholdItemReason> itemReasons;
+  }
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class ThresholdItemReason extends StripeObject {
+    List<String> lineItemIds;
+    Long usageGte;
   }
 }
--- a/src/main/java/com/stripe/model/Subscription.java
+++ b/src/main/java/com/stripe/model/Subscription.java
@@ -21,6 +21,7 @@ public class Subscription extends ApiResource implements MetadataStore<Subscript
   BigDecimal applicationFeePercent;
   String billing;
   Long billingCycleAnchor;
+  BillingThresholds billingThresholds;
   Boolean cancelAtPeriodEnd;
   Long canceledAt;
   Long created;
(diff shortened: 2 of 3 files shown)


remi-stripe:
leaving approval to mick/ob since it's our process. Thanks a lot for the quick review @clintonb-stripe! Other two PRs incoming in a bit

src/main/java/com/stripe/model/Invoice.java
```diff
@@ -59,6 +60,7 @@
   Long subtotal;
   Long tax;
   BigDecimal taxPercent;
+  String thresholdReason;
```
mickjermsurawong-stripe: Should this be typed as `ThresholdReason`?

src/main/java/com/stripe/model/Invoice.java
```diff
@@ -404,4 +406,20 @@ public Invoice voidInvoice(Map<String, Object> params, RequestOptions options)
...
+  public static class ThresholdReason extends StripeObject {
+    Long amountGte;
+    List<ItemReason> itemReasons;
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class ItemReason extends StripeObject {
+      List<String> lineItemIds;
+      Long usageGte;
```
mickjermsurawong-stripe: Should the innermost class, when referred outside this class (in normal usage) will be
`Invoice.ThresholdItemReason` instead of the 3-level nesting as `Invoice.ThresholdReason.ItemReason` here? I think the former reads better.
So we can implement the following:
```
public static class ThresholdReason extends StripeObject {
  List<ThresholdItemReason> itemReasons
}
// the inner-most class is now at nested in Invoice instead
public static class ThresholdItemReason extends StripeObject {
}
```
Actually, in the autogen SDK, we currently enforce nesting at most two levels to flatten classes. It help to flatten long resource dependency we have in pay-server annotation too.
The `Charge.Level3.LineItem` will be in autogen (hypothetically) as `Charge.Level3LineItem` (although we decided on unwinding it to top level resource)
```
public class Charge {
  public static class Level3 extends StripeObject {
    List<Level3LineItem> lineItems;
    ....
  }
  public static class Level3LineItem extends StripeObject {
     ....
  }
}
```

src/main/java/com/stripe/model/Invoice.java
remi-stripe: @mickjermsurawong-stripe I don't think so no. We don't do "list of constants" in java AFAIK. @ob-stripe do you agree?

ob-stripe: That's not what Mick meant. The `threshold_reason` attribute is an object, not a string.

remi-stripe: ohhh I misunderstood. That `reason` naming is throwing me off. Nice catch mick, fixing!

src/main/java/com/stripe/model/Invoice.java
remi-stripe: Fixed!

remi-stripe:
@mickjermsurawong-stripe Fixed both! Can you take another look?

mickjermsurawong-stripe:
yay! thank you @remi-stripe  and @ob-stripe!
LGTM

remi-stripe:
Had to rebase as it conflicted with https://github.com/stripe/stripe-java/pull/653

remi-stripe:
tests passed, re-approving

PR #657, opened 2019-01-18
Better handling of namespaced resources

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Better handling of namespaced resources, by using the sub-package's name as the namespace (e.g. if the sub-package is `com.stripe.model.issuing`, `issuing` is the resource's namespace).

This way we don't have to manually add new namespaces.

This is still a bit too implicit / magic for my taste and I'd rather we explicitly declare the full Stripe object name in each model class, but we can do that later.

### DIFF
--- a/src/main/java/com/stripe/net/ApiResource.java
+++ b/src/main/java/com/stripe/net/ApiResource.java
@@ -86,18 +86,13 @@ public abstract class ApiResource extends StripeObject {
         .replaceAll("([a-z0-9])([A-Z])", "$1_$2")
         .toLowerCase();
 
-    // Issuing or Sigma resources are in their own package. Until we can support adding OBJECT_NAME
-    // to all classes, we use this dirty trick to properly format the API endpoints
-    if (clazz.getName().contains("com.stripe.model.issuing.")) {
-      className = "issuing/" + className;
-    } else if (clazz.getName().contains("com.stripe.model.sigma.")) {
-      className = "sigma/" + className;
-    } else if (clazz.getName().contains("com.stripe.model.radar.")) {
-      className = "radar/" + className;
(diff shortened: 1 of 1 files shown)


remi-stripe:
LGTM though I wonder if that will break at some point to rely on that 4 levels. Again I prefer to not approve core changes so re-assigning to @mickjermsurawong-stripe

src/main/java/com/stripe/net/ApiResource.java
```diff
@@ -86,18 +86,12 @@ private static String className(Class<?> clazz) {
...
-    } else if (clazz.getName().contains("com.stripe.model.reporting.")) {
-      className = "reporting/" + className;
-    } else if (clazz.getName().contains("com.stripe.model.terminal.")) {
-      className = "terminal/" + className;
+    // Handle namespaced resources by checking if the class is in a sub-package, and if so prepend
+    // it to the class name
+    String[] parts = clazz.getPackage().getName().split("\\.");
+    if (parts.length == 4) {
+      // The first three parts are always "com.stripe.model", the fourth part is the sub-package
+      className = parts[3] + "/" + className;
```
mickjermsurawong-stripe: nit: to remi's point, maybe we can consider just assert that we are not handling when length > 4

mickjermsurawong-stripe:
Left a nit comment, but I don't feel too strongly about; I doubt we will have subpackage within current package.
LGTM
thank you @ob-stripe!

src/main/java/com/stripe/net/ApiResource.java
ob-stripe: Good idea. I added an assertion that the package's name has either 3 or 4 parts.

ob-stripe:
Released as 7.18.1.

PR #659, opened 2019-01-21
Detect custom OpenAPI files and start stripe-mock from test suite

mickjermsurawong-stripe:
Check for custom open api files to run stripe-mock process from test suite.
This is the same work already implemented in other libs
https://github.com/stripe/stripe-dotnet/pull/1445
https://github.com/stripe/stripe-ruby/pull/715

I've verified that with the custom files detected, the build would pass in travis [here](https://github.com/stripe/stripe-java/compare/mickjermsurawong/stripe-mock-fixture...mickjermsurawong/stripe-mock-fixture-with-specs)

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/test/java/com/stripe/BaseStripeTest.java
+++ b/src/test/java/com/stripe/BaseStripeTest.java
@@ -68,7 +68,7 @@ public class BaseStripeTest {
   @BeforeClass
   public static void checkStripeMock() throws Exception {
     if (StripeMockProcess.start()) {
-      port = String.valueOf(StripeMockProcess.getPort());
+      port = StripeMockProcess.getPort();
       return;
     }
 
@@ -105,7 +105,7 @@ public class BaseStripeTest {
 
   /**
    * Activates usage of stripe-mock by overriding the API host and putting a test key
-   * into the environment. This is required independent of how strip-mock is started.
+   * into the environment. This is required independent of how stripe-mock is started.
    */
   @Before
   public void setUpStripeMockUsage() {
--- a/src/test/java/com/stripe/BaseStripeTest.java
+++ b/src/test/java/com/stripe/BaseStripeTest.java
@@ -46,11 +46,32 @@ public class BaseStripeTest {
   private String origClientId;
   private String origUploadBase;
 
+  static {
+    // To only stop stripe-mock process after all the test classes.
+    // Alternative solution using @AfterClass will stop the stripe-mock after
+    // every test class.
+    Runtime.getRuntime().addShutdownHook(new Thread(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          StripeMockProcess.stop();
+        } catch (Exception e) {
+          // ignore any exception from trying to stop embedded stripe-mock
+        }
+      }
+    }));
+  }
+
   /**
    * Checks that stripe-mock is running and up-to-date.
    */
   @BeforeClass
   public static void checkStripeMock() throws Exception {
+    if (StripeMockProcess.start()) {
+      port = String.valueOf(StripeMockProcess.getPort());
+      return;
+    }
+
     port = System.getenv().get("STRIPE_MOCK_PORT");
     if (port == null) {
       port = "12111";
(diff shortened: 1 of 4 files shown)


src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -83,11 +104,11 @@ public static void checkStripeMock() throws Exception {
   }
 
   /**
-   * Activates stripe-mock by overriding the API host and putting a test key
-   * into the environment.
+   * Activates usage of stripe-mock by overriding the API host and putting a test key
+   * into the environment. This is required independent of how strip-mock is started.
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: fixed!

src/test/java/com/stripe/StripeMockProcess.java
```diff
@@ -0,0 +1,148 @@
...
+            String.valueOf(port),
+            "-spec",
+            getPathSpec(),
+            "-fixtures",
+            getPathFixture()
+        );
+
+    // stripe-mock output/error will be the same as this current test parent process
+    processBuilder.redirectOutput(ProcessBuilder.Redirect.INHERIT);
+    processBuilder.redirectError(ProcessBuilder.Redirect.INHERIT);
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: Reverted to default, set to `Redirect.Pipe`
https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.Redirect.html#PIPE

src/test/java/com/stripe/StripeMockProcess.java
```diff
@@ -0,0 +1,153 @@
...
+      ));
+      System.exit(1);
+    }
+
+    pid = getProcessId(process);
+
+    if (isAlive(process)) {
+      System.out.println(String.format("Started stripe-mock, PID = %s", pid));
+      System.out.println("Finding port bound to stripe-mock...");
+      port = detectBoundPort(process);
```
mickjermsurawong-stripe: This would block if there's no process output. 
I could potentially make this non-blocking as a thread, and continually ping within the timeout.
Given the approach of output string pattern matching, we already assume the predictable pattern of our internal tooling, maybe it's not worthwhile to implement the time-out pattern catching when there's no output.
I also put this block after checking that stripe-mock is alive, to reduce chance of hanging when somehow the process exits.

Alternatively, I thought about using `lsof -Pan -p <stripe mock PID> -i` and get the response. This can make our port finding independent of stripe-mock implementation, but the downside is perhaps how we have variance to lsof string output which makes pattern matching more fragile. Mine looks something like this.
```
COMMAND     PID             USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
stripe-mo 20288 mickjermsurawong    3u  IPv6 0x4dxxxxx      0t0  TCP *:51516 (LISTEN)
```

mickjermsurawong-stripe: I believe `readPartial` in ruby's implementation will have the same problem on "It blocks only if ios has no data immediately available." [doc](https://ruby-doc.org/core-2.2.2/IO.html#method-i-readpartial)

mickjermsurawong-stripe:
thank you ob, I implemented a similar pattern for finding port to that in Ruby! ptal @ob-stripe

mickjermsurawong-stripe:
hmm cibot doesn't reassign. let me try again:
r? @ob-stripe

src/test/java/com/stripe/StripeMockProcess.java
ob-stripe: Nice, thanks for digging into this Mick. I think we can just leave it at this. There's really no reason for the stripe-mock process to be alive and not output anything, and if that ever happens we can just look into it.

src/test/java/com/stripe/StripeMockProcess.java
```diff
@@ -0,0 +1,153 @@
...
+            getPathFixture()
+        );
+
+    try {
+      process = processBuilder.start();
+    } catch (IOException e) {
+      System.out.println(String.format(
+          "Error while starting stripe-mock, port = %d, fixtures = %s and %s, "
+              + "error message = %s",
+          port, getPathSpec(), getPathFixture(), e.getMessage()
```
ob-stripe: Can you update this message to remove the port (since it will be `null` now that stripe-mock chooses the port itself)?

mickjermsurawong-stripe: whoops my bad. fixed!

mickjermsurawong-stripe:
Addressed the PR feedback. Thank you for the thorough review!
ptal @ob-stripe

PR #660, opened 2019-01-31
Add support for Account and Person API changes for 2019-02-18

remi-stripe:
This PR contains the changes related to the new API version `2019-02-19`. This changes the `Account` and `Person` resources extensively but keeps the backwards compatible approach.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Person.java
+++ b/src/main/java/com/stripe/model/Person.java
@@ -10,6 +10,7 @@ import java.math.BigDecimal;
 import java.util.List;
 import java.util.Map;
 
+import lombok.AccessLevel;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
@@ -39,6 +40,7 @@ public class Person extends ApiResource implements MetadataStore<Person>, HasId 
   String maidenName;
   @Getter(onMethod = @__({@Override})) Map<String, String> metadata;
   String phone;
+  Relationship relationship;
   Requirements requirements;
   @SerializedName("ssn_last_4_provided") Boolean ssnLast4Provided;
   Verification verification;
@@ -84,8 +86,6 @@ public class Person extends ApiResource implements MetadataStore<Person>, HasId 
     return null;
   }
 
-  // TODO: Move this to a top level class instead of duplicating
-  // with the LegalEntity resource
   @Getter
   @Setter
   @EqualsAndHashCode(callSuper = false)
@@ -95,8 +95,6 @@ public class Person extends ApiResource implements MetadataStore<Person>, HasId 
     Long year;
   }
 
-  // TODO: Move this to a top level class instead of duplicating
-  // with the LegalEntity resource
   @Getter
   @Setter
   @EqualsAndHashCode(callSuper = false)
@@ -116,11 +114,25 @@ public class Person extends ApiResource implements MetadataStore<Person>, HasId 
   public static class Relationship extends StripeObject {
     Boolean accountOpener;
     Boolean director;
-    Boolean executive;
     Boolean owner;
     BigDecimal percentOwnership;
-    Boolean representative;
     String title;
+
+    /**
+     * The {@code executive} attribute.
+     * @deprecated This is not required anymore.
+     * @see <a href="https://stripe.com/docs/upgrades#2019-02-19">API version 2019-02-19</a>
+     */
+    @Deprecated
+    Boolean executive;
+
+    /**
+     * The {@code representative} attribute.
+     * @deprecated This is not required anymore.
+     * @see <a href="https://stripe.com/docs/upgrades#2019-02-19">API version 2019-02-19</a>
+     */
+    @Deprecated
+    Boolean representative;
   }
 
   @Getter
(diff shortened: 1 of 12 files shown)


src/main/java/com/stripe/model/Person.java
```diff
@@ -132,16 +144,62 @@ protected String getInstanceUrl() {
...
   @EqualsAndHashCode(callSuper = false)
   public static class Verification extends StripeObject {
     String details;
     String detailsCode;
-    String document;
-    String documentBack;
+    VerificationDocument document;
+    // TODO: Figure out how to handle the change in shape and deprecate the next 2
+    //String document;
+    //String documentBack;
```
remi-stripe: @ob-stripe Can you point me to the right solution to handle this? I know we do that in the library today, I'm just not sure where :p

ob-stripe: We'll need to write yet another custom deserializer :(

This is basically the same case as `evidence` on `Dispute` objects. You'll need to do something like this:
1. Name the new `VerificationDocument` attribute `documentSubObject`.
2. Write a custom deserializer (similar to [`DisputeDataDeserializer`](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/DisputeDataDeserializer.java)) that looks at the shape of the `document` attribute. If it's a string, then use its value to fill `document`. If if's a hash, then deserialize it as a `VerificationDocument` and fill `documentSubObject`.

src/main/java/com/stripe/model/Person.java
```diff
@@ -132,16 +144,62 @@ protected String getInstanceUrl() {
...
   }
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class VerificationDocument extends StripeObject {
+    @Getter(lombok.AccessLevel.NONE) @Setter(lombok.AccessLevel.NONE) ExpandableField<File> back;
+    String details;
+    String detailsCode;
+    @Getter(lombok.AccessLevel.NONE) @Setter(lombok.AccessLevel.NONE) ExpandableField<File> front;
```
ob-stripe: Try changing these (and for `back` too) to just `AccessLevel.NONE`, without the `lombok.` prefix.

remi-stripe: Can you clarify why that's needed since other expandable fields have both?

ob-stripe: That's not the case AFAICT. On the latest `master`:
```
❯ ag "lombok\.AccessLevel\.NONE" --stats
0 matches
```

ob-stripe:
@remi I've pushed a commit that adds a custom deserializer for `Person.Verification.document` and added a few tests, ptal.

src/main/java/com/stripe/model/PersonVerificationDeserializer.java
```diff
@@ -0,0 +1,88 @@
...
+public class PersonVerificationDeserializer implements JsonDeserializer<Person.Verification> {
+  /**
+   * Deserializes a person.verification JSON payload into a {@link Person.Verification} object.
+   */
+  @Override
+  public Person.Verification deserialize(JsonElement json, Type typeOfT,
+      JsonDeserializationContext context) throws JsonParseException {
+    Gson gson = new GsonBuilder()
+        .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+        .registerTypeAdapter(ExpandableField.class, new ExpandableFieldDeserializer())
```
remi-stripe: Why don't you need to register all deserializers?

src/main/java/com/stripe/model/PersonVerificationDeserializer.java
```diff
@@ -0,0 +1,88 @@
...
+        .create();
+    if (json.isJsonNull()) {
+      return null;
+    }
+
+    if (!json.isJsonObject()) {
+      throw new JsonParseException("Dispute type was not an object, which is problematic.");
```
remi-stripe: should not mention dispute

src/main/java/com/stripe/model/PersonVerificationDeserializer.java
ob-stripe: We only use this instance to deserialize `Person.Verification`, `Person.VerificationDocument` and `File`, so we don't need any of the other custom deserializers. (Ideally we'd always use the global deserializer instance in `ApiResource`, but then that'd cause an infinite recursion problem, cf. #433).

src/main/java/com/stripe/model/Person.java
```diff
@@ -132,16 +144,123 @@ protected String getInstanceUrl() {
...
+     *
+     * @return the {@code document_back} attribute
+     * @deprecated Prefer using the {@link #getDocumentSubObject} method instead.
+     * @see <a href="https://stripe.com/docs/upgrades#2019-02-19">API version 2019-02-19</a>
+     */
+    @Deprecated
+    @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<File> documentBack;
+
+    // <editor-fold desc="document">
+    @Deprecated
```
ob-stripe: We should copy the Javadoc comments for `document` on the `getDocument()` and `getDocumentObject()` getters, otherwise library users won't be able to see it. Same thing for `documentBack` -> `getDocumentBack()` + `getDocumentBackObject()`.

remi-stripe: @ob-stripe Fixed, can you confirm it's what you wanted?

ob-stripe: Yup, thanks!

PR #661, opened 2019-01-31
Add Stripe client telemetry to request headers

jameshageman-stripe:
Adds opt-in client telemetry similar to https://github.com/stripe/stripe-node/pull/557, https://github.com/stripe/stripe-go/pull/766, https://github.com/stripe/stripe-ruby/pull/696, etc.

Request metrics are recorded in `LiveStripeResponseGetter` and stored on a thread safe buffer using `ConcurrentLinkedQueue`. If the buffer exceeds 100 elements, metrics are dropped. When a request is sent, if the buffer is not empty, one of the metrics is popped off and sent as JSON in the `X-Stripe-Client-Telemetry` header.

Clients can enable the telemetry using:

```java
Stripe.enableTelemetry = true;
```

r? @brandur-stripe 
cc @dcarney-stripe @akropp-stripe

### DIFF
--- a/src/main/java/com/stripe/Stripe.java
+++ b/src/main/java/com/stripe/Stripe.java
@@ -18,8 +18,8 @@ public abstract class Stripe {
   public static volatile String apiKey;
   public static volatile String apiVersion;
   public static volatile String clientId;
-  public static volatile String partnerId;
   public static volatile boolean enableTelemetry = false;
+  public static volatile String partnerId;
 
   // Note that URLConnection reserves the value of 0 to mean "infinite
   // timeout", so we use -1 here to represent an unset value which should
--- a/src/main/java/com/stripe/Stripe.java
+++ b/src/main/java/com/stripe/Stripe.java
@@ -19,6 +19,7 @@ public abstract class Stripe {
   public static volatile String apiVersion;
   public static volatile String clientId;
   public static volatile String partnerId;
+  public static volatile boolean enableTelemetry = false;
 
   // Note that URLConnection reserves the value of 0 to mean "infinite
   // timeout", so we use -1 here to represent an unset value which should
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -535,11 +535,11 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
       ApiResource.RequestMethod method, String url, Map<String, Object> params,
       Class<T> clazz, ApiResource.RequestType type, RequestOptions options)
       throws StripeException {
-    long requestStart = System.currentTimeMillis();
+    long requestStartMs = System.currentTimeMillis();
 
     StripeResponse response = rawRequest(method, url, params, type, options);
 
-    long requestDurationMS = System.currentTimeMillis() - requestStart;
+    long requestDurationMs = System.currentTimeMillis() - requestStartMs;
 
     int responseCode = response.code();
     String responseBody = response.body();
@@ -562,7 +562,7 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
     }
 
     if (Stripe.enableTelemetry && prevRequestMetrics.size() < MAX_REQUEST_METRICS_BUFFER_SIZE) {
-      prevRequestMetrics.add(new RequestMetrics(requestId, requestDurationMS));
+      prevRequestMetrics.add(new RequestMetrics(requestId, requestDurationMs));
     }
 
     return resource;
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -43,6 +43,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Scanner;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLSocketFactory;
 
@@ -50,6 +51,7 @@ import lombok.Cleanup;
 
 public class LiveStripeResponseGetter implements StripeResponseGetter {
   private static final String DNS_CACHE_TTL_PROPERTY_NAME = "networkaddress.cache.ttl";
+  private static final int MAX_REQUEST_METRICS_BUFFER_SIZE = 100;
 
   private static final class Parameter {
     public final String key;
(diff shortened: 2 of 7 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -523,8 +534,12 @@ private static StripeResponse rawRequest(
       ApiResource.RequestMethod method, String url, Map<String, Object> params,
       Class<T> clazz, ApiResource.RequestType type, RequestOptions options)
       throws StripeException {
+    long requestStart = System.currentTimeMillis();
```
devshorts: Some of the spacing in this file doens't really line up, can you make sure the spacing is correct? (if you use intellij the standard formatter will probably work just fine)

src/main/java/com/stripe/net/RequestMetrics.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    public RequestMetrics(String requestId, long requestDurationMS) {
+        this.request_id = requestId;
+        this.request_duration_ms = requestDurationMS;
+    }
+
+    public Payload payload() {
+        Payload p = new Payload();
+        p.last_request_metrics = this;
```
devshorts: I think this will cause a memory leak as you're going to be holding onto an indefinite chain of references, so the GC can never garbage collect the request metrics objects.  It would be better to create a new request metrics payload that is a copy of the current

src/main/java/com/stripe/net/RequestMetrics.java
```diff
@@ -0,0 +1,21 @@
+package com.stripe.net;
+
+public class RequestMetrics {
+    public String request_id;
+    public long request_duration_ms;
```
devshorts: does GSON Support annotations for json field aliases? It would be better to not mix non standard java variable names just for json support. https://google.github.io/gson/apidocs/com/google/gson/annotations/SerializedName.html

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
jameshageman-stripe: I agree that this looks a little off, but I followed the spacing that was already being used. Reformatting with IntelliJ's formatter would end up changing the whole file.

src/main/java/com/stripe/net/RequestMetrics.java
jameshageman-stripe: Yes! Updated.

jameshageman-stripe:
Updated! 
ptal @akropp-stripe @brandur-stripe

src/main/java/com/stripe/net/RequestMetrics.java
jameshageman-stripe: Chatted in person but I'll record the discussion here:

This won't cause a memory leak because `Payload` and `RequestMetrics` are different types, so we aren't creating any chain of `RequestMetrics` objects.

That being said, this pattern of using an inner class was kind of strange, so I pulled it out into `ClientTelemetryPayload`.

src/main/java/com/stripe/Stripe.java
```diff
@@ -19,6 +19,7 @@
   public static volatile String apiVersion;
   public static volatile String clientId;
   public static volatile String partnerId;
+  public static volatile boolean enableTelemetry = false;
```
brandur-stripe: Can you keep this list alphabetized?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
brandur-stripe: Can we call this one `requestStartMS` (i.e. add the `MS` prefix here as well)? It's not a time object like other languages.

jameshageman-stripe:
@brandur-stripe 

> All the tests around threading are nice, but I'm personally pretty comfortable assuming that ConcurrentLinkedQueue's implementation is correct and isn't going to cause any problems around thread safety.

I'd assume the same thing, but I think the purpose of the test is to make sure our logic is thread safe, not `ConcurrentLinkedQueue`'s.

> I'd prefer to just augment verifyRequest to be able to deal with headers, and then just check that we got the telemetry one. If we were going to move everything over to the new testing system, then that would be a totally different story.

Makes sense, let me see what I can do!

brandur-stripe:
> I'd assume the same thing, but I think the purpose of the test is to make sure our logic is thread safe, not `ConcurrentLinkedQueue`'s.

Ah, right. I just meant that our logic is a pretty thin shim around the queue — `ConcurrentLinkedQueue` is doing all the heavy lifting when it comes to guaranteeing thread safety.

jameshageman-stripe:
@brandur-stripe re:
> I'd prefer to just augment verifyRequest to be able to deal with headers, and then just check that we got the telemetry one. If we were going to move everything over to the new testing system, then that would be a totally different story.

I poked around to try to see how I could write the tests using our existing tools, and I ran into the following problem:

`verifyRequest` verifies calls to the public `StripeResponseGetter` interface, but the headers are added behind the scenes by the static method `LiveStripeResponseGetter::getHeaders`. Mockito does not support mocking static code, so trying to spy on the headers will require a non-trivial refactor.

Maybe it's a result of how I modified the `LiveStripeResponseGetter`, but I think it will be more difficult than it seems to rewrite the new tests using mocks and spies. Do you think we could reconsider adding `MockWebServer`? We have used the same end-to-end style strategy to test the implementation of client telemetry in [stripe-go](https://github.com/stripe/stripe-go/blob/20fc3632aeb2d005900fe729462524927b553962/stripe_test.go#L246-L279), [stripe-python](https://github.com/stripe/stripe-python/blob/448f837bee57cb5736a95c447ecbc27332690c09/tests/test_integration.py#L145-L179), and [stripe-node](https://github.com/stripe/stripe-node/blob/master/test/telemetry.spec.js#L71-L92).

ob-stripe:
> Mockito does not support mocking static code, so trying to spy on the headers will require a non-trivial refactor.

Yeah, that's the same conclusion I reached back when I wrote `verifyRequest`. It was good enough for most of our tests, where we mostly want to assert that each resource method sends the correct request (i.e. HTTP verb + path) to the API, but it doesn't let us inspect headers.

I think `MockWebServer` might be the path of least resistance here, and it's just a dev dependency so it's not too bad.

brandur-stripe:
Okay, thanks for looking into it at least!

I don't feel too strongly about the new dependency, but if we decide to include it, it'd be nice to have some sort of idea of what our end game is so that we don't end up with half the tests running on one system and half running on the other a year for now. Stated differently: is `MockWebServer` the direction we want to start moving overall?

But anyway, if OB's happy with it, then LGTM too.

jameshageman-stripe:
Thanks @brandur-stripe. I'll leave it to you and @ob-stripe to merge this when you're ready.

brandur-stripe:
Ack. @ob-stripe Mind taking a quick look before we merge?

ob-stripe:
Released as 7.19.0.

PR #665, opened 2019-02-01
Configure Gradle signing plugin to use gpg-agent

ob-stripe:
r? @brandur-stripe @hans-stripe 

This PR configures the Gradle signing plugin to use gpg-agent instead of old-style GPG keyring files: https://docs.gradle.org/current/userguide/signing_plugin.html#sec:using_gpg_agent.

This needs to be coordinated with a change to our internal release tool, so let's not merge this just yet! You can test this PR by calling `./gradlew clean sign`, or if you want to specify a custom GPG dir and key ID (like the release tool will):
```
GNUPGHOME="/path/to/gnupg" GRADLE_OPTS="-Dorg.gradle.project.signing.gnupg.keyName=12345678" ./gradlew clean sign
```

### DIFF
--- a/deploy.gradle
+++ b/deploy.gradle
@@ -94,7 +94,12 @@ afterEvaluate { project ->
     signing {
         required { isReleaseBuild() &&
         (gradle.taskGraph.hasTask("uploadArchives") || gradle.taskGraph.hasTask("publish"))}
+        useGpgCmd()
         sign configurations.archives
+    }
+
+    tasks.withType(Sign) {
+        onlyIf { isReleaseBuild() && project.hasProperty('signing.gnupg.keyName') }
     }
 
     task makeJavadocs(type: Javadoc, dependsOn: delombok) {


ob-stripe:
🤔 Hrm, not sure why this fails on Travis. Apparently this causes the `sign` task to be executed, which I don't think was the case before. I'll investigate.

hans-stripe:
https://github.com/stripe/stripe-android/blob/master/stripe/deploy.gradle and https://github.com/stripe-ios/stripe-terminal-android/blob/master/stripeterminal/deploy.gradle also do this signing step but don't seem to have the `useGpgCmd()` line. I think this means they're probably both good to go, and also they might have a hint for how to make the tests pass.

ob-stripe:
The Travis issue appears to be caused by a bug with `useGpgCmd()`: https://github.com/gradle/gradle/issues/5064. I've set up the recommended workaround and updated the PR, and Travis no longer tries to create signatures.

@hans-stripe Unfortunately, I'm pretty sure that both those projects still use the keyring file setup. Gradle's documentation is pretty clear that unless you use `useGpgCmd()`, the signing plugin defaults to the old implementation which needs a keyring file.

hans-stripe:
Argh I read the diffs backwards, too early in the morning. Good to know that those need to be fixed too.

ob-stripe:
Going to attempt a release with this. Will revert if needed.

ob-stripe:
Released as 7.18.1.

PR #669, opened 2019-02-05
[wip] Typed params builder pattern

mickjermsurawong-stripe:
- This PR explores the builder pattern for typed params. Particularly, it is nested builder pattern to hide the sub-entities in the final params created.
  - Each internal builder cannot be built independently, but only receive more content, children or to the parent builder. The syntax is as follows;
     - `begin*` for entering subhash description
     - `beginArray*` for entering an array descriptoin
     - `done` returns the description to outer builder.
     - `build` only available to top-level builder to return finishes the main param object
  - The nested sub-entities are never exposed to the user; only the nested builders are. The motivation here is that we can rely on OpenAPI spec for each URL operation without trying to share the subhash with others. Each method will have its own typed parameters.
- This approach will still coexist with passing top-level hash object for legacy reason, and for passing beta or gated params that are not publicly available.

However, there are the following downsides:
- This approach is inconsistent with other typed library where it takes sub-entities.
- Clunky expression when the params have multiple nested entities, compared to current approach where each sub-hash can be built independently and add to the top-level hash. 
- If our ideal world is to have builder pattern where each input argument is an actual object instead of builders, instead of chained nested pattern, maybe this is not a good step to move into.

While not exposing the internal sub-classes to the merchant, builder pattern still allows navigation and exploring the params. For testing purposes, however, I'm considering serializing this structure to a hash so merchant can still test their param building.

Please see the test `AccountCreateParamsTest.java` for the interface. 
The implementation although verbose will be autogen if we are happy with this approach. (I also updated to 1.8 just to get easier syntax as I was exploring this approach -- will make a java7 compatible one in the autogen code)

r? @brandur-stripe @ob-stripe
Thanks @rattrayalex-stripe for the discussions!
cc @stripe/api-libraries


src/main/java/com/stripe/model/AccountCreateParams.java
```diff
@@ -0,0 +1,182 @@
...
+      return new AccountCreateParams(type, legalEntityBuilder.build());
+    }
+  }
+
+  /**
+   * Inner classes and their builders.
+   */
+  @Getter
+  @Setter
+  static class AdditionalOwner {
```
mickjermsurawong-stripe: Each internal class has package visibility and hidden from the users.

src/test/java/com/stripe/model/AccountCreateParamsTest.java
```diff
@@ -0,0 +1,90 @@
...
+    AccountCreateParams.Address address2 = legalEntity.getAdditionalOwners().get(1).getAddress();
+    assertEquals("London", address2.getCity());
+    assertEquals("UK", address2.getCountry());
+    assertEquals("456", address2.getPostalCode());
+
+    assertEquals("foo_type", createParams.getType());
+  }
+
+  @Test
+  public void testBuilderCreateParamsWithIterations() {
```
mickjermsurawong-stripe: Users can still programmatically manipulate a subgraph of this builder pattern.
However this nested builder pattern makes it hard to build sub-hash independently in different scope.

rattrayalex-stripe:
I think getting some input from Java folks, like @esandler-stripe or wall.

rattrayalex-stripe:
How hard would it be to allow people to pass builders to builders? eg; allow both `.beginAddress().setCity()...` and `.setAddress(AddressBuilder)`, or similar.

mickjermsurawong-stripe:
@rattrayalex-stripe that won't be any much more work. 
I like that approach. Seems like we won't be exposing the sub-entities of the params, and still allowing ease of composing builder in a more modular way (composing each sub-builder independently).
Although I suspect it's not too common a pattern; builder usually takes in an actual argument instead of another builder itself..

esandler-stripe:
@graj-stripe is another person would have good insights into java style. Mind taking a look g$?

src/main/java/com/stripe/model/AccountCreateParams.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class AccountCreateParams {
+
+  private String type;
```
graj-stripe: I think typically with a builder you want to make the class being built immutable - I would make these `final` and drop the `@Setter` above

src/test/java/com/stripe/model/AccountCreateParamsTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Map;
+import java.util.stream.IntStream;
+import org.junit.Test;
+
+public class AccountCreateParamsTest {
+  @Test
+  public void testBuilderCreateParams() {
+    AccountCreateParams.Builder builder = AccountCreateParams.builder();
+    AccountCreateParams createParams = builder
+        .beginLegalEntity()
```
graj-stripe: For me this is an odd builder. The manual indentation here makes it easy to understand in this review, but I think a typical automatic code formatter will put these at the same indentation level, making this really difficult to understand visually.

I would probably just go with a protobuf style builder (with lots of parentheses) even though it's more verbose.

src/main/java/com/stripe/model/AccountCreateParams.java
graj-stripe: It's not clear to me what this buys us when the builder for the object itself is public

brandur-stripe:
Wow, this is cool @mickjermsurawong-stripe! Very much looking forward to the day that we have nice typed systems like this one for parameter building across the whole library. Talk about upgrade incentive!

mickjermsurawong-stripe:
thank you @graj-stripe  
For constructing request parameters, I would like users not to use the internal objects of the params elsewhere. The internal builder itself also doesn't have a public `build` method.. So I made the interface such that its only use case is to be built for the top-level parameter. 
> It's not clear to me what this buys us when the builder for the object itself is public

Ah thank you for reference to google protobuff builder. If that's the standard pattern, I think that's something we can offer here too. To you concern about the nested object being hard to mutate. I suspect the lifecycle of request parameters here would be quite short; user create the request param close to when our interface accepts to to make http request. Mutating the params might not be too common of a use case. If they have many parameter requests to make, I think creating new param request instead of mutating them might be a good practice too.

@brandur-stripe  
- This is exploratory PR for when we decide on requirements to do typed params before releasing Java sdk 8.0. Please let me know if you have concern on inconsistency with other typed libs. 
- Also with this approach, each parameter request will have its own sub-entities; we will see duplicate sub-hashes (although not exposed to users). However, this will address the problem sharing a param that were originally common but later differentiated. I think @remi-stripe agree with having dedicated params for each method too, if that can be autogen.

src/test/java/com/stripe/model/AccountCreateParamsTest.java
esandler-stripe: +1 as well to the protobuf pattern for the visualization benefits. Especially in the future world of PaymentIntents, it seems like more Stripe APIs have requests that are nicely broken down into logical subhashes. Having a separate subbuilder for each logical subhash makes it easy to visualize the requests and also to have code reuse as people move from Charges -> PaymentIntents etc.

mickjermsurawong-stripe: ack! thank you everyone for the feedback

mickjermsurawong-stripe:
Thank you everyone for the feedback here! On which particular implementations of builder pattern, I think it is agreed to have an intermediate sub-builders explicitly constructed and passed to parent builders.

I assume here that Java builder pattern is a common one, and in fact preferred over normal constructor. At the cost of more template code in the builder.
The benefits as I understand are:
- More explicit than having a constructor with many arguments but no keyword argument like in Ruby.  
- Allows incremental building of object and ensures that object is used when it is ready (plus allows final internal validation at the build method) By contrast, having instantiated object and exposed setters allows abuse of premature object usage if it's still in the "building" phase -- an illegal state that it shouldn't be used yet.

Please let me know if there's argument against using the builder pattern in general

PR #670, opened 2019-02-06
Changes to the Payment Intent APIs for the next API version

remi-stripe:
We are going to ship some changes to the Payment Intent APIs in the next API version. Though are fairly self-contained and pave the way for future changes. Changes are:

* `status`: remove `require_source` and `require_source_action` and add `requires_payment_method` and `requires_action`
* `next_source_action` renamed to `next_action`
* `allowed_source_types` renamed to `payment_method_types`
* `authorize_with_url` renamed to `redirect_to_url` both in the type of next action and in the hash that describes that next action.

cc @stripe/api-libraries @danwang-stripe 

cc @mickjermsurawong-stripe To confirm naming for the new classes make sense (now that they are inline)

### DIFF
--- a/src/main/java/com/stripe/model/PaymentIntent.java
+++ b/src/main/java/com/stripe/model/PaymentIntent.java
@@ -18,7 +18,6 @@ import lombok.Setter;
 public class PaymentIntent extends ApiResource implements MetadataStore<PaymentIntent>, HasId {
   @Getter(onMethod = @__({@Override})) String id;
   String object;
-  List<String> allowedSourceTypes;
   Long amount;
   Long amountCapturable;
   Long amountReceived;
@@ -36,8 +35,9 @@ public class PaymentIntent extends ApiResource implements MetadataStore<PaymentI
   PaymentIntentLastPaymentError lastPaymentError;
   Boolean livemode;
   @Getter(onMethod = @__({@Override})) Map<String, String> metadata;
-  PaymentIntentSourceAction nextSourceAction;
+  NextAction nextAction;
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Account> onBehalfOf;
+  List<String> paymentMethodTypes;
   String receiptEmail;
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Review> review;
   ShippingDetails shipping;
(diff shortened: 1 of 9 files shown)


mickjermsurawong-stripe:
@remi-stripe 
- the payment `NextAction`, `NextActionRedirectToUrl` nested in `PaymentIntent` looks great with me. 
I see that in the the latest OpenAPI spec. (But client metadata is wrong, I will go and update it to reflect similarly this PR)
- Haven't seen the change on `allowed_source_types`  in pay-server yet.
- We currently don't have enum on the `status`, so not sure if it's worth specifying that in comments. But that means we have to maintain the comment too which is inconsistent. I'm okay with what it is now.

remi-stripe:
@mickjermsurawong-stripe thanks a lot for the review, this definitely helps vet the approach!

remi-stripe:
@mickjermsurawong-stripe API version is going out tomorrow so marking this as ready to merge now that stripe-mock is updated and assigning to you for a full review.

src/main/java/com/stripe/model/PaymentIntent.java
```diff
@@ -335,4 +355,20 @@ public void setDestinationObject(Account c) {
     }
     // </editor-fold>
   }
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class NextAction extends StripeObject {
+    NextActionRedirectToUrl redirectToUrl;
+    String type;
```
mickjermsurawong-stripe: I think we need the additional field? At least it shows up in my autogen PR https://github.com/stripe/stripe-java/pull/676/files.. Also confirm that it's present in public OpenAPI
```
    Map<String, Object> useStripeSdk;
```

mickjermsurawong-stripe:
ptal @remi-stripe, maybe we are missing a field on `NextAction`

src/main/java/com/stripe/model/PaymentIntent.java
remi-stripe: @mickjermsurawong-stripe This is undocumented/secret right now though so not something we'd want to ship for now.

mickjermsurawong-stripe:
ah good to know.. Will make a PR to undocument that in pay-server.
LGTM

src/main/java/com/stripe/model/PaymentIntent.java
remi-stripe: Confirmed, we want the parameter/enum value but this property should **not** be in the libraries since it's opaque and specific to stripe.js

PR #672, opened 2019-02-08
Parity with 7.20 - subscription schedule and payment method

mickjermsurawong-stripe:
r? @remi-stripe 
- This PR includes addition to subscription schedules resources and endpoints
Not sure you've worked on Java already. But I suspect this is what you would have written by hand.
This correspond to change in pay-server https://git.corp.stripe.com/stripe-internal/pay-server/pull/130623
cc @stripe/api-libraries


src/main/java/com/stripe/model/PaymentIntent.java
```diff
@@ -134,7 +134,7 @@
    * fulfill a payment using the provided source.
    */
   @SerializedName("next_source_action")
-  PaymentIntentSourceAction nextSourceAction;
+  PaymentIntentNextSourceAction nextSourceAction;
```
remi-stripe: This is unrelated to the review (or a change in openapi). Is this because of some client-metadata changes?
Asking out of curiosity, this should go away since we're now doing `next_action` instead anyway in the next API version

src/main/java/com/stripe/model/PaymentIntentNextSourceActionValueAuthorizeWithUrl.java
```diff
@@ -10,7 +10,7 @@
 @Getter
 @Setter
 @EqualsAndHashCode(callSuper = false)
-public class PaymentIntentSourceActionValueAuthorizeWithUrl extends StripeObject {
+public class PaymentIntentNextSourceActionValueAuthorizeWithUrl extends StripeObject {
```
remi-stripe: So this is tricky. We have a custom deserializer for that class. When you're renaming, you're breaking the custom deserializer. Is this expected? Is this part of autogen already too?

cc @ob-stripe who can explain this better than I would.

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -0,0 +1,406 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class SubscriptionSchedule extends ApiResource
+    implements HasId, MetadataStore<SubscriptionSchedule> {
+  /**
+   * Either `charge_automatically`, or `send_invoice`. When charging automatically, Stripe will
```
remi-stripe: Minor but I think we usually put those in `<code> charge_automatically <code>`. cc @ob-stripe to decide if it's worth doing.

Also for autogen we likely want enums for all of this right? (#idontknowjava)

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -0,0 +1,406 @@
...
+   */
+  @SerializedName("billing")
+  String billing;
+
+  /**
+   * Define thresholds at which an invoice will be sent, and the subscription advanced to a new
+   * billing period.
+   */
+  @SerializedName("billing_thresholds")
+  Subscription.BillingThresholds billingThresholds;
```
remi-stripe: I like that we re-use it "magically". You likely do that with client_metadata though after our talk today I still really hope that it's the same resource in pay-server and so we should be able to know that automatically.

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -0,0 +1,406 @@
...
+  ExpandableField<Customer> customer;
+
+  /** Unique identifier for the object. */
+  @Getter(onMethod = @__({@Override}))
+  @SerializedName("id")
+  String id;
+
+  /** The subscription schedule's default invoice settings. */
+  @SerializedName("invoice_settings")
+  SubscriptionScheduleInvoiceSettings invoiceSettings;
```
remi-stripe: This is one of the examples where we are "nesting it" but really it's likely that in 4 months we share that hash with another resource and so maybe it should be top-level.

I gave up for .NET and go and did what you did (but not nested because those languages can't)

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -0,0 +1,406 @@
...
+  /** Behavior of the subscription schedule and underlying subscription when it ends. */
+  @SerializedName("renewal_behavior")
+  String renewalBehavior;
+
+  /**
+   * Interval and duration at which the subscription schedule renews for when it ends if
+   * `renewal_behavior` is `renew`.
+   */
+  @SerializedName("renewal_interval")
+  SubscriptionScheduleRevision.RenewalInterval renewalInterval;
```
remi-stripe: I think this is backwards. IMO the class should be in `SubscriptionSchedule` and `SubscriptionScheduleRevision` should do this instead:
```
SubscriptionSchedule.RenewalInterval renewalInterval;
```

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -0,0 +1,406 @@
...
+    return (this.subscription != null) ? this.subscription.getExpanded() : null;
+  }
+
+  public void setSubscriptionObject(Subscription expandableObject) {
+    this.subscription =
+        new ExpandableField<Subscription>(expandableObject.getId(), expandableObject);
+  }
+
+  /** Retrieves the list of your subscription schedules. */
+  public SubscriptionScheduleCollection list(Map<String, Object> params) throws StripeException {
```
remi-stripe: @ob-stripe totally random but do we really need those duplicate collection classes? (nothing to do with autogen, just realizing I haven't asked in a while)

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -0,0 +1,406 @@
...
+    @Setter(lombok.AccessLevel.NONE)
+    ExpandableField<Coupon> coupon;
+
+    /** The end of this phase of the subscription schedule. */
+    @SerializedName("end_date")
+    Long endDate;
+
+    /** Plans to subscribe during this phase of the subscription schedule. */
+    @SerializedName("plans")
+    List<SubscriptionScheduleRevision.PhaseItem> plans;
```
remi-stripe: Same as before, I think this should exist in this class. Also some FUD on naming of the class. I wish the property was named `items` and not `plans` and it'd make more sense, but because it's not I'm less a fan.

src/main/java/com/stripe/model/SubscriptionScheduleInvoiceSettings.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.google.gson.annotations.SerializedName;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class SubscriptionScheduleInvoiceSettings extends StripeObject {
```
remi-stripe: Can you clarify why that one is top level instead of in `SubscriptionSchedule` and then used as `SubscriptionSchedule.InvoiceSettings` in the revision class?

src/main/java/com/stripe/model/SubscriptionScheduleRevision.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Map;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class SubscriptionScheduleRevision extends StripeObject implements HasId {
+  /** Time at which the object was created. Measured in seconds since the Unix epoch. */
```
remi-stripe: Should all of those be real javadocs? (sorry train of thoughts, nothing specific to that property, just wondered)

remi-stripe:
@mickjermsurawong-stripe Added some comments but I think I can mostly copy-paste most of this in java (though it be better after some of the things I mentioned)

src/main/java/com/stripe/model/PaymentIntent.java
mickjermsurawong-stripe: Changed to nested `NexAction`

src/main/java/com/stripe/model/PaymentIntentNextSourceActionValueAuthorizeWithUrl.java
mickjermsurawong-stripe: Changed to `NextActionRedirectToUrl`

src/main/java/com/stripe/model/SubscriptionSchedule.java
mickjermsurawong-stripe: Now nested in `SubscriptionSchedule`

src/main/java/com/stripe/model/SubscriptionScheduleInvoiceSettings.java
mickjermsurawong-stripe: Now nested in `SubscriptionSchedule`

src/main/java/com/stripe/model/SubscriptionSchedule.java
mickjermsurawong-stripe: also nested in `SubscriptionSchedule`

src/main/java/com/stripe/model/Account.java
```diff
@@ -38,6 +38,14 @@
   @Setter(lombok.AccessLevel.NONE)
   ExpandableField<File> businessLogo;
 
+  /**
+   * (ID of a [file upload](https://stripe.com/docs/guides/file-upload)) A larger logo for the
+   * account that will be used in Checkout instead of the smaller logo and without the account's
+   * name next to it if provided. Must be at least 128px x 128px.
+   */
+  @SerializedName("business_logo_large")
+  String businessLogoLarge;
```
mickjermsurawong-stripe: @remi-stripe  New field appearing, not in binding, but now also in APIref 
https://stripe.com/docs/api/accounts/object#account_object-business_logo_large

src/main/java/com/stripe/model/Person.java
```diff
@@ -200,10 +200,6 @@ public Person delete(Map<String, Object> params, RequestOptions options) throws
     @SerializedName("director")
     Boolean director;
 
-    /** Whether the person has a significant control of the account’s legal entity. */
-    @SerializedName("executive")
-    Boolean executive;
```
mickjermsurawong-stripe: `executive` field in now hidden in OpenAPI spec. I'll make a change in pay-server to add it back..
Put it in TODO here
https://git.corp.stripe.com/stripe-internal/sdk-autogen-java/blob/8eac5b0e7eedeedbfaff8980c6b0005d24ef961e/generated/diff_change_log_fields.md#todo-add-back-from-pay-server

mickjermsurawong-stripe:
ptal @remi-stripe 
This should be consistent with the latest `stripe-java` 7.20 now.
What i did is also rebasing `autogen/8.0-beta` to the latest `7.20` master, so what you see is the diff from the latest change in OpenAPI spec from this PR https://git.corp.stripe.com/stripe-internal/pay-server/pull/130623

src/main/java/com/stripe/model/Account.java
remi-stripe: Yeah we just added it for backwards compatibility reasons. Since go/.net will be on the next API version on 2/19 and java will be on 2/26 I don't think it's a big deal if we don't add it immediately/ignore it.

src/main/java/com/stripe/model/PaymentIntent.java
```diff
@@ -554,6 +554,42 @@ public PaymentIntent capture(Map<String, Object> params, RequestOptions options)
...
+    @SerializedName("type")
+    String type;
+
+    /**
+     * When confirming a PaymentIntent with Stripe.js, Stripe.js depends on the contents of this
+     * dictionary to invoke authentication flows. The shape of the contents is subject to change and
+     * is only intended to be used by Stripe.js.
+     */
+    @SerializedName("use_stripe_sdk")
+    Map<String, Object> useStripeSdk;
```
remi-stripe: Same as the other PR we likely should remove this in the library.

src/main/java/com/stripe/model/Person.java
remi-stripe: It's going away, please keep it removed! (also we should avoid linking to internal links on public repos!)

mickjermsurawong-stripe: ah got it! will take no action on this.

src/main/java/com/stripe/model/Account.java
mickjermsurawong-stripe: ah okay it will just follow OpenAPI spec doc. When the avedon come I suspect it will be removed into business profile object right?

remi-stripe: Correct! Just include it. I was lazy and did not do it because the hope is that no one really needs it and since autogen ships soon (and .net/go have the new API version next week) it wasn't worth the chunder of code + PR + release :p

mickjermsurawong-stripe:
thank you @remi-stripe!
The only action to take is then to undocument `use_stripe_sdk`!

PR #673, opened 2019-02-09
Add support for Subscription Schedule APIs

remi-stripe:
Add support for Subscription Schedule APIs

* Add Subscription Schedule
* Add Subscription Schedule Revisions
* Add `invoice_settings` on the Customer

cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/SubscriptionSchedule.java
+++ b/src/main/java/com/stripe/model/SubscriptionSchedule.java
@@ -1,0 +1,486 @@
+package com.stripe.model;
+
+import com.google.gson.annotations.SerializedName;
+
+import com.stripe.exception.StripeException;
+import com.stripe.net.ApiResource;
+import com.stripe.net.RequestOptions;
+
+import java.math.BigDecimal;
+
+import java.util.List;
+import java.util.Map;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class SubscriptionSchedule extends ApiResource
+    implements MetadataStore<SubscriptionSchedule>, HasId {
+  /**
+   * Either `charge_automatically`, or `send_invoice`. When charging automatically, Stripe will
+   * attempt to pay the underlying subscription at the end of each billing cycle using the default
+   * source attached to the customer. When sending an invoice, Stripe will email your customer an
+   * invoice with payment instructions.
+   */
+  @SerializedName("billing")
+  String billing;
+
+  /**
+   * Define thresholds at which an invoice will be sent, and the subscription advanced to a new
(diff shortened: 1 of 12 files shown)


remi-stripe:
This is now ready to be reviewed. r? @mickjermsurawong-stripe

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -0,0 +1,459 @@
...
+  }
+
+  public void setSubscriptionObject(Subscription expandableObject) {
+    this.subscription =
+        new ExpandableField<Subscription>(expandableObject.getId(), expandableObject);
+  }
+
+  // <editor-fold desc="cancel">
+  /**
+   * Pay a subscription schedule.
```
mickjermsurawong-stripe: Should be cancel a subscription schedule?

remi-stripe: damn copy/paste and replace. Fixed all 8 of them.

src/main/java/com/stripe/model/SubscriptionScheduleRevisionCollection.java
```diff
@@ -0,0 +1,56 @@
+package com.stripe.model;
+
+import com.stripe.Stripe;
+import com.stripe.exception.StripeException;
+import com.stripe.net.ApiResource;
+import com.stripe.net.RequestOptions;
+import java.util.Map;
+
+public class SubscriptionScheduleRevisionCollection
```
mickjermsurawong-stripe: Hmm.. actually no other resource returns this collection, so we can never get an instance of this collection to make the API request (with the existing url)..

mickjermsurawong-stripe: I think we can go with the followings..
In `SubscriptionSchedule.java` we have
`public SubscriptionScheduleRevisionCollection revisions() `
Then, we remove the `list` method from the `SubscriptionScheduleRevisionCollection.java`..
I fixed my client metadata and the generated outcome looks like this..
https://github.com/stripe/stripe-java/blob/abaada5f48449f495886efdda0eed0c5583bbe0b/src/main/java/com/stripe/model/SubscriptionSchedule.java#L338-L356


I believe this is similar to account/person relationship

mickjermsurawong-stripe:
ptal @remi-stripe just on the revision collection because it is currently not returned anywhere..

src/main/java/com/stripe/model/SubscriptionScheduleRevisionCollection.java
remi-stripe: @mickjermsurawong-stripe Nice catch, fixed!

mickjermsurawong-stripe:
thank @remi-stripe 
LGTM

PR #677, opened 2019-02-12
Add support for `transfer_data[amount]` on Charge

remi-stripe:
We decided to support `transfer_data[amount]` on Charges for 2/19, at least for a few months. We think developers should use `application_fee_amount` instead but many integrations already used `destination[amount]` and would be quite blocked if we removed this.

r? @mickjermsurawong-stripe  
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Charge.java
+++ b/src/main/java/com/stripe/model/Charge.java
@@ -606,6 +606,7 @@ public class Charge extends ApiResource implements MetadataStore<Charge>, HasId 
   @Setter
   @EqualsAndHashCode(callSuper = false)
   public static class TransferData extends StripeObject {
+    Long amount;
     @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Account> destination;
 
     // <editor-fold desc="destination">


src/main/java/com/stripe/model/Charge.java
```diff
@@ -606,6 +606,7 @@ public void setRuleObject(Rule rule) {
   @Setter
   @EqualsAndHashCode(callSuper = false)
   public static class TransferData extends StripeObject {
+    Long amount;
```
mickjermsurawong-stripe: Should we mark this with @Deprecated or only for payment-intent beta user only?
It will be removed in 8.0x with version pin.

remi-stripe: @mickjermsurawong-stripe No it won't be removed this is not limited by an API version for Charges and we will actively support it for a long time.

mickjermsurawong-stripe: or add comments to use `application_fee_amount` instead

mickjermsurawong-stripe:
@remi-stripe, just a note on the documentation for this legacy field
LGTM

Issue #605, opened 2018-10-16
Arrays in webhooks previous attributes parsed as Object[]


Flo354:
Hi,

I am using stripe-java version 7.1.0 and I am currently experiencing something weird with Webhooks.
When getting `event.getData().getPreviousAttributes()`, I obtain a `Map<String, Object>` which is fine.

Within this map, if I am getting the field `verification`, is it also a `Map<String, Object>`.
But why an array, like `verification -> fields_needed` is an `Object[]` instead of a `List<Object>`.

Actually, I am asking this question to myself, because with my JSON library (Jackson JSON), When I am using this code: 

```java
String jsonStr = "some json";
Map<String, Object> result = mapper.readValue(jsonStr, Map.class);
``` 

When I am recovering  `verification -> fields_needed`, I am getting an `ArrayList<Object>`. And I find this more natural.

ob-stripe:
Hi @Flo354, thanks for the report!

Stripe's Java library uses a custom deserializer for the `data` attribute of event objects. It is this custom deserializer that decodes JSON lists as Java arrays instead of `List` objects (cf. [here](https://github.com/stripe/stripe-java/blob/ebbc8bedb88e369d2103e67eb083e8bd42027152/src/main/java/com/stripe/model/EventDataDeserializer.java#L88-L97)).

I agree that `List` objets would feel more natural here and be more consistent with the rest of the API, so I'd be inclined to change the behavior here. Such a change would break backwards compatibility, but I think we were planning on releasing a new major version soon anyway, so we could bundle this change too.

@brandur-stripe @remi-stripe Any strong feels around this?

remi-stripe:
I agree that List would make a lot more sense and I would tag as Future for when we do the next major revision.

Flo354:
Thanks!

I'll wait for the next major release.

ob-stripe:
Quick update: this didn't make it into the newly released major version 8.0.0, but it's still on our roadmap and will be fixed in a future major version.

ob-stripe:
Fixed in stripe-java 9.0.0.

PR #679, opened 2019-02-19
Basic primitives for typed params 

mickjermsurawong-stripe:
- This PR adds support for autogen typed params. The usage examples of generated code in https://github.com/stripe/stripe-java/pull/680
- It adds abstract class `ApiRequestParams` to be extended by top-level params, and an inner interface `ApiRequestParams.Enum`. df04964
  - `ApiRequestParams` exposes to `toMap` making it compatible with the form-encoding logic. This map is obtained from GSON using the serialized name annotation generated. 
  - `ApiRequestParams.Enum` has getter for its raw value, this is used in custom serializer to convert `EMPTY` enum into null value in the param map. The `EMPTY` enum (used to unset values in some API) is equivalent to empty string "" in OpenAPI spec, but our [current implementation](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/net/LiveStripeResponseGetter.java#L365-L368) requires it to be explicit `null`
```java
public class FooParams extends ApiRequestParams {
  String fooString;
  FooEnum fooEnum;

  public FooEnum implements ApiRequestParams.Enum {
    @SerializedName("my_foo")
    MY_FOO("my_foo"),
    @SerializedName("")
    EMPTY("my_foo");
    // implemented method required by enum
    @Getter private final String value
  }
}
```
- This PR also does refactoring in extracting common logic from converting GSON to map that is used in `EventDataDeserializer`.  6c020c8.
- The final logic added here is to add comparison logic in the `ParamMapMatcher` in testing. This allows more comparison of untyped Map that is robust to difference in `Object[]` vs `List<Object`, and `Long` vs `Integer` f22e048
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/ApiRequestParams.java
+++ b/src/main/java/com/stripe/net/ApiRequestParams.java
@@ -9,6 +9,7 @@ import com.google.gson.TypeAdapterFactory;
 import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
+
 import java.io.IOException;
 import java.util.Map;
 
--- a/src/main/java/com/stripe/net/ApiRequestParams.java
+++ b/src/main/java/com/stripe/net/ApiRequestParams.java
@@ -12,8 +12,18 @@ import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
 import java.util.Map;
 
+/**
+ * Super class to all api request params objects. This common abstraction is internally used in
+ * {@link ApiResource#request(ApiResource.RequestMethod, String, ApiRequestParams, Class,
+ * RequestOptions)}. It also exposes a convenient method converting the typed parameter into
+ * the legacy support of untyped {@code Map<String, Object>} param.
+ */
 public abstract class ApiRequestParams {
-
+  /**
+   * Interface implemented by all enum parameter to get the actual string value that Stripe API
+   * expects. Internally, it used in custom serialization
+   * {@link ApiRequestParams.HasEmptyEnumTypeAdapterFactory} converting empty string enum to null.
+   */
   public interface Enum {
     String getValue();
   }
@@ -59,6 +69,20 @@ public abstract class ApiRequestParams {
     }
   }
 
+  /**
+   * Convenient method to convert this typed request params into an untyped map. This map is
+   * composed of {@code Map<String, Object>}, {@code List<Object>}, and basic Java data types.
+   * This allows you to test building the request params and verify compatibility with your
+   * prior integrations using the untyped params map
+   * {@link ApiResource#request(ApiResource.RequestMethod, String, Map, Class, RequestOptions)}.
+   *
+   * <p>The peculiarity of this conversion is that `EMPTY` {@link ApiRequestParams.Enum} with raw
+   * value of empty string will be converted to null. This is compatible with the existing
+   * contract enforcing no empty string in the untyped map params.
+   *
+   * <p>Because of the translation from `EMPTY` enum to null, deserializing this map back to a
+   * request instance is lossy. The null value will not be converted back to the `EMPTY` enum.
+   */
   public Map<String, Object> toMap() {
     JsonObject json = GSON.toJsonTree(this).getAsJsonObject();
     return UNTYPED_MAP_DESERIALIZER.deserialize(json);
--- a/src/main/java/com/stripe/net/ApiRequestParams.java
+++ b/src/main/java/com/stripe/net/ApiRequestParams.java
@@ -1,0 +1,67 @@
+package com.stripe.net;
+
+import com.google.gson.FieldNamingPolicy;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonObject;
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+import java.util.Map;
+
+public abstract class ApiRequestParams {
+
+  public interface Enum {
+    String getValue();
+  }
+
+  private static final Gson GSON = new GsonBuilder()
+      .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+      .registerTypeAdapterFactory(new HasEmptyEnumTypeAdapterFactory())
+      .create();
+
+  private static final UntypedMapDeserializer UNTYPED_MAP_DESERIALIZER =
+      new UntypedMapDeserializer();
+
+  private static class HasEmptyEnumTypeAdapterFactory implements TypeAdapterFactory {
+    @SuppressWarnings("unchecked")
+    @Override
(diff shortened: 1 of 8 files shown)


src/main/java/com/stripe/net/ApiRequestParams.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private static final Gson GSON = new GsonBuilder()
+      .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+      .registerTypeAdapterFactory(new HasEmptyEnumTypeAdapterFactory())
+      .create();
+
+  private static final UntypedMapDeserializer UNTYPED_MAP_DESERIALIZER =
+      new UntypedMapDeserializer();
+
+  private static class HasEmptyEnumTypeAdapterFactory implements TypeAdapterFactory {
```
mickjermsurawong-stripe: Main logic of handling serializing EMPTY enum to null

src/main/java/com/stripe/net/ApiResource.java
```diff
@@ -163,13 +163,25 @@ public static String urlEncode(String str) throws UnsupportedEncodingException {
...
                 String url, Map<String, Object> params, Class<T> clazz,
                 RequestOptions options) throws StripeException {
     return ApiResource.stripeResponseGetter.request(method, url, params, clazz,
         ApiResource.RequestType.NORMAL, options);
   }
 
+  public static <T extends StripeCollectionInterface<?>> T requestCollection(
+      String url, ApiRequestParams params, Class<T> clazz, RequestOptions options)
+      throws StripeException {
+    return requestCollection(url, params.toMap(), clazz, options);
```
mickjermsurawong-stripe: Change in `ApiResource` request invocation to support `ApiRequestParams`

src/main/java/com/stripe/net/UntypedMapDeserializer.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+
+public class UntypedMapDeserializer {
```
mickjermsurawong-stripe: Old logic ported from `EventDataDeserializer`. New tests are added to `UntypedMapDeserializerTest`

src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -506,18 +508,77 @@ public ParamMapMatcher(Map<String, Object> other) {
...
       }
     }
+
+    @Override
+    public String toString() {
+      return other.toString();
+    }
+  }
+
+  private static Boolean compareParamObjects(Object thisValue, Object otherValue) {
```
mickjermsurawong-stripe: Additional logic to make testing of typed param and comparison with old untyped map easier.
In the subsequent PR, I rely on the existing tests that has untyped params. I want to make sure that using typed params will have the same effect.

src/test/java/com/stripe/functional/StripeResponseTest.java
```diff
@@ -25,7 +26,7 @@ public void testResponseIncluded() throws StripeException {
     final RequestOptions requestOptions = RequestOptions.builder()
         .setIdempotencyKey(idempotencyKey)
         .build();
-    final Customer customer = Customer.create(null, requestOptions);
+    final Customer customer = Customer.create((Map<String, Object>) null, requestOptions);
```
mickjermsurawong-stripe: Compiler complains that null now can imply both null for untyped param Map or typed `ApiRequestParams`. Existing integration with null map will likely break loudly with complier checks.

src/main/java/com/stripe/net/UntypedMapDeserializer.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Deserialize JSON into untyped map.
+   * {@code JsonArray} is represented as {@code Object[]}.
```
ob-stripe: I'd prefer that we use lists instead of arrays everywhere in the library. `List`s are (IMHO) more flexible and easier to reason about, and make the code easier to maintain. See also #605.

mickjermsurawong-stripe:
thank youu @ob-stripe for the reviews so far on this `stripe-java`!
I've updated the untyped deserializer to use List, and the comments in `EventData` accordingly.
Will add this change to the migration guide as it is not obvious from the API and we might silently break merchant.

mickjermsurawong-stripe:
ptal @ob-stripe

src/main/java/com/stripe/net/ApiRequestParams.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.gson.JsonObject;
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+import java.util.Map;
+
+public abstract class ApiRequestParams {
```
brandur-stripe: Mind giving this a docblock explaining what it's for?

src/main/java/com/stripe/net/UntypedMapDeserializer.java
brandur-stripe: And similarly here, can we do a docblock?

src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -21,9 +21,11 @@
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import java.util.Set;
```
brandur-stripe: Nit, but can you tighten up the spacing here? I assume `java.util.Set` belongs with the group above instead of with `lombok.*`.

mickjermsurawong-stripe:
thank you @brandur-stripe and @ob-stripe for the review!
I've updated the java doc.
ptal @brandur-stripe

src/test/java/com/stripe/BaseStripeTest.java
brandur-stripe: Tiny follow along nit: I think what the original author was trying to do here was separate out different "groups" of imports. So all the `java` would land together, and then you'd have `com.*`, `org.*`, etc. in their own sections. If that's the case, I think the convention would be a double space between `java.*` and `lombok.*`.

brandur-stripe:
Thanks @mickjermsurawong-stripe! Added one more tiny comment, but LGTM.

src/test/java/com/stripe/BaseStripeTest.java
mickjermsurawong-stripe: ah sorry @brandur-stripe: I simply used auto-formatting and assumed that was the standard one. I now fixed the import statement in my intellij setting to follow the order of the grouping here. With that though I didn't add double space between `java.*` and `lombok.*` as it the current one doesn't.

mickjermsurawong-stripe:
Thank you @brandur-stripe! I fixed the import order in other files I touched here too.

mickjermsurawong-stripe:
thank you! merging this to beta branch

PR #681, opened 2019-02-22
Add `status_transitions` and mark `date` as deprecated on `Invoice`

remi-stripe:
Apparently we've always had `created` on `Invoice` even though it did not exist!

r? @mickjermsurawong-stripe 
cc @stripe/api-libraries


mickjermsurawong-stripe:
thank you @remi-stripe 
LGTM

PR #683, opened 2019-02-26
Gen typed params and latest resources: autogen java spec a5eee30

mickjermsurawong-stripe:
This is the latest updates that introduced typed params and new resources.
Jar from this is also used to run codes to verify api sample refs are up-to-date
cc @stripe/api-libraries


src/main/java/com/stripe/param/AccountCreateParams.java
```diff
@@ -0,0 +1,2642 @@
...
+   */
+  @SerializedName("account_token")
+  String accountToken;
+
+  /**
+   * Either a token, like the ones returned by [Stripe.js](https://stripe.com/docs/stripe.js), or a
+   * dictionary containing a user's bank account details.
+   */
+  @SerializedName("bank_account")
+  Object bankAccount;
```
remi-stripe: Why is this one an Object when `Company` is a class. I'm assuming it's because it has different shapes but what does the code using it looks like?

Also, I'm realizing that `bank_account` should not be supported at all. Only `external_account` should be supported. `bank_account` was removed in 2015 in theory: https://stripe.com/docs/upgrades#2015-10-01

src/main/java/com/stripe/param/AccountCreateParams.java
```diff
@@ -0,0 +1,2642 @@
...
+   * The email address of the account holder. For Standard accounts, Stripe will email your user
+   * with instructions on how to set up their account. For Custom accounts, this is only to make the
+   * account easier to identify to you: Stripe will never directly email your users.
+   */
+  @SerializedName("email")
+  String email;
+
+  /** Specifies which fields in the response should be expanded. */
+  @SerializedName("expand")
+  List<String> expand;
```
remi-stripe: Can you confirm whether we're always going to have this property on all params even if nothing is expandable?

src/main/java/com/stripe/param/AccountCreateParams.java
```diff
@@ -0,0 +1,2642 @@
...
+  @SerializedName("metadata")
+  Map<String, String> metadata;
+
+  /**
+   * The set of capabilities you want to unlock for this account. Each capability will be inactive
+   * until you have provided its specific requirements and Stripe has verified them. An account may
+   * have some of its requested capabilities be active and some be inactive.
+   */
+  @SerializedName("requested_capabilities")
+  List<RequestedCapability> requestedCapabilities;
```
remi-stripe: I did not know we were going with enums in params. Is that expected? While I'm fine in principle, I would say that enum on resources are higher priority than on params (personal opinion). This is a bad example as it's not returned in the API but I've confirmed that `disabled_reason` for example on the `Account` is not an enum.

src/main/java/com/stripe/param/AccountCreateParams.java
```diff
@@ -0,0 +1,2642 @@
...
+    @SerializedName("address_kana")
+    AddressKana addressKana;
+
+    /** The Kanji variation of the company's primary address (Japan only). */
+    @SerializedName("address_kanji")
+    AddressKanji addressKanji;
+
+    /** Whether information has been collected from the company's directors. */
+    @SerializedName("directors_provided")
+    Boolean directorsProvided;
```
remi-stripe: Random but wanted to make sure we won't have the same issue as we had in go for years where this sends `false` if it's not initialized. We had to move to pointers and that was a lot of work.

src/main/java/com/stripe/param/AccountCreateParams.java
```diff
@@ -0,0 +1,2642 @@
...
+    public Builder setBankAccount(String bankAccount) {
+      this.bankAccount = bankAccount;
+      return this;
+    }
+
+    /**
+     * Either a token, like the ones returned by [Stripe.js](https://stripe.com/docs/stripe.js), or
+     * a dictionary containing a user's bank account details.
+     */
+    public Builder setBankAccount(Map<String, Object> bankAccount) {
```
mickjermsurawong-stripe: @remi-stripe when a field is polymorphic, the builder will have explicit setters for each type. So using the builder, one will have this effect of `setBankAccount` showing explicit options for all the possible types.
image

The getter of the built object though will have return type Object that users have to cast, but that's not the expected use case.

src/main/java/com/stripe/param/AccountCreateParams.java
mickjermsurawong-stripe: Yup this is expected. Enum on resource introduced now will cause too much incompatibility with 7.x. But here since we start from zero, we introduce Enum as well.

src/main/java/com/stripe/param/CustomerCreateParams.java
```diff
@@ -0,0 +1,611 @@
...
+      this.shipping = shipping;
+      return this;
+    }
+
+    public Builder setSource(String source) {
+      this.source = source;
+      return this;
+    }
+
+    public Builder setSource(Map<String, Object> source) {
```
remi-stripe: I don't think we should support a map here. In theory, we should go all the way and support passing the correct list of params that are valid for the `source` hash like card's number, cvc, etc.
If we do, there's a big documentation risk as we need to make it clear no one should use this.

src/main/java/com/stripe/param/AccountCreateParams.java
mickjermsurawong-stripe: ahh thank you remi! no. we should not expect this. I will have to fix this in OpenAPI spec.

src/main/java/com/stripe/param/AccountCreateParams.java
remi-stripe: Ack okay. Does Java let you pass a string there easily instead of an enum value? There are a lot of properties that have gated enum values

src/main/java/com/stripe/param/AccountCreateParams.java
remi-stripe: Ack okay. I got confused because I expected the String or Class of params for bank accounts like @ob-stripe did in .net with AnyOf

src/main/java/com/stripe/param/SubscriptionCreateParams.java
```diff
@@ -0,0 +1,758 @@
...
+  Billing billing;
+
+  /**
+   * A future timestamp to anchor the subscription's [billing
+   * cycle](https://stripe.com/docs/subscriptions/billing-cycle). This is used to determine the date
+   * of the first full invoice, and, for plans with `month` or `year` intervals, the day of the
+   * month for subsequent invoices.
+   */
+  @SerializedName("billing_cycle_anchor")
+  String billingCycleAnchor;
```
remi-stripe: This should be `long` right? since it's a unix timestamp

src/main/java/com/stripe/param/SubscriptionCreateParams.java
```diff
@@ -0,0 +1,758 @@
...
+
+  /**
+   * Unix timestamp representing the end of the trial period the customer will get before being
+   * charged for the first time. This will always overwrite any trials that might apply via a
+   * subscribed plan. If set, trial_end will override the default trial period of the plan the
+   * customer is being subscribed to. The special value `now` can be provided to end the customer's
+   * trial immediately.
+   */
+  @SerializedName("trial_end")
+  String trialEnd;
```
remi-stripe: This should be a string for `now` or a `long` for unix timestamp. Would it be cleaner to handle this the right way?

src/main/java/com/stripe/param/CustomerCreateParams.java
```diff
@@ -0,0 +1,611 @@
...
+
+    private String id;
+
+    private String invoicePrefix;
+
+    private InvoiceSettings invoiceSettings;
+
+    private Map<String, String> metadata;
+
+    private Object shipping;
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: Do you mean the `EMPTY` enum for `setShipping`?

src/main/java/com/stripe/param/SubscriptionCreateParams.java
```diff
@@ -0,0 +1,758 @@
...
+  @SerializedName("items")
+  List<Item> items;
+
+  /**
+   * A set of key-value pairs that you can attach to a `Subscription` object. It can be useful for
+   * storing additional information about the subscription in a structured format.
+   */
+  @SerializedName("metadata")
+  Map<String, String> metadata;
+
```
remi-stripe: We're missing `plan` though I guess it's on purpose because we consider it deprecated and we're fine forcing merchants to not support it? (it sounds okay especially as there's always the escape hatch anyway, just asking to confirm)

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+import java.util.Map;
+import lombok.Getter;
+
+@Getter
+public class TokenCreateParams extends ApiRequestParams {
+  @SerializedName("account")
+  Account account;
+
+  @SerializedName("amount")
+  Long amount;
```
remi-stripe: This should not be in this API. Please make sure to remove/document in pay-server so that it can't leak again. This is a really old and sensitive parameter that lets you do card testing really easily :(

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+   * with [Stripe Connect](https://stripe.com/docs/connect). Also, this can be used only with an
+   * [OAuth access token](https://stripe.com/docs/connect/standard-accounts) or [Stripe-Account
+   * header](https://stripe.com/docs/connect/authentication). For more details, see [Shared
+   * Customers](https://stripe.com/docs/connect/shared-customers).
+   */
+  @SerializedName("customer")
+  String customer;
+
+  @SerializedName("email")
+  String email;
```
remi-stripe: AFAIK this is useful only for Legacy Checkout in go. This should not be in the library

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  /** Specifies which fields in the response should be expanded. */
+  @SerializedName("expand")
+  List<String> expand;
+
+  @SerializedName("external_id")
+  String externalId;
```
remi-stripe: This is internal/sensitive field used by our SDKs like Stripe.js and should not leak in the library

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("iovation_blackbox")
+  String iovationBlackbox;
```
remi-stripe: This is internal/sensitive field used by our SDKs like Stripe.js and should not leak in the library

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("ip")
+  String ip;
```
remi-stripe: This is internal/sensitive field used by our SDKs like Stripe.js and should not leak in the library

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("payment_user_agent")
+  String paymentUserAgent;
```
remi-stripe: This is internal/sensitive field used by our SDKs like Stripe.js and should not leak in the library

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("pii")
+  Pii pii;
+
+  @SerializedName("recipient")
+  String recipient;
```
remi-stripe: We likely should remove recipient from this, it's been deprecated for years

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("referrer")
+  String referrer;
```
remi-stripe: This is internal/sensitive field used by our SDKs like Stripe.js and should not leak in the library

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("request_id")
+  String requestId;
```
remi-stripe: I don't understand where that one is coming from but it should not be a parameter at all.

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("user_agent")
+  String userAgent;
```
remi-stripe: This is internal/sensitive field used by our SDKs like Stripe.js and should not leak in the library

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2527 @@
...
+
+  @SerializedName("validation_type")
+  ValidationType validationType;
```
remi-stripe: Same as `amount`, can be used for card-testing, we need to explicitly hide this.

src/main/java/com/stripe/param/AccountCreateParams.java
remi-stripe: It is likely easier to just have it always. Though I'm fine with not having it.

src/main/java/com/stripe/param/CustomerCreateParams.java
mickjermsurawong-stripe: Ah we don't have this encoded in OpenAPI spec for now.. Let me see if I can fix it upstream..

src/main/java/com/stripe/param/CustomerCreateParams.java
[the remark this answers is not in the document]
remi-stripe: Yes. This feels quite weird and I noticed it in other places. I guess we might not have a choice. Just looks quite hacky/wrong to me so wanted to raise.

src/main/java/com/stripe/param/CustomerCreateParams.java
remi-stripe: Yeah this is one of those where I'm ambivalent. I want us to support it properly. But then more and more merchants will send raw PANs because it looks easy in the library. Not sure what the right product decision is here. Maybe the hash is easier because less people would find it.

src/main/java/com/stripe/param/AccountCreateParams.java
mickjermsurawong-stripe: We have `.toMap` exposed to users so they can always add gated/beta content the Map, and use the old way of calling API

mickjermsurawong-stripe:
Thanks Remi!

On tokens there those internal fields, I will have to undoc them from public spec.

On the deprecated fields that are introduced, like `bank_account` on Account create, and `customer/id` on Customer create, they are the "forever_changes" version change; we remove from the public doc, but continue supporting. In OpenAPI if I understand correctly, those are now included for param log redaction, and also for internal testing. So they should be removed from SDK spec only.

So I think we likely want to ignore this forever change in SDK spec, and follow the latest params instead similar to that in API refs instead. That's a little tricky because the API refs also does lots of custom handling for these fields..

src/main/java/com/stripe/param/AccountCreateParams.java
mickjermsurawong-stripe: This is removed and now we have `externalAccount` String token instead.
The documentation is still wrong though. I have a PR for gated docs that doesn't change that in API refs (saying that it's still a hash)

src/main/java/com/stripe/param/AccountCreateParams.java
mickjermsurawong-stripe: Yup we currently have it as default except request with `DELETE` endpoint

src/main/java/com/stripe/param/AccountCreateParams.java
mickjermsurawong-stripe: Added test to verify this f8ede6b

src/main/java/com/stripe/param/CardUpdateOnCustomerParams.java
```diff
@@ -0,0 +1,448 @@
...
+import com.stripe.net.ApiRequestParams;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Getter;
+
+@Getter
+public class CardUpdateOnCustomerParams extends ApiRequestParams {
+  /** The name of the person or business that owns the bank account. */
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: Updated. Card params now have only card-related values

src/main/java/com/stripe/param/CardUpdateOnCustomerParams.java
```diff
@@ -0,0 +1,448 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.stripe.net.ApiRequestParams;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Getter;
+
+@Getter
+public class CardUpdateOnCustomerParams extends ApiRequestParams {
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: I suspect the pattern of usage is from the method call where they look at method signature, instead of going through our list of params. 
Do you think naming like `CustomerCardUpdate` and `AccountCardUpdate` would be better? We can do `CustomerCardUpdate` and `PayoutCardUpdate` but that will be more confusing.

src/main/java/com/stripe/param/CardUpdateOnCustomerParams.java
```diff
@@ -0,0 +1,448 @@
...
+  /** The name of the person or business that owns the bank account. */
+  @SerializedName("account_holder_name")
+  String accountHolderName;
+
+  /** The type of entity that holds the account. This can be either `individual` or `company`. */
+  @SerializedName("account_holder_type")
+  AccountHolderType accountHolderType;
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: Addressed

src/main/java/com/stripe/param/CustomerCreateParams.java
```diff
@@ -0,0 +1,611 @@
...
+   * invoice; a positive amount increases the amount due on an invoice.
+   */
+  @SerializedName("account_balance")
+  Long accountBalance;
+
+  @SerializedName("coupon")
+  String coupon;
+
+  @SerializedName("customer")
+  String customer;
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: Addressed

src/main/java/com/stripe/param/CustomerCreateParams.java
```diff
@@ -0,0 +1,611 @@
...
+  @SerializedName("email")
+  String email;
+
+  /** Specifies which fields in the response should be expanded. */
+  @SerializedName("expand")
+  List<String> expand;
+
+  /** A custom ID to use for the customer. */
+  @SerializedName("id")
+  String id;
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: It now goes away together with `customer`

src/main/java/com/stripe/param/CustomerCreateParams.java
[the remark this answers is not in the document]
mickjermsurawong-stripe: We actually discussed this during the implementation phase, and agreed on `EMPTY` 
https://paper.dropbox.com/doc/Polymorphic-params-in-stripe-java--AZOgnte1Z7stWBn8JIdFsIe4Ag-amkaRe4C2QWqjVWcONtWx

src/main/java/com/stripe/param/CustomerCreateParams.java
mickjermsurawong-stripe: As discussed, it's a now a String

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: This and the token fields below are addressed

mickjermsurawong-stripe:
@remi-stripe 
I've addressed all the comments except one on typing of `billingCycleAnchor`, `trialEnd` in `SubscriptionCreateParams.java`. I'll look at how to resolve that.

Could you help to look at a few more critical params please?

mickjermsurawong-stripe:
ptal @remi-stripe

src/main/java/com/stripe/param/AccountCreateParams.java
```diff
@@ -0,0 +1,2587 @@
...
+   * By default, providing an external account sets it as the new default external account for its
+   * currency, and deletes the old default if one exists. To add additional external accounts
+   * without replacing the existing default for the currency, use the bank account or card creation
+   * API.
+   */
+  @SerializedName("external_account")
+  String externalAccount;
+
+  @SerializedName("from_recipient")
+  String fromRecipient;
```
remi-stripe: This has been undocumented for a while, why is it appearing here when others undocumented things don't appear?

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2512 @@
...
+      this.card = card;
+      return this;
+    }
+
+    public Builder setCard(CreditCard card) {
+      this.card = card;
+      return this;
+    }
+
+    public Builder setCard(EncryptedCard card) {
```
remi-stripe: This is secret and gated and I don't think it should be here

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2512 @@
...
+      return this;
+    }
+
+    /** The bank account this token will represent. */
+    public Builder setBankAccount(BankAccount bankAccount) {
+      this.bankAccount = bankAccount;
+      return this;
+    }
+
+    public Builder setCard(ApplePay card) {
```
remi-stripe: I don't think this should be in stripe-java. If it's what I think it's about creating a card token from an Apple Pay PKToken and should only be in stripe-ios

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2512 @@
...
+      this.card = card;
+      return this;
+    }
+
+    public Builder setCard(CreditCard card) {
```
remi-stripe: CreditCard is a risky name for a class like this. It implies Debit cards would not work

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2512 @@
...
+      this.card = card;
+      return this;
+    }
+
+    public Builder setCard(EncryptedCard card) {
+      this.card = card;
+      return this;
+    }
+
+    public Builder setCard(SwipeData card) {
```
remi-stripe: I don't think we should support that publicly.

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2512 @@
...
+    String cardType;
+
+    @SerializedName("currency")
+    String currency;
+
+    @SerializedName("customer")
+    String customer;
+
+    @SerializedName("default_for_currency")
+    Boolean defaultForCurrency;
```
remi-stripe: I don't think this is a real thing on token creation so I don't get why it's here

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2512 @@
...
+    Map<String, String> metadata;
+
+    @SerializedName("name")
+    String name;
+
+    @SerializedName("number")
+    String number;
+
+    @SerializedName("number_looks_valid")
+    Boolean numberLooksValid;
```
remi-stripe: same, this looks like really leaking something

src/main/java/com/stripe/param/TokenCreateParams.java
```diff
@@ -0,0 +1,2512 @@
...
+      }
+
+      /**
+       * Add a key/value pair to `metadata` map. A map is initialized for the first `put/putAll`
+       * call, and subsequent calls add additional key/value pairs to the original map. See {@link
+       * CreditCard#metadata} for the field documentation.
+       */
+      public Builder putMetadata(String key, String value) {
+        if (this.metadata == null) {
+          t
```
remi-stripe: This should not be here, there's no 3DS class on token creation and if there is it's some weird internal thing I would say.

src/main/java/com/stripe/param/SourceCreateParams.java
```diff
@@ -0,0 +1,1105 @@
...
+      @Getter private final String value;
+
+      Interval(String value) {
+        this.value = value;
+      }
+    }
+
+    public enum NotificationMethod implements ApiRequestParams.Enum {
+      @SerializedName("deprecated_none")
+      DEPRECATED_NONE("deprecated_none"),
```
remi-stripe: Can you clarify what this one even is?

src/main/java/com/stripe/param/SourceCreateParams.java
```diff
@@ -0,0 +1,1105 @@
...
+
+      @SerializedName("email")
+      EMAIL("email"),
```
remi-stripe: What is that `EMAIL()` thing

src/main/java/com/stripe/param/SubscriptionCreateParams.java
mickjermsurawong-stripe: @remi-stripe, I realized I haven't fixed this type inconsistencies

src/main/java/com/stripe/param/SubscriptionCreateParams.java
mickjermsurawong-stripe: below here too

src/main/java/com/stripe/param/SubscriptionCreateParams.java
mickjermsurawong-stripe: @remi-stripe I still haven't addressed this "missing" `plan` field and also Customer create too.
Are we okay with releasing without first?

src/main/java/com/stripe/param/AccountCreateParams.java
mickjermsurawong-stripe: addressed in https://github.com/stripe/stripe-java/pull/711/commits/760b5b5713915de21ce7368dace7e88e9c5dc85f#diff-a9f0113cf10a4a96bc61507995cc5a57R62

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: Addressed

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: Addressed

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: Addressed

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: Removed 3d secured

src/main/java/com/stripe/param/SourceCreateParams.java
mickjermsurawong-stripe: This goes away when we decide to undoc mandate

src/main/java/com/stripe/param/SourceCreateParams.java
mickjermsurawong-stripe: Same because of undoc mandate

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: addressed

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: addressed

src/main/java/com/stripe/param/TokenCreateParams.java
mickjermsurawong-stripe: Addressed

PR #686, opened 2019-03-07
Add support for `cancel_at` on `Subscription`

remi-stripe:
Add support for `cancel_at` on `Subscription`. This is gated but many users have access to it and I missed it while releasing `SubscriptionSchedule`.

r? @mickjermsurawong-stripe 
@stripe/api-libraries


mickjermsurawong-stripe:
thanks @brandur!

PR #688, opened 2019-03-13
Adding support for PaymentMethod

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Charge.java
+++ b/src/main/java/com/stripe/model/Charge.java
@@ -1,5 +1,7 @@
 package com.stripe.model;
 
+import com.google.gson.annotations.SerializedName;
+
 import com.stripe.exception.StripeException;
 import com.stripe.net.ApiResource;
 import com.stripe.net.RequestOptions;
@@ -30,6 +32,7 @@ public class Charge extends ApiResource implements MetadataStore<Charge>, HasId 
   AlternateStatementDescriptors alternateStatementDescriptors;
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE)
       ExpandableField<BalanceTransaction> balanceTransaction;
+  PaymentMethod.BillingDetails billingDetails;
   Boolean captured;
   Long created;
   String currency;
@@ -48,6 +51,8 @@ public class Charge extends ApiResource implements MetadataStore<Charge>, HasId 
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE) ExpandableField<Order> order;
   Outcome outcome;
   Boolean paid;
+  String paymentMethod;
+  PaymentMethodDetails paymentMethodDetails;
   String receiptEmail;
   String receiptNumber;
   String receiptUrl;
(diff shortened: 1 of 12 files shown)


src/main/java/com/stripe/model/Charge.java
```diff
@@ -602,6 +606,473 @@ public void setRuleObject(Rule rule) {
...
+      @SerializedName("swift_code")
+      String swiftCode;
+    }
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class AchDebit extends StripeObject {
+      @SerializedName("account_holder_type")
+      String accountNumber;
```
mickjermsurawong-stripe: Should be `accountHolderType`?

src/main/java/com/stripe/model/Charge.java
```diff
@@ -602,6 +606,473 @@ public void setRuleObject(Rule rule) {
...
+      String preferredLanguage;
+
+      @SerializedName("verified_name")
+      String verifiedName;
+    }
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class Bitcoin extends StripeObject {
```
mickjermsurawong-stripe: Schema is quite different from what autogen has.
```
    public static class Bitcoin extends StripeObject {
      @SerializedName("address")
      String address;

      @SerializedName("amount")
      Long amount;

      @SerializedName("amount_charged")
      Long amountCharged;

      @SerializedName("amount_received")
      Long amountReceived;

      @SerializedName("amount_returned")
      Long amountReturned;

      @SerializedName("refund_address")
      String refundAddress;
    }
```

src/main/java/com/stripe/model/Charge.java
```diff
@@ -602,6 +606,473 @@ public void setRuleObject(Rule rule) {
...
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class Card extends StripeObject {
+      @SerializedName("brand")
+      String brand;
+
+      @SerializedName("checks")
+      PaymentMethod.Card.Checks checks;
```
mickjermsurawong-stripe: In payflow code, there are actually two classes of checks. One in the details here, and the other one in payment method, eventhough the schemas are exactly the same
```
lib/payment_flows/private/payment_methods/card/checks/details_api_resource.rb
lib/payment_flows/private/payment_methods/card/checks/api_resource.rb 	
```

remi-stripe:
@mickjermsurawong-stripe Thanks a lot for catching the issues. I have re-generated openapi and added it to the PR so that all tests pass.

Can you do a full review on that one to confirm everything is as expected and ready to merge (beyond a new stripe-mock)?

src/main/java/com/stripe/model/PaymentMethod.java
```diff
@@ -0,0 +1,383 @@
...
+        String.format(
+            "%s%s",
+            Stripe.getApiBase(), String.format("/v1/payment_methods/%s/detach", this.getId()));
+    return request(ApiResource.RequestMethod.POST, url, params, PaymentMethod.class, options);
+  }
+
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public class BillingDetails extends StripeObject {
```
mickjermsurawong-stripe: Missing `static` here for the inner class.

src/main/java/com/stripe/model/Charge.java
```diff
@@ -602,6 +607,487 @@ public void setRuleObject(Rule rule) {
     }
   }
 
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class PaymentMethodDetails extends StripeObject {
+    @SerializedName("account")
+    Account account;
```
mickjermsurawong-stripe: I think the field name should be `stripe_account` instead?
And we can consider renaming the class below to `StripeAccount`?

mickjermsurawong-stripe:
@remi-stripe 
I did build a jar and does a diff with autogen. The two comments above were results I found.
ptal @remi-stripe

remi-stripe:
@mickjermsurawong-stripe Thanks for catching those. I fixed the 2 issues, PTAL!

mickjermsurawong-stripe:
@remi-stripe Sorry, I was only looking at `PaymentMethod`, but I think we are missing an expandable field `paymentMethod` field in `PaymentIntent` too? 
The internal site: `docs/api/payment_intents/object#payment_intent_object-payment_method`.
I think we miss that in other typed libs too.

remi-stripe:
@mickjermsurawong-stripe so glad you caught this, I just fixed it. PTAL (will fix the other libs)

src/main/java/com/stripe/model/Charge.java
```diff
@@ -30,6 +32,7 @@
   AlternateStatementDescriptors alternateStatementDescriptors;
   @Getter(AccessLevel.NONE) @Setter(AccessLevel.NONE)
       ExpandableField<BalanceTransaction> balanceTransaction;
+  PaymentMethod.BillingDetails billingDetails;
```
mickjermsurawong-stripe: Verified that the backend here does share resource with PaymentMethod

mickjermsurawong-stripe:
thank you @remi-stripe!
LGTM

mickjermsurawong-stripe:
oh I assume you will do version update etc on this PR too?

remi-stripe:
Moved to latest stripe-mock and removed old openapi spec.

PR #689, opened 2019-03-14
Add support for new properties on Reporting resources

remi-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries


src/main/java/com/stripe/model/reporting/ReportRun.java
```diff
@@ -93,6 +94,7 @@ public static ReportRun retrieve(String id, Map<String, Object> params,
   @Setter
   @EqualsAndHashCode(callSuper = false)
   public final class Parameters extends StripeObject {
+    List<String> columns;
```
mickjermsurawong-stripe: Ah i see.. It's only added today that 8.x doesn't have it.

mickjermsurawong-stripe:
LGTM

PR #690, opened 2019-03-16
[payment methods] source: spec3.sdk.yaml@spec-9d9decd in master

mickjermsurawong-stripe:
- This PR contains the generated payment method apis.
- It sadly just got the most updated Terminal endpoints that are incorrect now, but I think it's helpful if the PM features get reviewed or can help to the non-autogen one..
r? @remi-stripe 
cc @stripe/api-libraries


src/main/java/com/stripe/model/terminal/Location.java
```diff
@@ -100,4 +100,29 @@ public static LocationCollection list(Map<String, Object> params, RequestOptions
     String url = String.format("%s%s", Stripe.getApiBase(), "/v1/terminal/locations");
     return requestCollection(url, params, LocationCollection.class, options);
   }
+
```
mickjermsurawong-stripe: Terminal endpoints below actually would be correct, but it's not consistent. The client metadata now generate static methods

mickjermsurawong-stripe: Do you think we want to release this as 8.0 too?

src/main/java/com/stripe/model/Charge.java
```diff
@@ -194,6 +194,14 @@
...
   String paymentIntent;
 
+  /** ID of the payment method used in this charge. */
+  @SerializedName("payment_method")
+  String paymentMethod;
+
+  /** Transaction-specific details of the payment method used in the payment. */
+  @SerializedName("payment_method_details")
+  PaymentMethodDetails paymentMethodDetails;
+
```
remi-stripe: We're missing `billing_details` which is also shipping at the same time as `payment_method_details`. I realize it's also missing in my PRs though and I'll have to fix this but I think it's important to fix in pay-server

src/main/java/com/stripe/model/PaymentMethod.java
```diff
@@ -0,0 +1,455 @@
...
+       */
+      @SerializedName("cvc_check")
+      String cvcCheck;
+    }
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class ThreeDSecureUsage extends StripeObject {
+      /** 3D Secure is support on this card. */
```
remi-stripe: typo in the comment

src/main/java/com/stripe/model/terminal/Location.java
remi-stripe: Yeah, we need to quickly talk about static methods (globally, not just java) because I think this is becoming critical. I also need to release those in the libs so it's on my todo list

src/main/java/com/stripe/model/Charge.java
```diff
@@ -887,6 +895,647 @@ public void setRuleObject(Rule expandableObject) {
...
+      /**
+       * Card brand. Can be `amex`, `diners`, `discover`, `jcb`, `mastercard`, `unionpay`, `visa`,
+       * or `unknown`.
+       */
+      @SerializedName("brand")
+      String brand;
+
+      /** Check results by Card networks on Card address and CVC at time of payment. */
+      @SerializedName("checks")
+      Checks checks;
```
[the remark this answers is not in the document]
remi-stripe: Okay I saw the commend in my other PR too. I'll fix mine to just duplicate the class instead of trying to be smart.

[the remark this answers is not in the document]
mickjermsurawong-stripe: @remi-stripe, I would prefer not sharing in this example, simply because it is what the back-end actually has. So I stated before that now we prefer sharing, I'm actually wrong. I think I was saying that because in most case backend code do share resources. So my position is rather resource sharing just as what our backend and open API has.

In the world that we share nothing, that's when we can ignore the backend code and openAPI references (with exceptions of sharinging core resource)

mickjermsurawong-stripe:
Thanks remi for the review, I'll wait for billing details and will add to this PR then.

mickjermsurawong-stripe:
I'm merging PM changes in first. There are 3 pending issues in the subsequent PR
1) Undashboard billing details so it appears on the spec, fix typo on 3D secure support, in this pay-sever #137513
2) Update terminal client metadata endpoints to reflect instance method #137444
3) Remove deprecated endpoints + checkout

PR #697, opened 2019-03-20
Pass version check flag to stripe-mock

mickjermsurawong-stripe:
- This passes `-strict-version-check` flag to stripe mock to ensure pinned library is testing against correct OpenAPI version of stripe-mock.
- The first commit failed with the error below, and passing override flag with pinned version resolved it.
```
com.stripe.functional.EphemeralKeyTest > testCreate FAILED
    com.stripe.exception.InvalidRequestException: Version sent in `Stripe-Version` header '2017-05-25' doesn't match version in OpenAPI specification '2019-03-14' which may have unintended consequences. This error was shown because stripe-mock  was started with `-stripe-version-check`.
```
r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/test/java/com/stripe/StripeMockProcess.java
+++ b/src/test/java/com/stripe/StripeMockProcess.java
@@ -50,8 +50,7 @@ public class StripeMockProcess {
             "-spec",
             getPathSpec(),
             "-fixtures",
-            getPathFixture(),
-            "-strict-version-check"
+            getPathFixture()
         );
 
     try {
--- a/src/test/java/com/stripe/StripeMockProcess.java
+++ b/src/test/java/com/stripe/StripeMockProcess.java
@@ -50,7 +50,8 @@ public class StripeMockProcess {
             "-spec",
             getPathSpec(),
             "-fixtures",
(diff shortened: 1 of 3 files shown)


src/test/java/com/stripe/StripeMockProcess.java
```diff
@@ -50,7 +50,8 @@ public static boolean start() throws IOException, InterruptedException, NoSuchFi
             "-spec",
             getPathSpec(),
             "-fixtures",
-            getPathFixture()
+            getPathFixture(),
+            "-strict-version-check"
```
mickjermsurawong-stripe: I assume we want to pass strict version check when we do custom build too? Or do we want to me more lenient here as we are developing and might be testing against the latest OpenAPI spec?

remi-stripe: Flagging that we didn't do that one in the other libs

mickjermsurawong-stripe:
Self-approving after removing the version check to be consistent with other typed client libs.
LGTM

PR #700, opened 2019-03-20
stripe-java v9.0.0

### LINKED ISSUE #497: Start using Error Prone
[Error Prone](http://errorprone.info/) is a project that runs additional checks and improves code quality.

This was suggested by @codylerum in #494. See the PR for more context.

### LINKED ISSUE #562: Odd SSL errors with Stripe Java API libs
There is something odd in the way the Java Stripe API wrapper configures SSL: When I use it in OpenLiberty (OL) container, it pick the configured SSL certificates store from the container, which is empty by default, and I start to see SSL errors like this one:

https://github.com/OpenLiberty/open-liberty/issues/4377#issuecomment-408978980

The issue was solved by adding either Stripe.Com or parent (DigiCert) certificates to the OL trust store configuration. However, this should be unnecessary, since Stripe API need only JDK's standard cacert , right?

I don't see the same issue with Auth0.Com or Amazon AWS JAVA SDK 2.0 java libs, all bypassing the container SSL configuration and working flawlessly (I presume with the default JDK SSL context and cacert files).

You can examine the 2 Java libs I mentioned - they are both on github - and see what you are doing differently.

### LINKED ISSUE #605: Arrays in webhooks previous attributes parsed as Object[]

ob-stripe:
This is the WIP branch for the next major stripe-java version, v9.0.0.

List of PRs (⚠️ = breaking changes):
- [x] ⚠️ Drop support for Java 1.7 (#698)
- [x] Upgrade dev dependencies (except JUnit) (#702)
- [x] Upgrade to JUnit 5 (#706)
- [x] ⚠️ Use `Optional` for unsafe event data object deserialization (#723)
- [x] ⚠️ Add support for typed parameters 🔥🎉🎆🕺 (#705)
- [x] Use Error Prone in builds (#713)
- [x] Another dependency version bump (#728)
- [x] Add missing Javadoc (#729)
- [x] Add parameter classes for non-generated resources (`ephemeral_key`, `event` and `file`) (#730)

Excluded from this release:
- Add support for Java 12 (#703): looks like Lombok doesn't support Java 12 yet

Fixes #562 (via #698), fixes #605 (via #705), fixes #497 (via #713).


ob-stripe:
I think we're good to release once #730 is merged. cc @remi-stripe for a last quick 👀 in case we missed anything.

mickjermsurawong-stripe:
thank you ob!

PR #702, opened 2019-03-20
Upgrade dev dependencies (except JUnit)

ob-stripe:

Upgrade all our dependencies to their latest versions, except JUnit. (The JUnit 4->5 migration is a bit more complex, so I'll tackle it separately.)

I've replaced the deprecated `osgi` module with `biz.aQute.bnd.builder`. I've verified that the `MANIFEST.MF` file in the JAR is basically the same.


mickjermsurawong-stripe:
Whoops sorry for the delay.. Didn't get to review the other PR when you remove support for java7 as well. Not sure I understand all the necessary changes in build file now. 
But the build failure in Openjdk 10 is due to flaky test, not related to build here.
Thank you @ob-stripe

PR #703, opened 2019-03-20
Start testing Java 12

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Start testing Java 12.

I had to disable a specific ErrorProne check with Java 12 because of an issue with ErrorProne itself (cf. https://github.com/stripe/stripe-java/pull/703#issuecomment-491465260). Once they fix the issue, we can re-enable the check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ matrix:
     - jdk: openjdk9
     - jdk: openjdk10
     - jdk: openjdk11
+    - jdk: openjdk12
     - jdk: openjdk-ea
   allow_failures:
     - jdk: openjdk-ea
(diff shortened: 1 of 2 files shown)


ob-stripe:
Looks like Travis doesn't support Java 12 yet. I'll keep this open for now.

ob-stripe:
It looks like Lombok doesn't yet support Java 12. Travis doesn't seem to support OracleJDK 12, but with OpenJDK 12 we're running into this error: https://github.com/rzwitserloot/lombok/issues/2082.

ob-stripe:
Okay, so the Lombok issue was fixed in 1.18.8, but there is another issue with Error Prone: https://github.com/google/error-prone/issues/1257.

They already have a fix for it: https://github.com/google/error-prone/pull/1107, but it's not merged yet. Once it's released, hopefully we can start testing Java 12 for real 🤞

brandur-stripe:
Wow, not an easy road getting this turned on :)

LGTM.

PR #706, opened 2019-03-20
Upgrade to JUnit 5

ob-stripe:
cc @stripe/api-libraries 

Upgrade to JUnit 5.

Lots of noise here, sorry! The main changes are:
- `org.junit.Assert` -> `org.junit.jupiter.api.Assertions`
- some assertion methods like `assertTrue` and `assertSame` now take the failed assertion message as their last argument instead of their first
- checking for thrown exception is now done via `assertThrows`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,10 +44,11 @@ repositories {
 dependencies {
     compile group: 'com.google.code.gson', name: 'gson', version:'2.8.5'
     testCompile group: 'com.google.guava', name: 'guava', version:'27.1-jre'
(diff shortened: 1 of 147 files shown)


mickjermsurawong-stripe:
yay thank you @ob-stripe for the change!
Putting assertion message last makes it much more readable.
LGTM

ob-stripe:
Thanks Mick!

Some of our tests are flaky on Travis... It's nothing new, but we should look into it at some point, having to trigger rebuilds is annoying.

PR #708, opened 2019-03-21
Fix issues detected by ErrorProne

ob-stripe:
cc @stripe/api-libraries 

Fix a bunch of issues reported by ErrorProne. Nothing major or user visible.

The remaining issues need to be fixed in the autogenerator itself:
- missing `@Override` annotations on interface method implementations
- `ScheduledQueryRun.Error` should be renamed to something else, as `Error` is already a class in `java.lang` and auto-imported


src/test/java/com/stripe/functional/StripeResponseTest.java
```diff
@@ -28,7 +26,7 @@ public void testResponseIncluded() throws StripeException {
     final Customer customer = Customer.create(null, requestOptions);
     final Customer retrievedCustomer = Customer.retrieve(customer.getId(), requestOptions);
     final StripeResponse response = retrievedCustomer.getLastResponse();
-    assertThat(response, instanceOf(StripeResponse.class));
+    assertNotNull(response);
```
ob-stripe: The `instanceOf()` assertion is useless: `response` is declared as a `StripeResponse` so we already know the type at compile-time.

src/test/java/com/stripe/functional/TelemetryTest.java
```diff
@@ -26,7 +28,7 @@
 public class TelemetryTest extends BaseStripeTest {
   @Test
   public void testTelemetryEnabled() throws StripeException, IOException, InterruptedException {
-    MockWebServer server = new MockWebServer();
+    @Cleanup MockWebServer server = new MockWebServer();
```
mickjermsurawong-stripe: ah I learned something new!
I suppose we never "close" this webserver resource in the first place?

mickjermsurawong-stripe:
wow thank you ob for the clean up!
yup the @override annotation is in the genrator PR now.
for `ScheduledQueryRun.Error` do you suggest doing it with 8.1 now? 
LGTM

src/main/java/com/stripe/exception/StripeException.java
```diff
@@ -41,14 +41,14 @@ public StripeException(String message, String requestId, String code, Integer st
...
+  public String getMessage() {
     String additionalInfo = "";
     if (code != null) {
       additionalInfo += "; code: " + code;
     }
     if (requestId != null) {
       additionalInfo += "; request-id: " + requestId;
     }
-    return super.toString() + additionalInfo;
+    return super.getMessage() + additionalInfo;
```
mickjermsurawong-stripe: ah actually question here please.
Do we actually change the behavior here? Such that if we print the error, it will return hashcode like java object?

src/test/java/com/stripe/functional/TelemetryTest.java
ob-stripe: Yes, previously the server was "leaked" because we never called `close()`. Lombok's [`@Cleanup`](https://projectlombok.org/features/Cleanup) annotation is super useful to deal with such resources and we already use it in a few other places.

src/main/java/com/stripe/exception/StripeException.java
ob-stripe: `toString()` calls `getMessage()`, so the behavior shouldn't change here. ErrorProne recommends overriding `getMessage()` instead of `toString()` though: https://errorprone.info/bugpattern/OverrideThrowableToString.

ob-stripe:
Thanks Mick!

> for `ScheduledQueryRun.Error` do you suggest doing it with 8.1 now?

If we rename the class to something else, it will be a breaking change, so we should do this in the next major version.

Alternatively, we could simply silence the warning. It's doubtful anyone would import a nested class directly -- users are more likely to import the parent class `ScheduledQueryRun`, and refer to the nested class with `ScheduledQueryRun.Error`, so there wouldn't be any ambiguity with the default `Error` class.

PR #713, opened 2019-03-29
Use ErrorProne in builds

### LINKED ISSUE #497: Start using Error Prone

ob-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries 

Use [ErrorProne](http://errorprone.info/) in builds.

Fixes #497.

There were only three issues remaining, two of which I think we can ignore:
- `ScheduledQueryRun.Error`: ErrorProne complains that `Error` is a builtin class in `java.lang` and we should change the name. Given that `ScheduledQueryRun.Error` is a nested class and it's very unlikely any user would import it directly, I think it's safe to ignore this.
- `assertTrue(subscription.getTaxPercent().equals(new BigDecimal("0.3")))`: ErrorProne warns that `BigDecimal.equals` compares the scale of the representation as well as the value (i.e. `1.0 != 1.00`). This is actually desirable in our case, so this is a false positive.

~~Unfortunately, the last one is another missing `@Override` in autogenerated code, here:~~ https://github.com/stripe/stripe-java/blob/7518b68b389d1675762d79dc0dcce29424cfdff4/src/main/java/com/stripe/model/ExternalAccountTypeAdapterFactory.java#L83

~~Not sure how hard it would be to fix this. wdyt?~~ Fixed, thanks Mick!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ plugins {
     id 'java'
     id 'maven'
     id 'io.franzbecker.gradle-lombok' version '2.1'
+    id "net.ltgt.errorprone" version "0.7.1"
     id 'checkstyle'
     id 'net.saliman.cobertura' version '2.6.0'
     id 'com.github.kt3k.coveralls' version '2.8.2'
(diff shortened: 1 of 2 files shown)


ob-stripe:
Oh oops, I just realized that `ScheduledQueryRun.java` is autogenerated too 🤦‍♂️

We'd need to have the generator add the `@SuppressWarning` annotation. I think this should be fairly simple (basically "if classname is `Error` then add the `@SuppressWarning` annotation).

mickjermsurawong-stripe:
@ob-stripe 
do you think it's worth renaming the class? I can update the client-metadata and then we release it together with v9.0?

mickjermsurawong-stripe:
On @override id. I can add that in the generator. Sorry didn't manage to get to. 
But will do that before we release v9.0.

ob-stripe:
Okay, this should be ready now. ptal @mickjermsurawong-stripe

mickjermsurawong-stripe:
LGTM
thank you @ob-stripe!

PR #715, opened 2019-03-30
Update plan param `name` to `nickname`

mickjermsurawong-stripe:
This should make build pass when we update stripe-mock to latest version
r? @remi-stripe 
cc @stripe/api-libraries


mickjermsurawong-stripe:
@remi-stripe 
It's in the spreadsheet we went through, so i removed them from openapi. `name` in plan is no longer there according to this version change https://stripe.com/docs/upgrades#2018-02-05

remi-stripe:
yep sorry I understand the why. But how was it still in openapi since it was a version change from a year ago?

PR #717, opened 2019-04-02
Add @Override to getter for unknown subtypes - [generated] source: spec3.sdk.yaml@non-master-spec-3c790a5 

mickjermsurawong-stripe:
This PR generates from the same spec OpenAPI spec at `3c790a5` that generated the last release[0], so there's no new support added, but it is built from generator that now adds `@Override` to fix ErrorPrune for integration branch 9

[0] https://github.com/stripe/stripe-java/commit/c5f23429ba7f460060cceaf59b7b8b9d3d502f7e
r? @ob-stripe 
cc @stripe/api-libraries


mickjermsurawong-stripe:
Just to confirm, we currently have 3 interface, `BalanceTransactionSource`, `ExternalAccount`, and `PaymentSource`, so the three unknown subtypes during deserialization now has the right `@Override`

PR #718, opened 2019-04-03
Rename param enums and add test to verify enum polymorphism 

mickjermsurawong-stripe:
- This PR renames `ApiRequestParams.Enum` to `ApiRequestParams.EnumParam`, and 
introduces `Empty` to `EmptyParam` in a param common package.
- This adds more test to reflect the polymorphic of enum that we accept both empty and non-empty enum in the same field
r? @ob-stripe 
cc @stripe/api-libraries


mickjermsurawong-stripe:
thank you @ob-stripe!

PR #720, opened 2019-04-05
Add delombokTest Gradle task

ob-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries 

Adds a `delombokTest` Gradle task to generate delombok'd sources for tests.

I've also changed the output directory for the existing `delombok` task, and checked that the tasks that depend on it (`sourcesJar` and `makeJavadocsJar`) still work.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,9 +71,21 @@ lombok {
 import io.franzbecker.gradle.lombok.task.DelombokTask
 
 task delombok(type: DelombokTask, dependsOn: compileJava) {
-    ext.outputDir = file("$buildDir/delombok")
+    ext.outputDir = file("$buildDir/delombok/main")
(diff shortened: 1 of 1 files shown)


mickjermsurawong-stripe:
Ob just for my understanding please, that this delomboked test will be just for completeness and now not further used in actual release? Seems like only the `delombok` (on the source) is used in `deploy.gradle`.
ptal @ob-stripe

ob-stripe:
@mickjermsurawong-stripe Correct! It's just there in case we need it for debugging Lombok-related issues in tests like we did this morning, but it's not actually being used anywhere in the release flow.

mickjermsurawong-stripe:
LGTM!
thank you @ob-stripe

PR #721, opened 2019-04-05
Stop testing Oracle JDK for Java >= 11

ob-stripe:
r? @brandur-stripe @mickjermsurawong-stripe 
cc @stripe/api-libraries 

From Java 11 onwards, only test with OpenJDK.

From Oracle's [blog post](https://blogs.oracle.com/java-platform-group/oracle-jdk-releases-for-java-11-and-later):
> From Java 11 forward, therefore, Oracle JDK builds and OpenJDK builds will be essentially identical.

Also, the script used by Travis to install JDK versions is deprecating support for Oracle JDK (https://github.com/sormuras/bach/issues/33).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,12 +39,9 @@ matrix:
     - jdk: oraclejdk9
     - jdk: openjdk9
     - jdk: openjdk10
-    - jdk: oraclejdk11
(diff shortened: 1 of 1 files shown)


ob-stripe:
Thanks Brandur!

PR #722, opened 2019-04-05
Verify deserialized boolean params to map

mickjermsurawong-stripe:
- This PR adds test on `toMap` to ensure that uninitialize booleans will not implicitly be serialized as false. It seems like it was a problem with some other client-libs before.
r? @ob-stripe 
cc @remi-stripe


mickjermsurawong-stripe:
yup Ob. I should add a check there too!

PR #723, opened 2019-04-06
Return unsafe deserialized event data object as Option

mickjermsurawong-stripe:
- Now that we support java language version > v8, we can use optional to support when the getting deserialized object from webhook.
- This replaces the boolean check method `deserialize` 

Currently, the expected integration will look like this.. 
```java
   Event event = Webhook.constructEvent(payload, sigHeader, secret);
   EventDataObjectDeserializer dataObjectDeserializer = event.getDataObjectDeserializer();
   if (dataObjectDeserializer.deserialize()) {
      StripeObject stripeObject = dataObjectDeserializer.getObject();
   }
```
The problem is rather, user might just go straight and `getObject()` which can be null if it is not safe.
Now
```java
   Event event = Webhook.constructEvent(payload, sigHeader, secret);
   EventDataObjectDeserializer dataObjectDeserializer = event.getDataObjectDeserializer();
   if (dataObjectDeserializer.getObject().isPresent()) {
     StripeObject stripeObject = dataObjectDeserializer.getObject().get();
     doSomething(stripeObject);
   } 
```
For unsafe workflow, it's still the same. User can force deserialization or can use compatability to make the Json compatible

The original implementation and discussion in here https://github.com/stripe/stripe-java/pull/654
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Event.java
+++ b/src/main/java/com/stripe/model/Event.java
@@ -38,7 +38,7 @@ public class Event extends ApiResource implements HasId {
    * <pre>
    *   Event event = getEvent(); // either from webhook or event endpoint
    *   EventDataObjectDeserializer deserializer = event.getDataObjectDeserializer();
-   *   StripeObject stripeObject = deserializer.getObject();
+   *   Optional&lt;StripeObject&gt; stripeObject = deserializer.getObject();
    * </pre>
    * You can ensure that webhook events has the same API version by creating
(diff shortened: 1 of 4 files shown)


mickjermsurawong-stripe:
self-approving after fixing one doc to use `Optional<StripeObject>` for event deserialization.
LGTM
thank you ob!

PR #729, opened 2019-04-09
Add missing Javadoc

ob-stripe:
cc @stripe/api-libraries 

Adds missing Javadoc on non-generated resource classes.

### DIFF
--- a/src/main/java/com/stripe/model/EphemeralKey.java
+++ b/src/main/java/com/stripe/model/EphemeralKey.java
@@ -1,5 +1,7 @@
 package com.stripe.model;
 
+import com.google.gson.annotations.SerializedName;
+
 import com.stripe.exception.StripeException;
 import com.stripe.net.ApiResource;
 import com.stripe.net.RequestOptions;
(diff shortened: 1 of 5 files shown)


src/main/java/com/stripe/model/EphemeralKey.java
```diff
@@ -43,24 +67,21 @@ public static EphemeralKey create(Map<String, Object> params, RequestOptions opt
...
-   * Delete an ephemeral key.
+   * Invalidates a short-lived API key for a given resource.
    */
   public EphemeralKey delete() throws StripeException {
     return delete(null);
   }
 
   /**
-   * Delete an ephemeral key.
+   * Invalidates a short-lived API key for a given resource.
```
remi-stripe: Minor but I have a preference for keeping "ephemeral key" since it's the object name. No strong feels though

src/main/java/com/stripe/model/Event.java
```diff
@@ -14,15 +16,55 @@
 @Setter
 @EqualsAndHashCode(callSuper = false)
 public class Event extends ApiResource implements HasId {
-  @Getter(onMethod = @__({@Override})) String id;
-  String object;
+  @SerializedName("account")
```
remi-stripe: This should have a documentation right?

src/main/java/com/stripe/model/File.java
```diff
@@ -15,73 +17,119 @@
...
 
-  // <editor-fold desc="create">
   /**
-   * Create a file.
+   * To upload a file to Stripe, you’ll need to send a request of type {@code multipart/form-data}.
+   * The request should contain the file you would like to upload, as well as the parameters for
+   * creating a file.
+   *
+   * <p>All of Stripe’s officially supported API libraries should have support for sending
+   * {@code multipart/form-data}.
```
remi-stripe: This description is "quite weird" lol. I know we want to stick to the docs so not blocking but a better wording is likely worth it

src/main/java/com/stripe/model/Event.java
mickjermsurawong-stripe: I made a pay-sever PR 142016 with `account` documentation for OpenAPI. We can consider including the doc as
```
For Connect webhook events, the identifier of the account that emitted the event.
```
I modified the doc from .NET here
https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/Events/Event.cs#L23
but emphasized that it's only on webhook but not events from endpoint

mickjermsurawong-stripe:
@ob-stripe 
- On the event "account", if the wording I proposed seems okay, we can add to the java doc? At least we will have parity with OpenAPI on this.
- For `File` doc on "All of Stripe’s officially supported API libraries...", I think it's okay to remove it from the non-autogen classes. But I'm fine leaving as is too.

src/main/java/com/stripe/model/EphemeralKey.java
ob-stripe: Fixed.

src/main/java/com/stripe/model/Event.java
ob-stripe: 👍  Fixed.

src/main/java/com/stripe/model/File.java
ob-stripe: I've removed the second paragraph for now to make it slightly less weird.

PR #730, opened 2019-04-09
Add non-autogen typed params Event/File/EphemeralKeys

mickjermsurawong-stripe:
- This PR implements typed params for the non-autogen classes: `Event`, `File` and `EphemeralKey`
- I used generator in this [branch](https://git.corp.stripe.com/stripe-internal/sdk-autogen-java/commit/9bee243aa18027104cbabec6487eb71caad8391f) to get parameters for the excluded files, and manually remove the special-case logic that we have 
- For file, there's additional handling of `File` and `FileInputStream` for typed params to make it compatible to the current untyped params usage. 
- For ones with custom logic in the method itself, I defer the calls to existing logic, and multi-part request that currently doesn't have interface for typed parameters.
- I did not add `EventRetrieve` and `FileRetrieve` params because they are essentially no-op params, having only expand while the model actually has nothing expandable.

- It's a large PR but reviewing by commit would be helpful. 

r? @ob-stripe @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/ApiResource.java
+++ b/src/main/java/com/stripe/net/ApiResource.java
@@ -166,6 +166,7 @@ public abstract class ApiResource extends StripeObject {
   public static <T> T request(ApiResource.RequestMethod method,
                               String url, ApiRequestParams params, Class<T> clazz,
                               RequestOptions options) throws StripeException {
+    checkNullTypedParams(url, params);
     return request(method, url, params.toMap(), clazz, options);
   }
 
@@ -179,6 +180,7 @@ public abstract class ApiResource extends StripeObject {
   public static <T extends StripeCollectionInterface<?>> T requestCollection(
       String url, ApiRequestParams params, Class<T> clazz, RequestOptions options)
       throws StripeException {
+    checkNullTypedParams(url, params);
     return requestCollection(url, params.toMap(), clazz, options);
   }
 
@@ -202,6 +204,19 @@ public abstract class ApiResource extends StripeObject {
     }
 
     return collection;
+  }
+
+  /**
+   * Invalidate null typed parameters.
+   * @param url     request url associated with the given parameters.
+   * @param params  typed parameters to check for null value.
+   */
+  public static void checkNullTypedParams(String url, ApiRequestParams params) {
+    if (params == null) {
+      throw new IllegalArgumentException(String.format("Found null params for %s. "
+          + "Please pass empty params using param builder via `builder().build()` instead.", url
+      ));
+    }
   }
 
   /**
--- a/src/main/java/com/stripe/param/EphemeralKeyCreateParams.java
+++ b/src/main/java/com/stripe/param/EphemeralKeyCreateParams.java
@@ -1,0 +1,81 @@
+package com.stripe.param;
+
+import com.google.gson.annotations.SerializedName;
+import com.stripe.net.ApiRequestParams;
+import java.util.ArrayList;
+import java.util.List;
+
+public class EphemeralKeyCreateParams extends ApiRequestParams {
+  /** The ID of the Customer you'd like to modify using the resulting ephemeral key. */
+  @SerializedName("customer")
+  String customer;
+
+  /** Specifies which fields in the response should be expanded. */
+  @SerializedName("expand")
+  List<String> expand;
+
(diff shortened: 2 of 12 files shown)


src/main/java/com/stripe/param/EphemeralKeyCreateParams.java
```diff
@@ -0,0 +1,81 @@
+package com.stripe.param;
```
mickjermsurawong-stripe: There's no change to this file from the generator. But I removed the comment that is generated.
```
 // Generated by com.stripe.generator.entity.SdkBuilder
```

src/main/java/com/stripe/param/EventListParams.java
```diff
@@ -0,0 +1,295 @@
+package com.stripe.param;
```
mickjermsurawong-stripe: No change to this list params from the generated

src/main/java/com/stripe/param/FileCreateParams.java
```diff
@@ -0,0 +1,253 @@
...
+  List<String> expand;
+
+  /**
+   * A file to upload. The file should follow the specifications of RFC 2388 (which defines file
+   * transfers for the `multipart/form-data` protocol).
+   *
+   * <p>Note this field is marked as transient to avoid JSON deserializer. Override
+   * {@link FileCreateParams#toMap()} makes sure that the returned map has this same file instance.
+   */
+  transient Object file;
```
mickjermsurawong-stripe: Change in params. From the spec, we would accept String value for the file. 
It is `Object` because the builder allows setting `File` and `FileInputStream`

src/main/java/com/stripe/param/FileCreateParams.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /**
+     * A file to upload. The file should follow the specifications of RFC 2388 (which defines file
+     * transfers for the `multipart/form-data` protocol).
+     */
+    public Builder setFile(java.io.FileInputStream file) {
+      this.file = file;
+      return this;
+    }
```
mickjermsurawong-stripe: Changes here from setting string to setting File.

src/main/java/com/stripe/param/FileCreateParams.java
```diff
@@ -0,0 +1,253 @@
...
+   * {@link Builder#setFile(java.io.FileInputStream)}; the file is not transformed or serialized
+   * at this level.
+   * @return Untyped parameters containing file object set in the builder.
+   */
+  @Override
+  public Map<String, Object> toMap() {
+    Object fileObject = this.file;
+    Map<String, Object> untypedParamWithPrimitiveTypes = super.toMap();
+    untypedParamWithPrimitiveTypes.put("file", fileObject);
+    return untypedParamWithPrimitiveTypes;
```
mickjermsurawong-stripe: Because `toMap` needs to contain the same file instance set in the builder

src/main/java/com/stripe/param/FileListParams.java
```diff
@@ -0,0 +1,272 @@
+package com.stripe.param;
+
+import com.google.gson.annotations.SerializedName;
+import com.stripe.net.ApiRequestParams;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.Getter;
+
+public class FileListParams extends ApiRequestParams {
```
mickjermsurawong-stripe: No change in this params.

src/main/java/com/stripe/net/ApiResource.java
```diff
@@ -166,6 +166,7 @@ public static String urlEncode(String str) throws UnsupportedEncodingException {
   public static <T> T request(ApiResource.RequestMethod method,
                               String url, ApiRequestParams params, Class<T> clazz,
                               RequestOptions options) throws StripeException {
+    checkNullTypedParams(url, params);
```
mickjermsurawong-stripe: Custom error message here instead of failing in the next line `toMap`

src/main/java/com/stripe/param/FileCreateParams.java
ob-stripe: 👍

mickjermsurawong-stripe:
thank you ob!

PR #735, opened 2019-04-10
Fix encoding of nested parameters in multipart requests

mickjermsurawong-stripe:
- This fixes encoding of nested params in multipart request. Currently, we assume that parameters used in multipart request is one-level deep and contains file object at root-level.
- This PR allows nested params, and makes assumption of root-level file object explicit.
- This is similar to the problem addressed in php https://github.com/stripe/stripe-php/pull/624

r? @ob-stripe 

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -367,6 +367,13 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
     } else if (value == null) {
       flatParams = new ArrayList<>();
       flatParams.add(new Parameter(keyPrefix, ""));
+    } else if ((value instanceof File) || (value instanceof InputStream)) {
+      throw new InvalidRequestException(String.format(
+          "java.io.File or java.io.InputStream %s is not supported at '%s' parameter. "
+          + "Please check our API reference for the parameter type, "
+          + "or use the provided parameter class instead.",
+          value, keyPrefix),
+          keyPrefix, null, null, 0, null);
     } else {
       flatParams = new ArrayList<>();
       flatParams.add(new Parameter(keyPrefix, value.toString()));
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -682,6 +651,56 @@ private static StripeResponse getMultipartStripeResponse(
...
+            new FileInputStream(currentFile));
+      } else if (value instanceof InputStream) {
+        @Cleanup InputStream inputStream = (InputStream) value;
+        if (inputStream.available() == 0) {
+          throw new InvalidRequestException(
+            "Must have available bytes to read on InputStream for key "
+              + key + ".", null, null, null, 0, null
+          );
+        }
+        multipartProcessor.addFileField(key, "blob", inputStream);
```
mickjermsurawong-stripe: The two branches above are the same as before. 
Only else here delegates the the flatten params used by `createQuery` in normal request method type

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -367,6 +367,13 @@ static String createQuery(Map<String, Object> params)
     } else if (value == null) {
       flatParams = new ArrayList<>();
       flatParams.add(new Parameter(keyPrefix, ""));
+    } else if ((value instanceof File) || (value instanceof InputStream)) {
+      throw new InvalidRequestException(String.format(
+          "java.io.File or java.io.InputStream %s is not supported at '%s' parameter. "
+          + "Please check our API reference for the parameter type, "
+          + "or use the provided parameter class instead.",
+          value, keyPrefix),
+          keyPrefix, null, null, 0, null);
```
mickjermsurawong-stripe: The error here could arise 
- if users pass file object to non-multipart request at any level.
- multi-part request params having nested file objects

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -607,46 +614,8 @@ private static StripeResponse getMultipartStripeResponse(
...
-            multipartProcessor.addFileField(key, "blob", inputStream);
-          } else {
-            // We only allow a single level of nesting for params
-            // for multipart
-            multipartProcessor.addFormField(key, (String) value);
-          }
-        }
-
+        multipartProcessor = new MultipartProcessor(conn, boundary, ApiResource.CHARSET);
+        encodeMultipartParams(multipartProcessor, params);
```
mickjermsurawong-stripe: Alternative design here is to have `encodeParams` as a method to multipartprocessor itself.
However, we still want to use the same `flattenParams` and `LiveRequestGetter.Parameter` within in class.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
ob-stripe: 👍  Excellent, thanks for adding an explicit check and error.

PR #739, opened 2019-04-11
[generated] source: spec3.sdk.yaml@spec-5869a77 in master

mickjermsurawong-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 
- Generated from master sdk generator without the fix for ID encoding yet, and on latest OpenAPI spec
- This contains:
 - `ConfirmationMethod` in `PaymentIntentCreateParams`
 - `AuthorizationControls` in issuing cardholders model/params
- If we can merge this first, and then this PR https://github.com/stripe/stripe-java/pull/737 can use the latest spec OpenAPI spec with sdk generator with fixed ID encoding

### DIFF
--- a/src/main/java/com/stripe/model/Invoice.java
+++ b/src/main/java/com/stripe/model/Invoice.java
@@ -244,11 +244,6 @@ public class Invoice extends ApiResource implements HasId, MetadataStore<Invoice
   @SerializedName("paid")
   Boolean paid;
 
-  /**
-   * The PaymentIntent associated with this invoice. The PaymentIntent is generated when the invoice
-   * is finalized, and can then be used to pay the invoice. Note that voiding an invoice will cancel
-   * the PaymentIntent.
-   */
   @SerializedName("payment_intent")
   PaymentIntent paymentIntent;
 
@@ -756,9 +751,9 @@ public class Invoice extends ApiResource implements HasId, MetadataStore<Invoice
   /**
    * Stripe automatically creates and then attempts to collect payment on invoices for customers on
    * subscriptions according to your <a
-   * href="https://dashboard.stripe.com/account/recurring">subscriptions settings</a>. However, if
-   * you’d like to attempt payment on an invoice out of the normal collection schedule or for some
-   * other reason, you can do so.
+   * href="https://dashboard.stripe.com/account/billing/automatic">subscriptions settings</a>.
+   * However, if you’d like to attempt payment on an invoice out of the normal collection schedule
+   * or for some other reason, you can do so.
    */
   public Invoice pay() throws StripeException {
     return pay((Map<String, Object>) null, (RequestOptions) null);
@@ -767,9 +762,9 @@ public class Invoice extends ApiResource implements HasId, MetadataStore<Invoice
   /**
    * Stripe automatically creates and then attempts to collect payment on invoices for customers on
    * subscriptions according to your <a
-   * href="https://dashboard.stripe.com/account/recurring">subscriptions settings</a>. However, if
-   * you’d like to attempt payment on an invoice out of the normal collection schedule or for some
-   * other reason, you can do so.
+   * href="https://dashboard.stripe.com/account/billing/automatic">subscriptions settings</a>.
+   * However, if you’d like to attempt payment on an invoice out of the normal collection schedule
+   * or for some other reason, you can do so.
    */
   public Invoice pay(RequestOptions options) throws StripeException {
     return pay((Map<String, Object>) null, options);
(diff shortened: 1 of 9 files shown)


src/main/java/com/stripe/model/Invoice.java
```diff
@@ -244,11 +244,6 @@
   @SerializedName("paid")
   Boolean paid;
 
-  /**
-   * The PaymentIntent associated with this invoice. The PaymentIntent is generated when the invoice
-   * is finalized, and can then be used to pay the invoice. Note that voiding an invoice will cancel
-   * the PaymentIntent.
-   */
```
remi-stripe: we should keep the docs here. I think we internally hid this for a bit but still

src/main/java/com/stripe/model/issuing/Cardholder.java
```diff
@@ -24,6 +25,9 @@
 @Setter
 @EqualsAndHashCode(callSuper = false)
 public class Cardholder extends ApiResource implements HasId, MetadataStore<Cardholder> {
+  @SerializedName("authorization_controls")
+  AuthorizationControls authorizationControls;
+
```
remi-stripe: While this is merged, it's in theory gated and coming with a new API version I think so we should consider removing for a few days

src/main/java/com/stripe/param/PaymentIntentCreateParams.java
```diff
@@ -37,6 +43,21 @@
   @SerializedName("confirm")
   Boolean confirm;
 
+  /**
+   * One of `automatic` (default) or `manual`.
+   *
+   * <p>When the confirmation method is `automatic`, a PaymentIntent can be confirmed using a
```
remi-stripe: is the `<p>` okay in javadoc?

src/main/java/com/stripe/param/PaymentIntentCreateParams.java
```diff
@@ -11,7 +11,13 @@
 import lombok.Getter;
 
 public class PaymentIntentCreateParams extends ApiRequestParams {
-  /** Amount intended to be collected by this PaymentIntent. */
+  /**
+   * A positive integer representing how much to charge in the [smallest currency
+   * unit](https://stripe.com/docs/currencies#zero-decimal) (e.g., 100 cents to charge $1.00 or 100
+   * to charge ¥100, a zero-decimal currency). The minimum amount is $0.50 US or [equivalent in
+   * charge
+   * currency](https://support.stripe.com/questions/what-is-the-minimum-amount-i-can-charge-with-stripe).
```
remi-stripe: do those markdown-like links work in javadoc?

src/main/java/com/stripe/param/PaymentIntentCreateParams.java
mickjermsurawong-stripe: Yup <p> is a standard tag to add a preceding empty line

src/main/java/com/stripe/param/PaymentIntentCreateParams.java
mickjermsurawong-stripe: No unfortunately. We should do one clean up on this. But the link still works.

ob-stripe: We should look into using [this project](https://github.com/Abnaxos/markdown-doclet) to enable the use of Markdown syntax in Javadoc.

src/main/java/com/stripe/model/Invoice.java
mickjermsurawong-stripe: Huh i realized it's a problem in OpenAPI generator here. 
Will make a fix.

mickjermsurawong-stripe: Okay since it's an old problem and I have a working PR. We can punt this first.

src/main/java/com/stripe/model/issuing/Cardholder.java
mickjermsurawong-stripe: We agreed now that this is okay

mickjermsurawong-stripe:
ptal @remi-stripe 
This will merge first and then I'll regen with url-encoded id fix in another PR

src/main/java/com/stripe/param/ChargeCreateParams.java
```diff
@@ -242,11 +243,11 @@ public ChargeCreateParams build() {
...
 
     /**
-     * A positive integer representing how much to charge, in the [smallest currency
-     * unit](https://stripe.com/docs/currencies#zero-decimal) (e.g., `100` cents to charge $1.00, or
-     * `100` to charge ¥100, a zero-decimal currency). The minimum amount is $0.50 USD or
-     * [equivalent in charge
-     * currency](https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts).
+     * A positive integer representing how much to charge in the [smallest currency
+     * unit](https://stripe.com/docs/currencies#zero-decimal) (e.g., 100 cents to charge $1.00 or
+     * 100 to charge ¥100, a zero-decimal currency). The minimum amount is $0.50 US or [equivalent
```
remi-stripe: why did we replace USD with US, it's just a worse wording :(

PR #743, opened 2019-04-15
[wip] Service-oriented method calls

mickjermsurawong-stripe:
- This PR implements the autogen for service approach in java + method params overloading to suggest parent resource.

- The first commit is basic primitives
- The second is generated.
- The following with commits prefixed "test: " are examples for specific use cases. 
- This PR is a **prototype** of the approach I'm in favor now. I will follow-up with a doc where we can discuss
  - Should it be service-oriented, or static methods on model? 
  - How to represent sub-resource methods

cc @ob-stripe, @brandur-stripe , @remi-stripe 
cc @stripe/api-libraries


src/test/java/com/stripe/service/CardServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    stubRequest(
+        ApiResource.RequestMethod.DELETE,
+        String.format("/v1/customers/%s/sources/%s", CUSTOMER_ID, CARD_ID),
+        null, Card.class, deleteCardData
+    );
+    final Card deletedCard = cardService.delete(
+        CUSTOMER_ID,
+        CARD_ID,
+        CardDeleteOnCustomerParams.builder().build());
```
mickjermsurawong-stripe: `CardDeleteOnCustomerParams` is actually empty, because delete does not require additional params now.
However, since we rely on params typedness, it's necessary we have delete params to distinguish the parent..
`CardDeleteOn<ParentParams>`

src/test/java/com/stripe/service/ExternalAccountServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+class ExternalAccountServiceTest extends BaseStripeTest {
+  public static final String ACCOUNT_ID = "acct_123";
+  public static final String BANK_ACCOUNT_ID = "btok_123";
+  public static final String CARD_ID = "card_123";
+
+  private ExternalAccountService externalAccountService = new ExternalAccountService();
```
mickjermsurawong-stripe: Currently, static methods on Card/BankAccount: list/retrieve/create are done on the polymorphic collection.
`ExternalAccountCollection` and `PaymentSourceCollection`.
We can consider having similar `ExternalAccountService` and `PaymentSourceService`,
or we can move `list/retrieve/create` to Card/BankAccount and similarly rely on params overloading

src/test/java/com/stripe/service/PersonServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Test
+  public void testList() throws StripeException {
+    final Map<String, Object> params = new HashMap<>();
+    params.put("limit", 1);
+
+    final PersonCollection persons = personService.list(
+        ACCOUNT_ID,
+        PersonListParams.builder()
+            .setLimit(1L).build());
```
mickjermsurawong-stripe: List person no longer has to be from account: `account.persons()`

src/test/java/com/stripe/service/SubscriptionItemServiceTest.java
```diff
@@ -0,0 +1,34 @@
...
+  private static final String ITEM_ID = "si_123";
+
+  private SubscriptionItemService subscriptionItemService = new SubscriptionItemService();
+
+  @Test
+  public void testUsageRecordSummaries() throws StripeException {
+    final UsageRecordSummaryCollection summaries = subscriptionItemService.usageRecordSummaries(
+        ITEM_ID,
+        SubscriptionItemUsageRecordSummariesParams.builder()
+            .setLimit(1L).build());
```
mickjermsurawong-stripe: This contrasts with (new) `PersonService#list` and (old) `Account#persons` where nested resource has its dedicated service.
UsageRecord, however, isn't a "core" resource, so should it need its dedicated service? 
The original way to get summaries is `SubscriptionItem#usageRecordSummaries`. 
We could implement
a) `UsageRecordSummaryService#list`  like person
b) `SubscriptionItemService#usageRecordSummaries` as done here.

src/test/java/com/stripe/service/LoginLinkServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    stubRequest(
+        ApiResource.RequestMethod.POST,
+        String.format("/v1/accounts/%s/login_links", ACCOUNT_ID),
+        params,
+        LoginLink.class,
+        getResourceAsString("/api_fixtures/login_link.json")
+    );
+
+    LoginLink link = loginLinkService.create(
```
mickjermsurawong-stripe: We originally have `LoginLink#createOnAccount` now with param overloading in service, we can drop parent's name `LoginLinkService#create`

src/test/java/com/stripe/service/PaymentIntentServiceTest.java
```diff
@@ -0,0 +1,42 @@
...
+    verifyRequest(
+        ApiResource.RequestMethod.POST,
+        String.format("/v1/payment_intents/%s/capture", PAYMENT_INTENT_ID)
+    );
+  }
+
+  @Test
+  public void testConfirm() throws StripeException {
+    final PaymentIntent confirmedPaymentIntent = paymentIntentService
+        .confirm(PAYMENT_INTENT_ID, PaymentIntentConfirmParams.builder().build());
```
mickjermsurawong-stripe: Non controversial example of custom method on payment intent; we can update without fetching intent object.

PR #746, opened 2019-04-17
[generated] source: spec3.sdk.yaml@spec-5c641d5 in master

remi-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries


remi-stripe:
removing r? because it's missing some fields on `Invoice` so I need to re-regen.

PR #749, opened 2019-04-18
Base api service to be subclassed by autogen

mickjermsurawong-stripe:
This PR implements base api-service to be subclassed by the autogenerated services.
This is thin layer of abstraction to the existing `ApiResource.request` methods. 
Now it does only a few things.
- url variable encoding
- nullsafe ApiRequestParam

Example of this usage is in PR https://github.com/stripe/stripe-java/pull/743
eg. [`PaymentIntentService.java`](https://github.com/stripe/stripe-java/pull/743/files#diff-7e3d9d207a4be27423928689b2d0a611)

r? @brandur-stripe 
cc @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/ApiService.java
+++ b/src/main/java/com/stripe/net/ApiService.java
@@ -1,0 +1,72 @@
+package com.stripe.net;
+
+import com.stripe.Stripe;
+import com.stripe.exception.InvalidRequestException;
+import com.stripe.exception.StripeException;
+import com.stripe.model.StripeCollectionInterface;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.MissingFormatArgumentException;
+import java.util.Optional;
+
+/**
+ * Super class to services that make API calls. The service sub-class takes in
+ * {@link ApiRequestParams} and returns {@link ApiResource} or
+ * {@link com.stripe.model.StripeCollection} similarly to the methods on the
+ * resource itself.
+ */
+public abstract class ApiService {
+  /**
+   * Helper class to handle input null params, so users can still conveniently pass null instead of
+   * invoking an empty builder.
+   */
+  private static class ApiServiceNullParams extends ApiRequestParams {
+  }
+
+  /**
+   * Build a resource url given the url format (containing `%s`) and url variables.
+   * The url variables will be UTF-8 encoded.
+   * @param urlFormat     standard string format to be used with {@link String#format}
+   * @param urlVariables  un-encoded url variables as arguments to the url format.
+   * @return url for the resource
+   */
+  protected String resourceUrl(String urlFormat, String... urlVariables)
+      throws InvalidRequestException {
+    List<String> urlVariableList = new ArrayList<>();
+    for (String variable : Arrays.asList(urlVariables)) {
+      urlVariableList.add(ApiResource.urlEncodeId(variable));
+    }
+
+    String format;
+    try {
+      format = String.format(urlFormat, urlVariableList.toArray());
+    } catch (MissingFormatArgumentException ex) {
+      throw new InvalidRequestException(String.format(
+          "Unable to create url with pattern `%s` with url variables `%s`. "
+              + "This is likely a problem with this current library version `%s`. "
+              + "Please contact support@stripe.com for assistance.",
+          urlFormat, urlVariableList, Stripe.VERSION),
+          null, null, null, 0, ex);
+    }
+
+    return String.format("%s%s", Stripe.getApiBase(), format);
+  }
+
+  protected static <T> T request(ApiResource.RequestMethod method,
+                              String url, ApiRequestParams params, Class<T> clazz,
+                              RequestOptions options) throws StripeException {
+    return ApiResource.request(method, url, nullSafeParams(params), clazz, options);
+  }
+
+  protected static <T extends StripeCollectionInterface<?>> T requestCollection(
+      String url, ApiRequestParams params, Class<T> clazz, RequestOptions options)
+      throws StripeException {
+    return ApiResource.requestCollection(url, nullSafeParams(params), clazz, options);
+  }
+
+  private static ApiRequestParams nullSafeParams(ApiRequestParams params) {
+    return Optional.ofNullable(params).orElse(new ApiServiceNullParams());
+  }
+}
--- a/src/main/java/com/stripe/net/ApiService.java
+++ b/src/main/java/com/stripe/net/ApiService.java
@@ -54,12 +54,39 @@ public abstract class ApiService {
     return String.format("%s%s", Stripe.getApiBase(), format);
   }
 
+  /**
+   * Make an api request with a valid url like that from {@code resourceUrl}. This method is more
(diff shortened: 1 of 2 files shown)


src/test/java/com/stripe/net/ApiServiceTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @Override
+    public String getId() {
+      return FOO_ID;
+    }
+  }
+
+  private static class FooCollection extends StripeCollection<Foo> {
+  }
+
+
```
brandur-stripe: Nit: consider tightening up the whitespace here.

src/test/java/com/stripe/net/ApiServiceTest.java
```diff
@@ -0,0 +1,176 @@
...
+  private static class FooParams extends ApiRequestParams {
+  }
+
+  private static class FooOnBarParams extends ApiRequestParams {
+  }
+
+  private static class FooService extends ApiService {
```
brandur-stripe: Were an omit and retrieve omitted here intentionally (I suspect that yes because they're trivial on top of the other method types, but just wanted to make sure).

src/main/java/com/stripe/net/ApiService.java
```diff
@@ -0,0 +1,72 @@
...
+              + "This is likely a problem with this current library version `%s`. "
+              + "Please contact support@stripe.com for assistance.",
+          urlFormat, urlVariableList, Stripe.VERSION),
+          null, null, null, 0, ex);
+    }
+
+    return String.format("%s%s", Stripe.getApiBase(), format);
+  }
+
+  protected static <T> T request(ApiResource.RequestMethod method,
```
brandur-stripe: Just given that you've done such a good job of documenting the other `protected` method, maybe consider documenting this one and `requestCollection` as well.

ob-stripe:
@mickjermsurawong-stripe Have you given any thoughts to using generic interfaces, like we do in stripe-dotnet? I think it would be very nice to have this here, especially since Java 8 has spport for default methods. We could have something like this:

```java
public interface Creatable<TResource, TParams>
{
    default TResource create(TParams params, RequestOptions options)
    {
        return request(ApiResource.RequestMethod.POST, url, params, TParams.class, options);
    }
}

public class FooService extends ApiService implements Creatable<Foo, FooCreateParams>
{
}

// FooService gets create() automatically \o/
```

... although `TParams.class` won't work, and it looks like actually getting the generic parameter's type at runtime is non-trivial in Java due to type erasure :shakes_fist: :java:

mickjermsurawong-stripe:
@ob-stripe thank you for raising this! I didn't explore the idea fully, but now that you mention I just got to ^^"

Ah just did a quick implementations, and we might get stuck on resource with multiple parents, and now we will overload at the params, the following isn't legal:
```
public class CardService extends ApiService implements 
Creatable<Card, CardCreateOnBankAccount>, 
Creatable<Card, CardCreateOnCustomer> 
```
The parametrized Creatable isn't distinct to Java. It runs into the same compiler as a class 
```
class Foo implements HasId, HasId {}
```

- This approach is also not legal. "Both methods have the same erasure"
```
public interface Creatable<A, B, C> {

  default A delete(String account, String id, B params, RequestOptions options)
      throws StripeException {
    return null;
  }
  
  default A delete(String account, String id, C params, RequestOptions options)
      throws StripeException {
    return null;
  }
}
```

- We could do this though.. 
```
interface Createable1<R, P>{}
interface Createable2<R, P>{}
```

In general, I think the code would look really nicer, but maybe it's less explicit. I suspect seeing clear url and seeing http methods used be of value to new users interacting with our lib.

I think we might still want to have distinct methods (not using default method in the interface) because of java doc. And when considering we already have to implement custom methods as distinct methods, these two ways of defining API methods on our class seem like more work on the generator (more maintenance cost).

I'm still inclined to generate explicit methods here. But would appreciate more of your input!

mickjermsurawong-stripe:
^ java 11 keeps failing
```
Couldn't determine a download url for 11-GPL on linux-x64
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk11" --workspace "/home/travis/.cache/install-jdk" --feature "11" --license "GPL" --cacerts" failed and exited with 1 during .
```

src/test/java/com/stripe/net/ApiServiceTest.java
mickjermsurawong-stripe: I wasn't as intentional about testing all methods, but wanted to cover both request(POST/DELETE) and request collection(GET).

I've added more methods still to cover the 5 standard methods plus the custom one.

src/main/java/com/stripe/net/ApiService.java
mickjermsurawong-stripe: Add the docs to both.

ob-stripe:
@mickjermsurawong-stripe Ah damn, the `Creatable<A, B, C>` pattern is the one I had in mind, but I didn't realize that it wouldn't work because of dang type erasure 😢 That said, it looks like in #743, instead of the type overload you went with a `PaymentSourceService` + `ExternalAccountService` approach. I'm not sure which is best, but with that approach we don't need the overloads, right?

Regarding default implementations, that's a great point about Javadoc, and since the code is going be to autogen'd DRY is not particularly an issue, so let's forget about that. I think there would still be value in having interfaces for common methods though, as it makes the library a bit easier to mock and test.

mickjermsurawong-stripe:
For the legacy bank account and card there will still be param overloading.
I intend to match what we have in current object-oriented way
- `Card` -> `CardService`
  - `delete` and `update` (with param overloading)
- `BankAccount` -> `BankAccountService`
  - `delete`, `update`,  (with param overloading) and `verify`.
- `ExternalAccountCollection` + `ExternalAccount` (interface) -> `ExternalAccountService` 
  - `list`, `retrieve`, `create` (from collection method) `update`, `delete` (from interface - legacy of breaking down abstract ExternalAccount class to interface)
- `PaymentSourceCollection` -> `PaymentSourceService`: 
  - `list`, `retrieve`, `create`.

The motivation for keeping the polymorphic service is that we can't loose ability to list all subtypes within one call (eg. get all external accounts for this account). And for update/delete/create (now taking only string and not raw pan) users might not know before hand what the token represent (card/bank account) so they might simply use the polymorphic service. Do you think this is reasonable? But we can continue the discussion on this `sdk-autogen-java` PR 68.

Ah sorry.. I'm not sure I understand fully the benefits of mock/test interface for each crud method. But I'd love to discuss more off this PR too. I assume this common interface considerations shouldn't block this PR here, right?

mickjermsurawong-stripe:
ptal @brandur-stripe 
I addressed feedback on doc and test completeness. 
This PR won't be merged yet; I renamed to [wip]

ob-stripe:
@mickjermsurawong-stripe Ah yep, good point about being able to retrieve all polymorphic types in a single call 👍

And yep, the interface stuff can be added later if we agree it's valuable, definitely not a blocker here.

mickjermsurawong-stripe:
Rebasing on the integration branch `autogen/service`
Pulling this in. Thank you brandur and ob!

PR #751, opened 2019-04-23
[wip] Uses GSON to flatten extra params

mickjermsurawong-stripe:


src/main/java/com/stripe/net/ApiRequestParams.java
```diff
@@ -70,6 +80,116 @@ public EnumParam read(JsonReader in) {
...
+    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+      if (!HasExtraParams.class.isAssignableFrom(type.getRawType())) {
+        return null;
+      }
+
+      // we actually want delegator of the subtype of `HasExtraParams`
+      final TypeAdapter<T> delegate = gson.getDelegateAdapter(this, type);
+      TypeAdapter<T> extraParams = new TypeAdapter<T>() {
+        @Override
+        public void write(JsonWriter out, T value) throws IOException {
```
mickjermsurawong-stripe: Core logic of flattening

src/main/java/com/stripe/net/ApiRequestParams.java
```diff
@@ -70,6 +80,116 @@ public EnumParam read(JsonReader in) {
...
+
+    private void writeJsonElement(JsonWriter jsonWriter, JsonElement element) throws IOException {
+      if (element.isJsonNull()) {
+        // this is needed because the json delegated at this point would only have null
+        // due to empty enum
+        boolean oldSerializedNull = jsonWriter.getSerializeNulls();
+        jsonWriter.setSerializeNulls(true);
+        jsonWriter.nullValue();
+        jsonWriter.setSerializeNulls(oldSerializedNull);
+      } else if (element.isJsonObject()) {
```
mickjermsurawong-stripe: handling null as a result of the other type adapter converting empty param to null

src/test/java/com/stripe/param/common/HasExtraParamsSerializationTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Override
+    public Map<String, String> getExtraParam() {
+      return this.extraParams;
+    }
+
+    static Map<String, Object> expectedParams(String enumString) {
+      Map<String, Object> paramMap = new HashMap<>();
+      paramMap.put("string_value", "foo");
+      paramMap.put("enum_value", enumString);
+      paramMap.put("hello", "world");
```
remi-stripe: Make sure that it properly encodes non alphanumeric characters like `payment_intent_data[token]` for example

mickjermsurawong-stripe: Thank you @remi-stripe for the reminder!
Actually my preferred implementation here https://github.com/stripe/stripe-java/pull/752 will not require users to put in form encoding key.. The API looks like this:

```
    Map<String, Object> paymentIntentData = ImmutableMap.of(
        "application_fee_amount", 99L,
        "token", "src_123"
    );
    SessionCreateParams sessionCreateParams = SessionCreateParams.builder()
        .putExtraParam("payment_intent_data", paymentIntentData)
        .build();

    assertEquals(
        // nested form encoding from "extra params"
        "payment_intent_data[application_fee_amount]=99&payment_intent_data[token]=src_123",
        LiveStripeResponseGetter.createQuery(sessionCreateParams.toMap())
    );
```

mickjermsurawong-stripe: Instead of 
```
    SessionCreateParams sessionCreateParams = SessionCreateParams.builder()
        .putExtraParam("payment_intent_data[application_fee_amount]", 99L)
        .putExtraParam("payment_intent_data[token]", "src_123")
        .build();
```

remi-stripe: I strongly prefer the latter though. At that point you're already in unsupported territory and it just seems easier, and that's what we go in Go and .NET

mickjermsurawong-stripe: ah okay.. yup that's reasonable for consistency, and more constrained usage of this extra-params might be better than offering full-blown sub-hash extra params.

cc @ob-stripe if you have more thoughts

ob-stripe: How hard would it be to support both? I quite like being able to pass maps for nested parameters.

remi-stripe: If you like that, we should just go with this proposal in that case

mickjermsurawong-stripe: If we support passing maps with the interface `Map<String, Object>`, we can already support the string key/value with form encoding. So no additional work is required.

One good thing providing this extra params as map value is in the far future when we move to different encoding, we won't be breaking the users here. i.e choice of our lower-level encoding is not exposed to users.

ob-stripe: 👍 Yeah, agreed.

mickjermsurawong-stripe:
Closing in favor of serialization logic from JSON to untyped map: https://github.com/stripe/stripe-java/pull/752
We also agreed to support extra params of string key/ object value

PR #752, opened 2019-04-23
Support extra parameters in typed params

mickjermsurawong-stripe:
- This PR implements serialization logic for the "extra params"--to flatten the extra params to its outer object.
- There are two parts to this PR. 
  - First, modify the current `UntypedMapDeserializer` that converts JSON to untyped map to take a custom `Strategy`.
  - Second, implement the flattening logic.

- Params object will have the field `extraParams` generated, but what's needed is the `ApiRequestParams.EXTRA_PARAMS_KEY` serialized name as this solution works by flattening JSON to the the untyped map for form-encoding.
```
public class SkuCreateParams extends ApiRequestParams {
  // ...
  @SerializedName(ApiRequestParams.EXTRA_PARAMS_KEY)
  Map<String, Object> extraParams;
  // ...
}
```

- Alternative solution considered here is implementing this a GSON level itself https://github.com/stripe/stripe-java/pull/751/files. The sketch of solution is
  - Define interface `HasExtraParams` so that we can define type adapter to this specific type
  - Each param/sub param to implements this
  - Declare the `extraParam` map as `transient` so that it won't get serialized and we don't need to know the "magic" param key to remove it from JSON
  - Custom type adapter to flatten the JSON. 
    - This proves more involved than I wished it to be. The problem is writing out the value immediately closes the JSON such that we cannot smuggle in the extra params. 
    - To handle that, it needs a distinct writer to first get hold of the main JSON param object, add the extra params, and decode the JSON object back to the writer syntax (beginObject, endObject...)

- I prefer this approach because it 
  - introduces less code, and we can reason more easily without GSON.
  - more efficient than GSON solution (at least the one I had) which requires decode/encode for every sub-params

- Motivation here is that, in the service methods, we want to get rid of all untyped params, but we need support to add arbitrary sub-hash for beta feature. The api of this extra params look like https://github.com/stripe/stripe-java/pull/753
Where all param builder supports
```
public Builder putExtraParam(String key, Object value);
public Builder putAllExtraParam(Map<String, Object> map);
```
```
    SkuCreateParams.Inventory inventory = SkuCreateParams.Inventory.builder()
        .putExtraParam("type", "bucket")
        .build();
    // Map<String, Object> inventory  would work too

    Map<String, String> additionalAttributes = new HashMap<>();
    additionalAttributes.put("attr2", "val2");

    SkuCreateParams typedParams = SkuCreateParams.builder()
        .putExtraParam("active", true)
        .putExtraParam("price", 499L)
        .putExtraParam("currency", "usd")
        .putExtraParam("inventory", inventory)
        // allow both put/putAll
        .putAllExtraParam(ImmutableMap.of(
            "product", "prod_123",
            "image", "http://example.com/foo.png"
        ))
        .build();
```
    
r? @brandur-stripe @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/ApiRequestParams.java
+++ b/src/main/java/com/stripe/net/ApiRequestParams.java
@@ -1,16 +1,5 @@
 package com.stripe.net;
 
-import com.google.gson.FieldNamingPolicy;
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
-import com.google.gson.JsonObject;
-import com.google.gson.TypeAdapter;
-import com.google.gson.TypeAdapterFactory;
-import com.google.gson.reflect.TypeToken;
-import com.google.gson.stream.JsonReader;
-import com.google.gson.stream.JsonWriter;
-
-import java.io.IOException;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/stripe/net/ApiRequestParams.java
```diff
@@ -23,70 +12,34 @@
...
 
-        @Override
-        public EnumParam read(JsonReader in) {
-          throw new UnsupportedOperationException(
-              "No deserialization is expected from this private type adapter for enum param.");
-        }
-      };
-      return (TypeAdapter<T>) paramEnum.nullSafe();
-    }
-  }
```
mickjermsurawong-stripe: The empty enum handling is moved to `ApiRequestParamsConverter`

src/main/java/com/stripe/net/ApiRequestParamsConverter.java
```diff
@@ -0,0 +1,125 @@
...
+        outerMap.put(key, untypedMapDeserializer.deserializeJsonElement(value));
+      }
+    }
+  }
+
+  /**
+   * Type adapter to convert an empty enum to null value to comply with the lower-lever encoding
+   * logic for the API request parameters.
+   */
+  private static class HasEmptyEnumTypeAdapterFactory implements TypeAdapterFactory {
```
mickjermsurawong-stripe: Existing code ported over.

src/main/java/com/stripe/net/UntypedMapDeserializer.java
```diff
@@ -17,28 +17,95 @@
...
+    };
+  }
+
+  /**
+   * Deserializer with a custom strategy.
+   * @param strategy definition of how JSON element should be deserialized and set in its outer map.
+   */
+  UntypedMapDeserializer(Strategy strategy) {
+    this.strategy = strategy;
+  }
```
mickjermsurawong-stripe: package access. Now used only for this flattening JSON.
The above public method is exposed for `EventDataDeserializer` in model package

mickjermsurawong-stripe:
Thank you @ob-stripe for the review!
Yah GSON does have this `@JsonAdapter(SomeAdapter.class)`, so we could possibly annotate on the `extraParams`. But it suffers the same problem (serialized content is localized and flattening to outer context require too much code churn) as in this PR #753 trying to solve the problem from typed object to JSON.

This PR--solving it at JSON to untyped map--during the conversion will already loose context of the object class, so annotation will no longer be accessible.

mickjermsurawong-stripe:
Merging this into integration branch `autogen/service`. Thank you @ob-stripe 
re-
LGTM

PR #756, opened 2019-04-24
Form-encode collection as array using iterator implementation

mickjermsurawong-stripe:
- Fixes form encoding when we encounter collection object other than list. DX-3857
- This assumes the underlying collection has a deterministic order on the iterator. 
r? @ob-stripe 


mickjermsurawong-stripe:
thank you @ob-stripe!
@remi-stripe would you me to release this as 9.5.1 or we will include it with tax id?

remi-stripe:
@mickjermsurawong-stripe release separately I would say. But tests are actively failing

mickjermsurawong-stripe:
The failing test was the treeset in my test doesn't have have explicit type for comparison: I was comparing using Function.identity, but I fixed it to explicit `String:toCompare` instead.
Self-approving again here.

PR #758, opened 2019-04-24
[non-gen service] FileService

mickjermsurawong-stripe:
- This PR implements service for `File` which is non-gen.
- The first commit is changes in params: now contains `extraParams` map required in the service approach.
- The second is the service and its test. It contains hand-written modification for file creation.
- There will be two other follow-ups for `EphemeralKeyService` and `EventService`, but I would like to get this reviewed first in case there are cross-cutting concerns I can improve on the other two.
r? @ob-stripe 


src/main/java/com/stripe/service/FileService.java
```diff
@@ -0,0 +1,82 @@
...
+   * multipart/form-data</code>.
+   */
+  public File create(FileCreateParams params, RequestOptions options) throws StripeException {
+    String url = String.format("%s%s", Stripe.getUploadBase(), "/v1/files");
+    return multipartRequest(
+        ApiResource.RequestMethod.POST,
+        url,
+        Optional.ofNullable(params).orElse(FileCreateParams.builder().build()).toMap(),
+        File.class,
+        options);
```
mickjermsurawong-stripe: The hand-modification part to use multipart

src/main/java/com/stripe/param/FileCreateParams.java
```diff
@@ -14,6 +14,15 @@
   @SerializedName("expand")
   List<String> expand;
 
+  /**
```
mickjermsurawong-stripe: Only extra params are added here.
I removed the part where generated code removed custom handling of `File` or `InputStream` object

PR #759, opened 2019-04-24
[non-gen service] EphemeralKeyService

mickjermsurawong-stripe:
- This PR implements service for EphemeralKey which is non-gen.
- The first commit is changes in params: now contains extraParams map required in the service approach.
- The second is the service and its test. It contains hand-written modification for ephemeral key checking for override version in request options.
- This PR is similar to https://github.com/stripe/stripe-java/pull/758 for `FileService`
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/service/EphemeralKeyService.java
+++ b/src/main/java/com/stripe/service/EphemeralKeyService.java
@@ -1,0 +1,39 @@
+package com.stripe.service;
+
+import com.stripe.exception.StripeException;
+import com.stripe.model.EphemeralKey;
+import com.stripe.net.ApiResource;
+import com.stripe.net.ApiService;
+import com.stripe.net.RequestOptions;
+import com.stripe.param.EphemeralKeyCreateParams;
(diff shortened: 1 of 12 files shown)


src/main/java/com/stripe/service/EphemeralKeyService.java
```diff
@@ -0,0 +1,39 @@
...
+  public EphemeralKey create(EphemeralKeyCreateParams params) throws StripeException {
+    return create(params, (RequestOptions) null);
+  }
+
+  /** Creates a short-lived API key for a given resource. */
+  public EphemeralKey create(EphemeralKeyCreateParams params, RequestOptions options)
+      throws StripeException {
+    if (options.getStripeVersionOverride() == null) {
+      throw new IllegalArgumentException("`stripeVersionOverride` must be specified in "
+          + "RequestOptions with stripe version of your mobile client.");
```
mickjermsurawong-stripe: The non-standard create method, hand-written

PR #760, opened 2019-04-25
[non-gen service] EventService

mickjermsurawong-stripe:
- This PR implements service for Event which is non-gen.
- The first commit is changes in params: now contains extraParams map required in the service approach.
- The second is the service and its test. Unlike other file, and ephemeral key, event service doesn't have custom logic at the API methods.

This PR is similar to #758 and #759 for `FileService`, and `EphemeralKeyService`
r? @ob-stripe 
@stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/ApiRequestParams.java
+++ b/src/main/java/com/stripe/net/ApiRequestParams.java
@@ -1,16 +1,5 @@
 package com.stripe.net;
 
-import com.google.gson.FieldNamingPolicy;
-import com.google.gson.Gson;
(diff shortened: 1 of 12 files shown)


mickjermsurawong-stripe:
Yup that's correct @ob-stripe, we still don't have that granularity yet. It is fairly simple to implement--having two whitelists--but I suspect we are better off with a simpler logic of gen/non-gen for both model and service given that it only affects Event now.

ob-stripe:
👍  Thanks for confirming. I also think this is fine for now.

PR #764, opened 2019-04-25
[wip] example to reproduce lombok doc error

mickjermsurawong-stripe:
This PR illustrates example of problem with delomboked doc, that requires full class name in the {@link}


mickjermsurawong-stripe:
```
// Generated by com.stripe.generator.entity.SdkBuilder
package com.stripe.param.checkout;

import com.google.gson.annotations.SerializedName;
import com.stripe.net.ApiRequestParams;

public class Foo extends ApiRequestParams {

  public static class PaymentIntentData {

    public static class Shipping {

      public static class Address {
        @SerializedName("city")
        String city;
        @SerializedName("country")
        String country;

      
        public static class Builder {
          String city;
          String country;

          public Builder setCity(String city) {
            this.city = city;
            return this;
          }

          public Builder setCountry(String country) {
            this.country = country;
            return this;
          }
        }
      }
    }

    NOTE: De-lomboked doc from the above gets pulled to a different class thus causing error.
    /**
     * See documentation in {@link Address#city}.
     */
    public static class TransferData {
      @SerializedName("destination")
      String destination;
    }
  }

  /**
   * If specified, successful charges will be attributed to the destination account for tax
   * reporting, and the funds from charges will be transferred to the destination account. The
   * ID of the resulting transfer will be returned on the successful charge's `transfer` field.
   */
  public enum PaymentMethodType implements ApiRequestParams.EnumParam {
    @SerializedName("card")
    CARD("card");
    private final String value;

    PaymentMethodType(String value) {
      this.value = value;
    }

    @Override
    @java.lang.SuppressWarnings("all")
    @lombok.Generated
    public String getValue() {
      return this.value;
    }
  }
}
```

PR #765, opened 2019-04-25
Generated service and extra params

mickjermsurawong-stripe:
- This contains generated code from `a38dabc` for new classes of `ApiService`. 
  - The first commit only contains generated params with `extraParams`
  - The second is generated service

For ease of review, please see test with message `test: ...` for different resource/functionality patterns that we support.


src/test/java/com/stripe/service/CardServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+    return String.format("{\"id\": \"%s\", \"object\": \"card\", \"deleted\": true}", CARD_ID);
+  }
+
+  @Test
+  public void testUpdateOnCustomer() throws StripeException {
+    final Card updatedCard = cardService.update(CUSTOMER_ID,
+        CARD_ID,
+        CardUpdateOnCustomerParams.builder()
+            .putMetadata("key", "value")
+            .build());
```
mickjermsurawong-stripe: Previous:
```
Customer customer = Customer.retrieve("cus_Ex91Gk3tgiSOea");
Card card = (Card) customer.getSources().retrieve("card_1ETEjk2eZvKYlo2CenMzzj3H");
card.update(makeParams());
```

src/test/java/com/stripe/service/PaymentSourceServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+        params,
+        PaymentSourceCollection.class,
+        stubbedCollection.toJson()
+    );
+
+    final PaymentSourceCollection sources = paymentSourceService.list(
+        CUSTOMER_ID,
+        PaymentSourceListParams.builder()
+            .setObject("card")
+            .setLimit(1L).build());
```
mickjermsurawong-stripe: Previous:
```
PaymentSourceListParams params = makeParams();
PaymentSourceCollection sources = Customer.retrieve(CUSTOMER_ID)
  .getSources().list(params);
```

src/test/java/com/stripe/service/BankAccountServiceTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    stubRequest(
+        ApiResource.RequestMethod.POST,
+        String.format("/v1/accounts/%s/external_accounts/%s", ACCOUNT_ID, BANK_ACCOUNT_ID),
+        params.toMap(),
+        BankAccount.class,
+        getBankAccountFixture()
+    );
+
+    BankAccount updatedBankAccount = service.update(ACCOUNT_ID, BANK_ACCOUNT_ID, params);
```
mickjermsurawong-stripe: Previous:
```
Card card = doFetchCard();
// depending on whether card has field `customer` or `account`, different endpoints are activated
Card updatedCard = card.update(getParams());
```

src/test/java/com/stripe/service/PersonServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Test
+  public void testList() throws StripeException {
+    final Map<String, Object> params = new HashMap<>();
+    params.put("limit", 1);
+
+    final PersonCollection persons = personService.list(
+        ACCOUNT_ID,
+        PersonListParams.builder()
+            .setLimit(1L).build());
```
mickjermsurawong-stripe: Previous:
```
Account account = Account.retrieve(ACCOUNT_ID);
PersonCollection persons = account.persons(listParams());
```

src/test/java/com/stripe/service/SubscriptionItemServiceTest.java
```diff
@@ -0,0 +1,34 @@
...
+  private static final String ITEM_ID = "si_123";
+
+  private SubscriptionItemService subscriptionItemService = new SubscriptionItemService();
+
+  @Test
+  public void testUsageRecordSummaries() throws StripeException {
+    final UsageRecordSummaryCollection summaries = subscriptionItemService.usageRecordSummaries(
+        ITEM_ID,
+        SubscriptionItemUsageRecordSummariesParams.builder()
+            .setLimit(1L).build());
```
mickjermsurawong-stripe: Previous:
```
SubscriptionItem item = SubscriptionItem.retrieve("si_DAyhPiOTjrdfTv");
UsageRecordSummaryCollection summaries = item.usageRecordSummaries(params);
```

src/test/java/com/stripe/service/LoginLinkServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+        params,
+        LoginLink.class,
+        getResourceAsString("/api_fixtures/login_link.json")
+    );
+
+    LoginLink link = loginLinkService.create(
+        ACCOUNT_ID,
+        LoginLinkCreateParams.builder()
+            .setRedirectUrl("url")
+            .build());
```
mickjermsurawong-stripe: Previous:
```
LoginLink link = LoginLink.createOnAccount(ACCOUNT_ID, params());
```

src/test/java/com/stripe/service/PaymentIntentServiceTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+class PaymentIntentServiceTest extends BaseStripeTest {
+  private static final String PAYMENT_INTENT_ID = "pi_123";
+
+  private PaymentIntentService paymentIntentService = new PaymentIntentService();
+
+  @Test
+  public void testCapture() throws StripeException {
+    final PaymentIntent capturedPaymentIntent = paymentIntentService
+        .capture(PAYMENT_INTENT_ID, PaymentIntentCaptureParams.builder().build());
```
mickjermsurawong-stripe: Previous:
```
PaymentIntent intent = PaymentIntent.retrieve(PAYMENT_INTENT_ID);
PaymentIntent captured = intent.capture(params());
```

mickjermsurawong-stripe:
Hi @graj-stripe , @jd-stripe 

We are revamping our `stripe-java` client libraries, and we are changing from having methods on object instance or its own class (non-service approach) to a service approach. Within our team, we think it is a way forward, but would like to get a second confirmation from java developer perspective.

Specifically, do you find this the service approach, having separation of model and API request methods in stateless service object for our client library, a common and reasonable pattern in the java world? This contrasts with our current approach of model class implementing logic of making API requests. 

Previous:
```
class ResourceX {
   String id;
   
   static ResourceX retrieve(String)
   ResourceX update(Params)  
}

String myId = fromMyDatabase();
// dispatches GET `/v1/resource_x/id`
ResourceX fetched = ResourceX.retreive(myId);  

// dispatches POST `/v1/resource_x/id`,
// where id value in url comes from the instance's own variable
ResourceX updated = fetched.update(params());
```
Now:
```
class XService {
  ResourceX retrieve(String);
  ResourceX update(String, Params):  
}
String myId = fromMyDatabase();
// directly dispatches POST `/v1/resource_x/my_id`
ResourceX updated = fetched.update(myId, params());
```

In non-service approach, the problem is with updating resource so we have to first `GET` before another call `POST` to do the update.

graj-stripe:
> Specifically, do you find this the service approach, having separation of model and API request methods in stateless service object for our client library, a common and reasonable pattern in the java world?

Yeah I think this separation makes it much easier to write unit tests as well - e.g. then all you have to do is inject mock services into your application.

In the (previous) example the resource is also mutable, which we probably want to avoid. It looks like the service version returns immutable objects? If so, that's a big win as well.

I didn't really look at the review, but from the example this looks like a good improvement.

mickjermsurawong-stripe:
tagging @daz-stripe as well

mickjermsurawong-stripe:
Thank you @graj-stripe! Yup when we move fully into the service-world, we can create Stripe object that is immutable, but now as we are transitioning over, service still returns the same mutable object. 
So now services will only accept string IDs and params object, returning Stripe resource deserialized.

mickjermsurawong-stripe:
Thanks @daz-stripe for the thumbs-up!

Ah now I realize the "mutable part" after this evening discussion.
> (previous) example the resource is also mutable, which we probably want to avoid

We actually never mutate the object in resource-oriented approach now. Although we currently allow setters to the object thus mutable. Users could instantiate and set IDs to our instance and do instance method like:
```
    Card card = new Card();
    card.setId("card_123");
    card.setCustomer("cust_123");
    card.update(getParams());
```
In the service world, we can enforce immutability without exposing setters.

On testability, for instance method on card object, we could possibly do this the following. But it's not clear to me how we could insert this mock in the integration flow: what's the origin source of this mock?
```
    Card card = Mockito.mock(Card.class);
    Card updatedCard = Mockito.mock(Card.class);
    when(card.update(params)).thenReturn(updatedCard);
```
It could work if we can mock `Card.create` returning mock card. But we can't mock static method create to begin with. So I think service mockability is a huge win here still.

Now passing to @ob-stripe to review just for the flagged part in commits `test: ...` please.
r? @ob-stripe 
cc @remi-stripe

mickjermsurawong-stripe:
Unassigning @ob-stripe as we want more thoughts on the design.
cc @rattrayalex-stripe on the current implementation we have.
cc @stripe/api-libraries

mickjermsurawong-stripe:
Closing because we won't use this current version of autogen to generate service classes

PR #767, opened 2019-04-30
Validate extra params conflict

mickjermsurawong-stripe:
- This PR adds validation on passing extra params that conflict with the existing typed params.
- Passing conflicting params may likely be a problem if users think that the extra params that now support object value (instead of only string value) will get automatically merged with the typed params.
- Motivation here is also that we want to release adding extra params independently of the service method changes. 

Now the following pattern will get validation error.
```
// pseudo class schema
FooParams {
  Address {
    Billing {
      line_1: String
      // extra params to support.
      line_3: String
    }
  }
}

// extra params which will conflict with existing params from root-level
Map<String, Object> myLine3Map = ImmutableMap.of(
  "billing", ImmutableMap.of(
    "line_3", "my extra param line3"
  )
);

// will get error here
FooParams fooParams = FooParams.builder()
	.setAddress(myTypedAddressParams())
        .putExtraParams("address", myLine3Map).build();
```
The error message is:
```
Found multiple param values for the same param key. 
This can happen because you passed additional parameters via 
`putExtraParam` that conflict with the existing params. 
Found param key `address` with values `<conflicting value1>` and `<conflicting value2>`. 
If you wish to pass additional params for nested parameters, 
you should add extra params at the nested params themselves, 
not from the top-level param.
```

r? @ob-stripe 
cc @remi-stripe  @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
+++ b/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
@@ -12,8 +12,8 @@ import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
 
 import com.stripe.Stripe;
-
 import com.stripe.param.common.EmptyParam;
+
 import java.io.IOException;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


mickjermsurawong-stripe:
Merging to the integration branch for adding extra params instead of the `autogen/service`.
re-
LGTM

PR #769, opened 2019-05-04
[DONOTMERGE] CustomerBalanceTransaction support

remi-stripe:
This is a WIP PR to help review the other PRs and see fields added on other resources too.

cc @ob-stripe @mickjermsurawong-stripe


src/main/java/com/stripe/param/CustomerTransactionsParams.java
```diff
@@ -0,0 +1,124 @@
+// Generated by com.stripe.generator.entity.SdkBuilder
+
+package com.stripe.param;
+
+import com.google.gson.annotations.SerializedName;
+import com.stripe.net.ApiRequestParams;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CustomerTransactionsParams extends ApiRequestParams {
```
remi-stripe: Hmmm this class is not named properly I thiiiiink. Not yet sure how to fix it.

mickjermsurawong-stripe: remi, what's the expected param class name do you have in mind?

Currently, if we name method of listing of customer balance transactions as `transactions`, the param class has it as  Customer (resource name) + Transactions (method name) + Params.
```
public CustomerBalanceTransactionCollection transactions(CustomerTransactionsParams params) {}
```

The similar example here is `persons` method on `Account`
```
public PersonCollection persons(AccountPersonsParams params) {}
```

remi-stripe: I expected `TransactionParams` in that world I think

mickjermsurawong-stripe: I think having resource prefixing (Customer) is necessary to have consistent namespace for the params on a given resource. 
Depending on the final name we agreed, I would be more inclined to having `Customer` + method name still.

remi-stripe:
Closing as we have released support.

PR #784, opened 2019-05-29
Make headers access case insensitive

brandur:
Makes access to header names case insensitive so that we can easily
support HTTP/2 where the convention is to have downcased header names.
This allows us to access something like `Request-Id` regardless of the
backend protocol that's in use.

~~@ob-stripe The one sketchy thing I did here was using Java 8 streams
(because the code looks so good!). We're technically on Java 8+ so this
might be fine, but curious to hear if you have any thoughts on it.
Thanks!~~

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/StripeHeaders.java
+++ b/src/main/java/com/stripe/net/StripeHeaders.java
@@ -1,14 +1,34 @@
 package com.stripe.net;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 public class StripeHeaders {
 
-  Map<String, List<String>> headers;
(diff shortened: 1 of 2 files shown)


brandur-stripe:
So NM on the streams ... Cobertura didn't seem too happy with them and I had to add some code to accommodate some edges anyway, so I ended up just falling back to a regular loop.

ob-stripe:
LGTM, but you have some Checkstyle errors due to tab characters. I'll set up an EditorConfig file for the project.

ob-stripe:
ptal @brandur-stripe

brandur-stripe:
Thanks Olivier! My editor turned out to be severely misconfigured in pretty much every way, hah.

PR #786, opened 2019-05-30
Use google-java-format to format code

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Google provides a code formatting tool called [google-java-format](https://github.com/google/google-java-format). Since we're already using Google's Java style guide, we can use the tool to ensure consistent code formatting.

This PR can be reviewed commit by commit:
1. add the Gradle plugin to use google-java-format
2. run `./gradlew goJF` to automatically format all files
3. add `./gradlew verGJF` in Travis to fail builds if code is not correctly formatted
4. update `README.md` with instructions on how to run the formatter


ob-stripe:
Hrm, looks like it's not quite working at the moment -- Checkstyle doesn't pass, which is weird since the Checkstyle rules enforce the same style guide that google-java-format applies. Taking a look now.

brandur-stripe:
(Let me know when you want another look, but generally very +1 to the idea!)

ob-stripe:
Ok, so it looks like the answer is that Checkstyle is simply not compatible with google-java-format: https://github.com/google/google-java-format/issues/356#issuecomment-474343332.

I'm going to close this PR and open another one that replaces Checkstyle with Spotless as recommended in the linked thread.

PR #790, opened 2019-06-18
Stop testing OracleJDK

ob-stripe:
cc @stripe/api-libraries 

It looks like Travis no longer supports Oracle JDK 8 or 9 (the former fails with `Expected feature release number in range of 9 to 14, but got: 8`, the latter fails with a download error).

I'm not sure if this is transient or not, but I think we can safely stop testing Oracle JDK. AFAIK the only technical differences between the two are some very specific components (like OpenJDK Font Renderer and Oracle JDK Flight Recorder), but stripe-java doesn't use those at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,13 +17,12 @@ before_install:
 
 env:
   global:
-  - STRIPE_MOCK_VERSION=0.57.0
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Sorry about the delay, but LGTM as well.

PR #799, opened 2019-07-11
Add `getRawJsonObject()` accessor

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds a `getRawJsonObject()` accessor to `StripeObject` that returns a raw Gson `JsonObject`. Since this forces us to deserialize the JSON response a second time, the backing field is lazily initialized the first time the method is called.

### DIFF
--- a/src/main/java/com/stripe/model/StripeObject.java
+++ b/src/main/java/com/stripe/model/StripeObject.java
@@ -3,6 +3,8 @@ package com.stripe.model;
 import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonObject;
+import com.stripe.net.ApiResource;
 import com.stripe.net.StripeResponse;
 import java.lang.reflect.Field;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/model/StripeObject.java
```diff
@@ -34,12 +36,30 @@ public void setLastResponse(StripeResponse response) {
     this.lastResponse = response;
   }
 
+  /**
+   * Returns the raw JsonObject exposed by the Gson library. This can be useful to access properties
+   * that are not declared in the Stripe Java library.
+   *
```
brandur-stripe: Would you mind also adding a note here that explains that because this exposes some of the GSON API, it should not be considered fully stable, and that it's vastly preferable to use the standard static properties on a `StripeObject`. It's possible that a future version of stripe-java could move off of GSON and this method will either go away or change significantly.

ob-stripe:
Added a comment as request. Pulling this in.

ob-stripe:
Released as 10.6.0.

PR #804, opened 2019-07-16
Add support for unsetting metadata keys

### LINKED ISSUE #788: Can't unset a metadata key
Java: 8 
Stripe java library: 10.0.2

If I try to unset a metadata key as described [in this issue](https://github.com/stripe/stripe-java/issues/236), setting it to null using a HashMap, it does not have any effect on the value in the stripe dashboard.

I think is a stripe-java problem, because it works with curl.

Example code:

```
Map<String, String> metadata = new HashMap<>();
metadata.put("key", null);
CustomerUpdateParams customerParams = CustomerUpdateParams.builder().putAllMetadata(metadata).build()
Customer.retrieve(customer).update(customerParams);
```

ob-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

Add support for unsetting metadata keys, by encoding `null` values in maps.

Fixes #788. 

(I'm opening the PR but assigning the review to myself because @mickjermsurawong-stripe actually wrote the code.)

### DIFF
--- a/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
+++ b/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
@@ -5,6 +5,7 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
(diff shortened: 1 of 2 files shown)


ob-stripe:
This looks good to me.

Ideally, we would encode `null` values in all maps, not just `Map<String, String>` (since any value in a map parameter would have to be explicitly added by the user), but this would add further complexity and we don't have any parameters of this shape anyway.

Thanks again @mickjermsurawong-stripe! Much appreciated.

ob-stripe:
Released as 10.9.1.

PR #818, opened 2019-08-07
Set default timeouts in RequestOptionsBuilder

### LINKED ISSUE #817: RequestOptionsBuilder does not set default timeouts
The RequestOptionsBuilders defaults to 0 for both connectTimeout as well as readTimeout:
https://github.com/stripe/stripe-java/blob/aa9be6cf41aa1528bfc612a6920dc5f9073d2525/src/main/java/com/stripe/net/RequestOptions.java#L103-L104

This is a big variation from the default timeouts in RequestOptions generally, declared here:
https://github.com/stripe/stripe-java/blob/aa9be6cf41aa1528bfc612a6920dc5f9073d2525/src/main/java/com/stripe/Stripe.java#L9-L10

This is a pretty large foot-gun in the library right now. If this builder is used, but connectTimeout and readTimeout are not explicitly set, you've arrived at an infinite timeout for requests to Stripe. 
https://github.com/stripe/stripe-java/blob/8aebffd7b3d9a9ad953efd4bdd62bf5daea2c623/src/main/java/com/stripe/net/LiveStripeResponseGetter.java#L215
https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#setConnectTimeout-int-

Any outage in networking to Stripe (or some types of outages of Stripe) can lead to cascading failures due to thread exhaustion of a service making use of the Stripe Java client in this way, which is the default. This is surprising behavior.

Straightforward proposed fix - default RequestOptionsBuilder connectTimeout and readTimeout to the typical Stripe default timeouts.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Use the default values for the connect and read timeouts in `RequestOptionsBuilder`. Previously they defaulted to 0, i.e. no timeout.

Fixes #817.

### DIFF
--- a/src/main/java/com/stripe/Stripe.java
+++ b/src/main/java/com/stripe/Stripe.java
@@ -6,8 +6,8 @@ import java.util.HashMap;
 import java.util.Map;
 
 public abstract class Stripe {
-  private static final int DEFAULT_CONNECT_TIMEOUT = 30 * 1000;
-  private static final int DEFAULT_READ_TIMEOUT = 80 * 1000;
+  public static final int DEFAULT_CONNECT_TIMEOUT = 30 * 1000;
+  public static final int DEFAULT_READ_TIMEOUT = 80 * 1000;
 
   public static final String API_VERSION = "2019-05-16";
   public static final String CONNECT_API_BASE = "https://connect.stripe.com";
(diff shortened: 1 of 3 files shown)


ob-stripe:
Released as 10.12.2.

bpicolo:
@ob-stripe @brandur-stripe  Thanks for the extremely fast response on this one. : )

PR #824, opened 2019-08-14
[generated] source: spec3.sdk.yaml@spec-a15e00e in master

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries @shengwei-stripe 

- Add support for `au_becs_debit` sources
- Add support for `mandate` parameter in source creation and update requests
- Add support for `amount` in source update requests
- Add support for `pin` attribute on `issuing.card` objects
- Add support for `digital_goods_applications` category for Issuing

### DIFF
--- a/src/main/java/com/stripe/param/SourceUpdateParams.java
+++ b/src/main/java/com/stripe/param/SourceUpdateParams.java
@@ -4,6 +4,7 @@ package com.stripe.param;
 
 import com.google.gson.annotations.SerializedName;
 import com.stripe.net.ApiRequestParams;
+import com.stripe.param.common.EmptyParam;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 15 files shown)


src/main/java/com/stripe/param/SourceUpdateParams.java
```diff
@@ -4,13 +4,18 @@
...
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import lombok.Getter;
 
 public class SourceUpdateParams extends ApiRequestParams {
+  /** Amount associated with the source. */
+  @SerializedName("amount")
+  Long amount;
+
```
remi-stripe: Flagging this should also be in the changelog.

ob-stripe: 👍 Nice catch.

ob-stripe:
Released as 10.15.0.

PR #834, opened 2019-08-30
Make stripeError transient on StripeException

### LINKED ISSUE #832: StripeException implements Serializable, but its stripeError field is not serializable
`StripeException` implements `Serializable`, but it has a `stripeError` field (of type `StripeError`) that is not serializable, and is not `transient`.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Make `stripeError` transient on `StripeException`. This ensures that `StripeException` is serializable, as it implements the `Serializable` interface (all exceptions do) and `StripeObject`s are not serializable.

I briefly considered making `StripeObject`s serializable, but it seems like the general consensus is that Java's serialization mechanism is flawed and shouldn't be used in new code (https://stackoverflow.com/a/53889695/5307473) so this is the easiest fix.

Fixes #832.

### DIFF
--- a/src/main/java/com/stripe/exception/StripeException.java
+++ b/src/main/java/com/stripe/exception/StripeException.java
@@ -9,7 +9,8 @@ public abstract class StripeException extends Exception {
   private static final long serialVersionUID = 2L;
 
   /** The error resource returned by Stripe's API that caused the exception. */
-  @Setter StripeError stripeError;
+  // transient so the exception can be serialized, as StripeObject does not implement Serializable
+  @Setter transient StripeError stripeError;
 
   private String code;
   private String requestId;


brandur-stripe:
Wow, TIL. Sorry for the nit, but mind throwing a comment on there that indicates that the `transient` flag allows the class to be serializable because `StripeError` is not? Would never have understood that by looking at it if you hadn't explained it in the PR description.

brandur-stripe:
ptal @ob-stripe

ob-stripe:
> mind throwing a comment on there that indicates that the `transient` flag allows the class to be serializable because `StripeError` is not? 

Done!

ob-stripe:
Released as 11.5.1.

PR #839, opened 2019-09-03
Deprecate `total_count`

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Deprecate `total_count`.


ob-stripe:
> Approving though shouldn't this be in the autogen instead?

`StripeCollection` isn't autogenerated.

ob-stripe:
I pushed a second commit to use the newer `onMethod_` syntax that we're already using in autogenerated files, e.g. https://github.com/stripe/stripe-java/blob/4a7a241288cc626a6ef0142a1c3ec79ba5174857/src/main/java/com/stripe/model/CustomerBalanceTransaction.java#L65

ob-stripe:
Released as 11.6.1.

PR #852, opened 2019-10-08
UpdateParams: setters for string properties now all accept EmptyParam

richardm-stripe:
r? ob-stripe


ob-stripe:
Released as 12.2.0.

PR #857, opened 2019-10-13
Refactor form encoding

ob-stripe:
cc @stripe/api-libraries 

Refactors how the library handles form encoding, by moving all form encoding related methods out of `LiveStripeResponseGetter` and into a new utility class `FormEncoder`.

This is similar to the architecture we use in stripe-dotnet.

The code itself is also very similar to stripe-dotnet, but stripe-dotnet's code was also based on stripe-java, so all in all it's not too different. The new code is a bit more clean and supports encoding a few more types that we didn't before (`Date`, enums, arrays of primitives).

None of the methods i removed from `LiveStripeResponseGetter` were public, so this is not a breaking change.


ob-stripe:
r?+ @richardm-stripe

richardm-stripe:
With the "I am not a Java programmer" this seems sensible enough to me and the test suite seems thorough. Obviously passing something cyclical will make this blow up.
It's regrettable that "Object" is passed around so much. but I don't really see a way to avoid!

ob-stripe:
> Obviously passing something cyclical will make this blow up.

As obvious as it is, I hadn't considered it before 😂 I think that would qualify as a user error and we don't need to go out of our way to catch it, but at the same time it might not be too hard to track the references we've already seen while flatting the parameters to make sure we're not processing the same reference twice. I'll give it a try.

ob-stripe:
Merging this into a temporary integration branch for now.

PR #858, opened 2019-10-15
Stop propagating UnsupportedEncodingException

ob-stripe:
r? @remi-stripe @richardm-stripe 
cc @stripe/api-libraries 

`UnsupportedEncodingException` when passing `"UTF-8"` as the encoding name can literally never happen (UTF-8 is _always_ a valid encoding), so throwing a checked exception is useless.

Newer (>=10) Java versions have overloads for `URLEncoder.encode/decode` that directly accept a `Charset` instead of a string for the encoding's name and don't throw the checked exception, but since we still need to support Java 8 and 9, I'm doing the next best thing: catch the checked exception ASAP and replace it with a runtime (i.e. unchecked) exception, so callers don't have to worry about handling an exception that will never be thrown in practice.

Also replace a couple `"UTF-8"` literals with `StandardCharsets.UTF_8.name()`, just in case the world decides to rename UTF-8 to something else. Better safe than sorry.

### DIFF
--- a/src/main/java/com/stripe/net/ApiResource.java
+++ b/src/main/java/com/stripe/net/ApiResource.java
@@ -22,6 +22,7 @@ import com.stripe.model.StripeRawJsonObject;
 import com.stripe.model.StripeRawJsonObjectDeserializer;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Objects;
 
@@ -101,19 +102,7 @@ public abstract class ApiResource extends StripeObject {
 
   protected static String instanceUrl(Class<?> clazz, String id, String apiBase)
       throws InvalidRequestException {
-    try {
-      return String.format("%s/%s", classUrl(clazz, apiBase), urlEncode(id));
-    } catch (UnsupportedEncodingException e) {
-      throw new InvalidRequestException(
-          "Unable to encode parameters to "
-              + CHARSET
-              + ". Please contact support@stripe.com for assistance.",
-          null,
-          null,
-          null,
-          0,
-          e);
-    }
+    return String.format("%s/%s", classUrl(clazz, apiBase), urlEncode(id));
   }
 
   protected static String subresourceUrl(Class<?> clazz, String id, Class<?> subClazz)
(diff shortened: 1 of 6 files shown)


src/main/java/com/stripe/net/ApiResource.java
```diff
@@ -153,16 +129,22 @@ private static String subresourceUrl(Class<?> clazz, String id, Class<?> subClaz
...
+
+    try {
       // Don't use strict form encoding by changing the square bracket control
       // characters back to their literals. This is fine by the server, and
       // makes these parameter strings easier to read.
       return URLEncoder.encode(str, CHARSET).replaceAll("%5B", "[").replaceAll("%5D", "]");
+    } catch (UnsupportedEncodingException e) {
+      // This can literally never happen, and lets us avoid having to catch
+      // UnsupportedEncodingException in callers.
+      throw new AssertionError("UTF-8 is unknown");
```
remi-stripe: This worries me a bit. The previous version explicitly raised an Stripe exception that you could catch, and now you get an assert that crashes your application right? I don't really grasp why this is a win. You say it can never happen but we do regularly get devs with encoding errors, what if one of their plan or coupon id has an invalid character in their database and such?

ob-stripe: The `URLEncoder.encode/decode` methods throw an `UnsupportedEncodingException` if you pass an unknown encoding's name. The actual string you're trying to encode is not relevant here. The only case in which this could happen is if you pass an invalid name, and `"UTF-8"` is _always_ a valid name, in all Java environments (it's part of the [specification](https://docs.oracle.com/javase/6/docs/technotes/guides/intl/encoding.doc.html)).

In more recent Java versions, you can do this instead:

```java
String encoded = URLEncoder.encode(stringToEncode, StandardCharsets.UTF_8);
```

and this doesn't throw `UnsupportedEncodingException` because you're directly passing an encoding (not an encoding's name).

So yes, this assert would crash your app if it ever failed, but it will never fail.

remi-stripe: Ahhhhh yeah okay I see. Damn that was confusing. Reading that code I thought it would throw the same way if the string you passed is in the wrong encoding compared to what you expect. But I see StackOverflow agrees with you so all good!

PR #868, opened 2019-10-24
[codegen] Update API Resources

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

Enable new codegen for `Event`, `EventData`, `EventRequest` and related classes.

There should be no functional changes other than the addition of the `Event.retrieve(String id, EventRetrieveParams params, RequestOptions options)` method which didn't exist before for some reason, plus support for getters and extra params on the params classes.


src/main/java/com/stripe/model/EventRequest.java
```diff
@@ -8,18 +8,19 @@
...
+public class EventRequest extends StripeObject implements HasId {
   /**
-   * ID of the API request that caused the event. If null, the event was automatic (e.g., Stripe’s
-   * automatic subscription handling). Request logs are available in the <a
-   * href="https://dashboard.stripe.com/logs">dashboard</a>, but currently not in the API.
+   * ID of the API request that caused the event. If null, the event was automatic (e.g., Stripe's
+   * automatic subscription handling). Request logs are available in the
+   * [dashboard](https://dashboard.stripe.com/logs), but currently not in the API.
    */
+  @Getter(onMethod_ = {@Override})
```
remi-stripe: Minor but what does this one change?

src/main/java/com/stripe/param/EventListParams.java
```diff
@@ -242,18 +296,21 @@ public Builder addAllType(List<String> elements) {
...
+      this.extraParams = extraParams;
       this.gt = gt;
       this.gte = gte;
       this.lt = lt;
       this.lte = lte;
     }
 
     public static Builder builder() {
-      return new com.stripe.param.EventListParams.Created.Builder();
+      return new Builder();
```
remi-stripe: I don't fully grasp this change and if it can cause any issue so flagging

src/main/java/com/stripe/model/EventRequest.java
ob-stripe: Practically speaking, nothing.

`EventRequest` now implements the `HasId` interface, and the `getId()` signature is defined in the interface, so concrete implementations need to have the `@Override` annotation.

src/main/java/com/stripe/param/EventListParams.java
ob-stripe: It's the usual fully-qualified vs. short name. We're in a class named `Builder`, so  `Builder` refers to the class itself.

ob-stripe:
Released as 14.2.0.

PR #870, opened 2019-10-25
Refactor request telemetry

ob-stripe:
r? @richardm-stripe 
cc @stripe/api-libraries 

Refactor the request telemetry code into a single `RequestTelemetry` class.

Technically a breaking change because `ClientTelemetryPayload` and `RequestMetrics` were public, though I doubt this will affect many (or any) users.


src/test/java/com/stripe/functional/TelemetryTest.java
```diff
@@ -52,18 +54,30 @@ public void testTelemetryEnabled() throws StripeException, IOException, Interrup
...
-    assertEquals(payload1.lastRequestMetrics.requestId, "req_1");
-    assertTrue(payload1.lastRequestMetrics.requestDurationMs > 30);
+    JsonObject requestMetrics1 =
+        jsonParser
+            .parse(telemetry1)
+            .getAsJsonObject()
+            .get("last_request_metrics")
+            .getAsJsonObject();
+    String requestId1 = requestMetrics1.get("request_id").getAsString();
+    Long requestDurationMs1 = requestMetrics1.get("request_duration_ms").getAsLong();
```
ob-stripe: This is a little bit more awkward because we can no longer deserialize the JSON string into a `ClientTelemetryPayload` instance so we're using a raw `JsonObject` instead.

src/test/java/com/stripe/functional/TelemetryTest.java
```diff
@@ -92,14 +106,16 @@ public void testTelemetryDisabled() throws StripeException, IOException, Interru
...
+
     @Cleanup MockWebServer server = new MockWebServer();
 
     for (int i = 0; i < 20; i++) {
       server.enqueue(
           new MockResponse()
               .setBody("{}")
               .addHeader("Request-Id", "req_" + i)
-              .setBodyDelay(50, TimeUnit.MILLISECONDS));
+              .setBodyDelay(100, TimeUnit.MILLISECONDS));
```
ob-stripe: The test was a little bit flaky. This makes it a bit slower but hopefully more consistent.

PR #878, opened 2019-11-01
Add `StripeRequest` object

ob-stripe:
r? @brandur-stripe @richardm-stripe 

Adds a `StripeRequest` class used to hold the request's properties. This will eventually serve to standardize the interface between `LiveStripeResponseGetter` and `HttpClient`.

PR #879, opened 2019-11-01
Add `HttpClient` abstract class

ob-stripe:

~(Targets `ob-request-object` until #878 is merged. You probably want to review #878 first.)~

Renames `HttpClient` to `HttpURLConnectionClient` (because this client uses `java.net.HttpURLConnection` to make HTTP requests), and adds a new `HttpClient` abstract class.

This will make it easy to provide alternative implementations, e.g. to use a different HTTP library or to write a mock client.

I went with an abstract class rather than an interface because I expect that `HttpClient` will have some concrete methods. For instance, the request retry logic should be applicable regardless of the concrete HTTP client implementation.

The diff is hard to read because git/GitHub failed to pick up that `HttpClient` was renamed to `HttpURLConnectionClient` :/


richardm-stripe:
Is the hope here to enable *users* to substitute their favorite HTTP library? Or is it strictly for developers of stripe-java.

richardm-stripe:
Regardless,
lgtm

ob-stripe:
Thanks Richard!

> Is the hope here to enable users to substitute their favorite HTTP library? Or is it strictly for developers of stripe-java.

Both. There hasn't been a ton of demand for being able to use alternative HTTP libraries, but some users have been asking for an easier way to mock requests in their Stripe integration tests.

brandur-stripe:
Great to finally see some movement on this Olivier! Nice one.

PR #880, opened 2019-11-01
Stop disabling the DNS cache

ob-stripe:

Stop disabling the DNS cache by setting the `networkaddress.cache.ttl` property to `0`.

It was never clear why we did this in the first place ([the commit](https://github.com/stripe/stripe-java/commit/01fd2dba2ec1885f68c52355813eece3975ba3f1) that introduced this change provided zero context), and I don't think a library should be unilaterally modifying global properties.

Tagging this as a breaking change on the off chance that some users do rely on this behavior. For those I think we should simply advise them to modify the property themselves, like AWS does: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-jvm-ttl.html.


richardm-stripe:
lgtm
Completely agree.

PR #886, opened 2019-11-05
Codegen for openapi 28a5c69 + new API version

remi-stripe:
Move to the latest API version and add new changes
* Move to API version `2019-11-05`
* Add `default_settings` on `SubscritionSchedule`
* Remove `billing_thresholds` and `collection_method`, `default_payment_method` `default_source` and `invoice_settings` from `SubscriptionSchedule.
* Add `charge` filter when listing `Dispute`.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/SubscriptionSchedule.java
+++ b/src/main/java/com/stripe/model/SubscriptionSchedule.java
@@ -24,26 +24,10 @@ import lombok.Setter;
 public class SubscriptionSchedule extends ApiResource
     implements HasId, MetadataStore<SubscriptionSchedule> {
   /**
-   * Define thresholds at which an invoice will be sent, and the subscription advanced to a new
-   * billing period.
-   */
-  @SerializedName("billing_thresholds")
-  Subscription.BillingThresholds billingThresholds;
-
-  /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -69,23 +53,8 @@
...
-  /**
-   * ID of the default payment source for the subscription schedule. If not set, defaults to the
-   * customer's default source.
-   */
-  @SerializedName("default_source")
-  @Getter(lombok.AccessLevel.NONE)
-  @Setter(lombok.AccessLevel.NONE)
-  ExpandableField<PaymentSource> defaultSource;
+  @SerializedName("default_settings")
+  DefaultSettings defaultSettings;
```
brandur-stripe: I notice that a lot of fields in this file seem to have a docblock, in case you want to add one.

remi-stripe: yeah it's autogen-ed from the spec so not something I should add manually. But nice catch, I'll get the team to fix it

PR #891, opened 2019-11-08
Replace Cobertura with JaCoCo

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

So I just noticed that our coverage reports have been [broken](https://travis-ci.org/stripe/stripe-java/jobs/608348053#L2191) for months, because we're using a lambda expression ([here](https://github.com/stripe/stripe-java/blob/1fe663f48821b3a6e9f0444388bf3fd339447ecf/src/main/java/com/stripe/net/ApiRequestParamsConverter.java#L113)) and Cobertura doesn't support them (see https://github.com/cobertura/cobertura/issues/274).

Given that the issue is 4 years old and still not fixed, I've followed the advice of someone in the thread and migrated from Cobertura to JaCoCo, another code coverage tool.

(Our coverage numbers are still fairly abysmal because ~nothing in `com.stripe.param` is tested, but that's a different story.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
   - ./gradlew --version
   - ./gradlew clean
   - if [ "$TRAVIS_JDK_VERSION" == "openjdk8" ]; then ./gradlew check; fi
-  - ./gradlew -Djdk.tls.client.protocols="TLSv1.2" cobertura coveralls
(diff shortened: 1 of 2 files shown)


ob-stripe:
My guess is that the Cobertura plugin crashed in a way that didn't cause Gradle to stop the execution, and a partial report was still produced and uploaded to coveralls.io by the Coveralls plugin. I'm not sure though -- it's definitely surprising that Gradle didn't stop and return an error.

PR #892, opened 2019-11-08
Add Github Actions CI workflow

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

More Github Actions!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,53 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  spotless:
+    name: Spotless
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@master
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+  tests:
+    name: Tests
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
```
brandur-stripe: This doesn't look quite right ;)

I'm not sure what we're actually running against here, but I assume it must be whatever Java is available in the default environment. I think there should be at least a Java setup setup somewhere?

ob-stripe: 🤦‍♂ Nice catch.

ob-stripe:
ptal @brandur-stripe

PR #896, opened 2019-11-12
Add `HttpContent` class

ob-stripe:
cc @stripe/api-libraries 

This PR:
- adds a new `HttpContent` class that represents an HTTP request's content, i.e. the body and the `Content-Type` header (which can depend on the body, e.g. for multipart requests the header contains the boundary used to separate fields in the body).
- adds a new `FormEncoder.createHttpContent()` method that builds and returns an `HttpContent` instance as needed (automatically choosing between form-data encoding and multipart encoding depending on whether all the parameters are strings or not)
- removes the now useless `FormEncoder.encodeMultipartParams()` method
- updates `HttpURLConnectionClient` to use `HttpContent`. The resulting code is a little bit weird, but will be refactored in a future PR
- updates `MultipartProcessor` to write to a provided `OutputStream` instead of directly to a connection's `OutputStream`
- removes the `MultipartProcessor.getBoundary()` method (the boundary is now initialized by `HttpContent`, and is a random UUIDv4 unless explicitly provided)
- adds a bunch of tests to ensure the encoding is working as expected.

### DIFF
--- a/src/main/java/com/stripe/net/FormEncoder.java
+++ b/src/main/java/com/stripe/net/FormEncoder.java
@@ -1,14 +1,12 @@
 package com.stripe.net;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Array;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
-import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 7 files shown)


src/main/java/com/stripe/net/FormEncoder.java
```diff
@@ -18,6 +16,30 @@
...
+
+    // If all parameters have been encoded as strings, then the content can be represented
+    // with application/x-www-form-url-encoded encoding. Otherwise, use
+    // multipart/form-data encoding.
+    if (flatParams.stream().allMatch(kvp -> kvp.getValue() instanceof String)) {
+      Collection<KeyValuePair<String, String>> flatParamsString =
+          flatParams.stream()
+              .filter(kvp -> kvp.getValue() instanceof String)
+              .map(kvp -> new KeyValuePair<String, String>(kvp.getKey(), (String) kvp.getValue()))
+              .collect(Collectors.toList());
```
brandur-stripe: Man, I'm jealous of your Very Modern Java skills (skillz?) ;)

src/main/java/com/stripe/net/KeyValuePair.java
```diff
@@ -0,0 +1,11 @@
+package com.stripe.net;
+
+import java.util.AbstractMap;
+
+class KeyValuePair<K, V> extends AbstractMap.SimpleEntry<K, V> {
```
brandur-stripe: Would you mind adding a docblock here explaining why this is needed on top of `AbstractMap.SimpleEntry`? (I don't think I completely understand myself.)

src/main/java/com/stripe/net/FormEncoder.java
ob-stripe: 😂

src/main/java/com/stripe/net/KeyValuePair.java
ob-stripe: Strictly speaking, it's not needed, the class does not implement any behavior on top of `AbstractMap.SimpleEntry`. It's purely semantics: in the context of form encoding, we're dealing with "lists of key/value pairs", not "map entries".

I'll add a comment to clarify.

PR #897, opened 2019-11-12
Add `Stopwatch` class

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Add a `Stopwatch` class to measure elapsed time, and use it for latency telemetry.

The class is basically a port of [its namesake from the .NET stdlib](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch?view=netframework-4.8). It does a bit more than we need, but it's still fairly straightforward.

This will help further refactors of `HttpURLConnectionClient`.

### DIFF
--- a/src/main/java/com/stripe/util/Stopwatch.java
+++ b/src/main/java/com/stripe/util/Stopwatch.java
@@ -1,0 +1,110 @@
+package com.stripe.util;
+
+import java.time.Duration;
+
+/**
+ * This class provides a set of methods and properties that can be used to accurately measure
+ * elapsed time.
+ */
+public class Stopwatch {
+  private long elapsed;
+  private boolean running;
+  private long startTimeStamp;
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/util/Stopwatch.java
```diff
@@ -0,0 +1,109 @@
...
+   * Gets the total elapsed time measured by the current instance.
+   *
+   * @return a {@link Duration} representing the total elapsed time measured by the current instance
+   */
+  public Duration getElapsed() {
+    return Duration.ofNanos(this.getRawElapsed());
+  }
+
+  /**
+   * Gets the current timestamp with the highest precision available.
```
brandur-stripe: Given that you've got a comment on here already, it might be worthwhile just mentioning that the time is monotonic. I had to look it up to make sure.

PR #898, opened 2019-11-13
Move all request properties in `StripeRequest`

ob-stripe:
cc @stripe/api-libraries 

This PR is larger than I would have liked -- sorry about that.

Here's what it does:
- modify `StripeRequest`. Instead of being a dumb data class exposing its constructor arguments, it now transforms the constructor arguments (URL as a string, map of parameters, `RequestOptions` modifiers) into HTTP request concepts (URL as a `URL`, `HttpContent` for the request's body, map of HTTP headers)
- move some of the logic that is common to all HTTP clients (building the strings for the `User-Agent` and `X-Stripe-Client-User-Agent` headers, handling telemetry) out of `HttpURLConnectionClient` and into the abstract parent `HttpClient` class. Now `HttpURLConnectionClient` simply sends the HTTP request and receives the response and nothing else.
- set up a new `ReflectionUtils` class to host a couple of reflection-based methods. Those are used to initialize the custom `URLStreamHandler` that can be provided by users. There is no context as to why this feature was introduced in stripe-java and I have serious doubts that it's still useful. I've kept it for now, but might remove it (or rework it to not require ugly reflection) in a future PR.


src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -1,32 +1,168 @@
...
+    } else {
+      url = new URL(sb.toString());
+    }
+    return url;
+  }
+
+  private static HttpContent buildContent(
+      ApiResource.RequestMethod method, Map<String, Object> params) throws IOException {
+    if (method != ApiResource.RequestMethod.POST) {
+      return null;
```
richardm-stripe: Would it be better to throw an exception than silently omitting the content?

ob-stripe: I don't think we'd want that -- it's expected that non-POST requests have no bodies. If we threw an exception here we'd have to conditionally call `buildContent()`. I don't see much of an upside right now (but I have more refactors incoming so I'll keep it in mind).

ob-stripe:
Thanks Richard!

src/main/java/com/stripe/net/StripeRequest.java
richardm-stripe: Makes sense. I have some paranoia about the prospect of contentful non-posts due to past experience (not at Stripe) ;)

ob-stripe: Fun fact (for some definitions of "fun"): Stripe's API will happily accept parameters in both query strings and request bodies, regardless of the HTTP method being used 🙈

PR #899, opened 2019-11-13
Remove `ApiResource.RequestType`

ob-stripe:
cc @stripe/api-libraries 

Remove `ApiResource.RequestType`. It is no longer useful: we now decide whether a request should be encoded as form-data or multiform based on the parameters. If there is at least one `File` or `Stream` value in the parameters, then the request will be encoded as multiform, otherwise it's encoded as form-data.

This affects the `StripeResponseGetter` public interface, so it is a breaking change (though it would only affect users providing custom implementations).

### DIFF
--- a/src/main/java/com/stripe/model/File.java
+++ b/src/main/java/com/stripe/model/File.java
@@ -99,7 +99,7 @@ public class File extends ApiResource implements HasId {
    */
   public static File create(Map<String, Object> params, RequestOptions options)
       throws StripeException {
-    return multipartRequest(
+    return request(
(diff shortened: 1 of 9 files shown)


ob-stripe:
> I think GitHub was sending us some emails about being on a legacy billing plan, and that might have something to do with it. Do you know off hand?

I'm actually at the GitHub Universe convention today and talked to a GitHub employee (@francisfuzz) about this. They've just launched billing for GitHub Actions and we've apparently hit a bug. Hopefully they fix it soon 🤞

PR #903, opened 2019-11-15
Serialize null values in all maps

### LINKED ISSUE #893: Cannot remove metadata entries with typed / builder update params
Note: I've noticed that with issuing authorizations and transactions but this might be happening with other objects as well.

To try and remove a metadata entry (currently set to `"foo": "bar"`) I try the following:
```java
authorization.update(AuthorizationUpdateParams.builder()
        .putMetadata("foo", null)
        .build());
```
However this does not seem to make any change to the object (the metadata entry is still there with the old value).

I tried calling `.putMetadata("foo", "")` but the SDK rejects empty strings.

The workaround I found was to call:
```java
authorization.update(AuthorizationUpdateParams.builder()
        .putExtraParam("metadata[foo]", EmptyParam.EMPTY)
        .build());
```
which worked.

Should it be possible to remove metadata entries with `putMetadata`?

ob-stripe:
cc @stripe/api-libraries 

This PR serializes `null` values in all `Map`s when converting typed parameters classes to untyped maps.

Previously, we were only serializing `null` values in `Map<String, String>` maps. The intention was that only `metadata` fields had this type and that's where we wanted to serialize values.

However, after #849, some metadata fields (such as [`TransactionUpdateParams.metadata`](https://github.com/stripe/stripe-java/blob/49db11cf0ab6707e1c1bbb97f6fa70e8ebffbc13/src/main/java/com/stripe/param/issuing/TransactionUpdateParams.java#L28)) were no longer typed as `Map<String, String>` but rather as `Object` because they became polymorphic (either a `HashMap<String, String>` or an `EmptyParam`).

Due to type erasure, the type parameters are no longer present at runtime and the type adapter that was supposed to serialize null values failed to do so because [this check](https://github.com/stripe/stripe-java/blob/49db11cf0ab6707e1c1bbb97f6fa70e8ebffbc13/src/main/java/com/stripe/net/ApiRequestParamsConverter.java#L101) failed.

This PR simplifies the type adapter by simply checking if the value is a `Map` (or any subclass such as `HashMap`). When that's the case, the type adapter sets `serializeNulls(true)` and then delegates to the normal adapter for the type. As a result, all `null` values are now serialized in all `Map`s. I think that's what we want anyway: if an entry is present in a map, it was necessarily added by the user and should be serialized.

Fixes #893.

### DIFF
--- a/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
+++ b/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
@@ -5,7 +5,6 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-import com.google.gson.JsonParseException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
 import com.google.gson.reflect.TypeToken;
@@ -14,8 +13,6 @@ import com.google.gson.stream.JsonWriter;
 import com.stripe.Stripe;
 import com.stripe.param.common.EmptyParam;
 import java.io.IOException;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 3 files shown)


ob-stripe:
Released as v15.3.1.

PR #908, opened 2019-11-26
Add support for Preview endpoint on CreditNote

remi-stripe:
Codegen for openapi ab6898a

r? @ob-stripe 
cc @stripe/api-libraries


src/main/java/com/stripe/model/CreditNote.java
```diff
@@ -317,6 +318,32 @@ public static CreditNote create(CreditNoteCreateParams params, RequestOptions op
...
+        ApiResource.RequestMethod.GET, url, params, CreditNote.class, options);
+  }
+
+  /** Get a preview of a credit note without creating it. */
+  public static CreditNote preview(CreditNotePreviewParams params) throws StripeException {
+    return preview(params, (RequestOptions) null);
+  }
+
+  /** Get a preview of a credit note without creating it. */
+  public static CreditNote preview(CreditNotePreviewParams params, RequestOptions options)
```
remi-stripe: Should all of those be named `previewCreditNote` since that's how we do it for void?

remi-stripe: we have gaveled not to rename

ob-stripe:
Released as 15.7.0.

PR #910, opened 2019-12-03
Use global timeouts in RequestOptions

### LINKED ISSUE #909: Global Timeout not used.
Class `RequestOptionsBuilder` in the constructor 
`public RequestOptionsBuilder() {
this.apiKey = Stripe.apiKey;
            this.clientId = Stripe.clientId;
            this.connectTimeout = 30000;
            this.readTimeout = 80000;
        }`
should read set connectTimeout and readTimeout using the get methods (`getConnectTimeout` and `getReadTimeout`) on `Stripe` class

ob-stripe:
cc @stripe/api-libraries 

Fixes #909.

### DIFF
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -110,8 +110,8 @@ public class RequestOptions {
     public RequestOptionsBuilder() {
       this.apiKey = Stripe.apiKey;
       this.clientId = Stripe.clientId;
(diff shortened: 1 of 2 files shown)


remi-stripe:
Lurk, isn't this a breaking change in theory? We're "suddenly" changing the timeouts though I can agree that we're doing the right thing.

There's a major going out **today** so we should still bundle this one in the major coming instead.

ob-stripe:
Not a breaking change IMO, but if we're releasing a major today we might as well bundle this.

PR #918, opened 2019-12-19
Multiple API changes

remi-stripe:
Codegen for openapi 4ec3068

* Adds `livemode` and `metadata` to Terminal `Reader` and `Location`
* Adds `ms` as a valid locale on Checkout `Session`

r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/param/issuing/CardCreateParams.java
+++ b/src/main/java/com/stripe/param/issuing/CardCreateParams.java
@@ -2,8 +2,6 @@ package com.stripe.param.issuing;
 
 import com.google.gson.annotations.SerializedName;
 import com.stripe.net.ApiRequestParams;
-import com.stripe.net.ApiRequestParams.EnumParam;
-import com.stripe.param.common.EmptyParam;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -59,7 +57,7 @@ public class CardCreateParams extends ApiRequestParams {
   @SerializedName("shipping")
   Shipping shipping;
 
-  /** Whether authorizations can be approved on this card. */
+  /** Whether authorizations can be approved on this card. Defaults to `inactive`. */
   @SerializedName("status")
   Status status;
 
@@ -256,7 +254,7 @@ public class CardCreateParams extends ApiRequestParams {
       return this;
     }
 
-    /** Whether authorizations can be approved on this card. */
+    /** Whether authorizations can be approved on this card. Defaults to `inactive`. */
     public Builder setStatus(Status status) {
       this.status = status;
       return this;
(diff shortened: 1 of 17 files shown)


src/main/java/com/stripe/param/issuing/CardCreateParams.java
```diff
@@ -3392,7 +3389,7 @@ public Builder setType(Type type) {
       }
 
       /** Packaging options. */
-      public Builder setType(EmptyParam type) {
+      public Builder setType(String type) {
```
richardm-stripe: What's this about?

src/main/java/com/stripe/param/issuing/CardholderCreateParams.java
```diff
@@ -66,7 +66,7 @@
   String phoneNumber;
 
   /**
-   * Specifies whether to permit authorizations on this cardholder's cards. Defaults to `inactive`.
+   * Specifies whether to permit authorizations on this cardholder's cards. Defaults to `active`.
```
richardm-stripe: This seems like a breaking change?

richardm-stripe: Is this accurate?

richardm-stripe:
Just a couple questions. The advertised changes seemed all in order

src/main/java/com/stripe/param/issuing/CardholderCreateParams.java
remi-stripe: The docs had been wrong for a year!

src/main/java/com/stripe/param/issuing/CardCreateParams.java
remi-stripe: That one I'm not sure, I think it's because they changed the way string/enums work in their API. cc @ob-stripe

ob-stripe: Internally, the `type` parameter was changed from a `StringParam` with a set of `values` to our more modern `APIEnumParam ` abstraction with `allow_all_strings_in_schema` set to `true` to avoid a breaking change.

In the OpenAPI spec, this caused `type` to be updated from a `string` with two possible `enum` values to a polymorphic `anyOf` type: either a `string` with two possible `enum` values (i.e. the previous definition), or a simple `string` (due to `allow_all_strings_in_schema: true`).

So it's expected that we now have two overloads for `type` (one with the `Type` enum type and one with `String`). I can't figure out why we had an overload with `EmptyParam` before though, since that was not a valid value for this parameter.

This is technically a breaking change, but I think it's fine to release this as a minor since I'm fairly sure using `EmptyParam` would have caused the API request to fail anyway.

PR #922, opened 2020-01-06
Codegen for openapi c896d1f

rattrayalex-stripe:
r? @ob-stripe 
cc @remi-stripe @richardm-stripe @cjavilla-stripe 
cc @stripe/api-libraries 

Commits are likely best reviewed individually.

### DIFF
--- a/src/main/java/com/stripe/model/Source.java
+++ b/src/main/java/com/stripe/model/Source.java
@@ -6,6 +6,7 @@ import com.stripe.exception.StripeException;
 import com.stripe.net.ApiResource;
 import com.stripe.net.RequestOptions;
 import com.stripe.param.SourceCreateParams;
+import com.stripe.param.SourceDetachParams;
 import com.stripe.param.SourceRetrieveParams;
 import com.stripe.param.SourceSourceTransactionsParams;
 import com.stripe.param.SourceUpdateParams;
@@ -207,6 +208,25 @@ public class Source extends ApiResource implements MetadataStore<Source>, Paymen
 
   /** Delete a specified source for a given customer. */
   public Source detach(Map<String, Object> params, RequestOptions options) throws StripeException {
+    String url =
+        String.format(
+            "%s%s",
+            Stripe.getApiBase(),
+            String.format(
+                "/v1/customers/%s/sources/%s",
+                ApiResource.urlEncodeId(this.getCustomer()),
+                ApiResource.urlEncodeId(this.getId())));
+    return ApiResource.request(
+        ApiResource.RequestMethod.DELETE, url, params, Source.class, options);
+  }
+
+  /** Delete a specified source for a given customer. */
+  public Source detach(SourceDetachParams params) throws StripeException {
+    return detach(params, (RequestOptions) null);
+  }
+
+  /** Delete a specified source for a given customer. */
+  public Source detach(SourceDetachParams params, RequestOptions options) throws StripeException {
     String url =
         String.format(
             "%s%s",
--- a/src/main/java/com/stripe/param/AccountUpdateParams.java
+++ b/src/main/java/com/stripe/param/AccountUpdateParams.java
@@ -227,6 +227,12 @@ public class AccountUpdateParams extends ApiRequestParams {
       return this;
     }
 
+    /** The business type. */
+    public Builder setBusinessType(EmptyParam businessType) {
+      this.businessType = businessType;
+      return this;
+    }
+
     /**
      * Information about the company or business. This field is null unless `business_type` is set
      * to `company`.
--- a/src/main/java/com/stripe/param/AccountUpdateParams.java
+++ b/src/main/java/com/stripe/param/AccountUpdateParams.java
@@ -22,7 +22,7 @@ public class AccountUpdateParams extends ApiRequestParams {
   @SerializedName("business_profile")
   BusinessProfile businessProfile;
 
-  /** The business type. Can be `individual` or `company`. */
+  /** The business type. */
   @SerializedName("business_type")
   Object businessType;
 
@@ -215,14 +215,14 @@ public class AccountUpdateParams extends ApiRequestParams {
       return this;
     }
 
-    /** The business type. Can be `individual` or `company`. */
-    public Builder setBusinessType(String businessType) {
+    /** The business type. */
+    public Builder setBusinessType(BusinessType businessType) {
       this.businessType = businessType;
       return this;
     }
 
-    /** The business type. Can be `individual` or `company`. */
-    public Builder setBusinessType(EmptyParam businessType) {
+    /** The business type. */
+    public Builder setBusinessType(String businessType) {
       this.businessType = businessType;
       return this;
     }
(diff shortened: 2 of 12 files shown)


richardm-stripe:
r? @remi-stripe 
this is in sync with main repo now

remi-stripe:
@richardm-stripe It's incorrect though. It needs to be regenerated with the latest openapi spec (that needs to be updated first). Happy to take if you prefer

richardm-stripe:
r? @remi-stripe 
Updated openapi https://github.com/stripe/openapi/commit/c896d1f0e5b89cee6b3e9d7a0c6837b27a791bb7 and synced the changes here:

src/main/java/com/stripe/model/Source.java
```diff
@@ -219,6 +220,25 @@ public Source detach(Map<String, Object> params, RequestOptions options) throws
...
+        String.format(
+            "%s%s",
+            Stripe.getApiBase(),
+            String.format(
+                "/v1/customers/%s/sources/%s",
+                ApiResource.urlEncodeId(this.getCustomer()),
+                ApiResource.urlEncodeId(this.getId())));
+    return ApiResource.request(
+        ApiResource.RequestMethod.DELETE, url, params, Source.class, options);
+  }
```
remi-stripe: @ob-stripe confirming that we're okay adding that endpoint to `Source` instead of requiring they go through the Delete on the Customer instead?

src/main/java/com/stripe/param/PaymentIntentCreateParams.java
```diff
@@ -51,18 +43,6 @@
...
-   *
-   * <p>When the confirmation method is `automatic`, a PaymentIntent can be confirmed using a
-   * publishable key. After `next_action`s are handled, no additional confirmation is required to
-   * complete the payment.
-   *
-   * <p>When the confirmation method is `manual`, all payment attempts must be made using a secret
-   * key. The PaymentIntent returns to the `requires_confirmation` state after handling
-   * `next_action`s, and requires your server to initiate each payment attempt with an explicit
-   * confirmation.
-   */
```
remi-stripe: One of those "we lost the basic string documentation". Sigh.

src/main/java/com/stripe/param/SubscriptionCancelParams.java
```diff
@@ -2,12 +2,18 @@
...
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import lombok.Getter;
 
 @Getter
 public class SubscriptionCancelParams extends ApiRequestParams {
+  /** Specifies which fields in the response should be expanded. */
+  @SerializedName("expand")
+  List<String> expand;
```
remi-stripe: Do we know why `expand` is suddenly appearing here?

src/main/java/com/stripe/param/issuing/CardCreateParams.java
```diff
@@ -3392,7 +3389,7 @@ public Builder setType(Type type) {
       }
 
       /** Packaging options. */
-      public Builder setType(EmptyParam type) {
+      public Builder setType(String type) {
```
remi-stripe: I don't fully grasp this change. We had an enum already, why now `Object` and also why did `setType` lose the `EmptyParam` one?

src/main/java/com/stripe/model/Source.java
ob-stripe: `Source.detach` already existed, we just lacked the overloads for typed parameters.

src/main/java/com/stripe/param/PaymentIntentCreateParams.java
ob-stripe: I guess this is because we're not documenting the individual enum values: https://stripe.com/docs/api/payment_intents/create#create_payment_intent-confirmation_method.

Not worth blocking this release, but we should come up with a better solution for client libs to document both the field and the possible values.

src/main/java/com/stripe/param/issuing/CardCreateParams.java
richardm-stripe: I also think this is breaking

src/main/java/com/stripe/param/SubscriptionCancelParams.java
ob-stripe: A deliberate change by (I think) @rattrayalex-stripe in this commit: https://github.com/stripe/stripe-java/pull/922/commits/8028045a7147b957057f87aca742bdb76e6839b4.

Some DELETE endpoints return a full object (subscription cancel, source detach, maybe others) so this lets users expand fields on the object in those requests.

src/main/java/com/stripe/param/issuing/CardCreateParams.java
ob-stripe: Not entirely sure why we had the `EmptyParam` overload in the first place, this was probably a side-effect of one of the hacks we added to the Java codegen to handle unsettable fields.

The change is likely because the parameter was converted from an old-style enum (`StringParam` with a set of `values`) to a new-style enum (`APIEnumParam`).

This is technically breaking, but I don't think it matters since sending an empty value in the creation request would have failed anyway.

ob-stripe: Actually this seems to be fixed already in the latest master. Can you try rebasing?

src/main/java/com/stripe/param/AccountUpdateParams.java
```diff
@@ -215,14 +215,14 @@ public Builder setBusinessProfile(BusinessProfile businessProfile) {
...
+    /** The business type. */
+    public Builder setBusinessType(BusinessType businessType) {
       this.businessType = businessType;
       return this;
     }
 
-    /** The business type. Can be `individual` or `company`. */
-    public Builder setBusinessType(EmptyParam businessType) {
+    /** The business type. */
+    public Builder setBusinessType(String businessType) {
```
richardm-stripe: hmm, this doesn't seem right

src/main/java/com/stripe/param/AccountUpdateParams.java
```diff
@@ -215,14 +215,14 @@ public Builder setBusinessProfile(BusinessProfile businessProfile) {
       return this;
     }
 
-    /** The business type. Can be `individual` or `company`. */
-    public Builder setBusinessType(String businessType) {
+    /** The business type. */
+    public Builder setBusinessType(BusinessType businessType) {
```
richardm-stripe: this might be wrong -- do we need all 3, BusinessType, String, and EmptyParam?

richardm-stripe: (If so, https://git.corp.stripe.com/stripe-internal/pay-server/pull/198350)

ob-stripe: We only need `EmptyParam` in update requests if the value is unsettable, which I don't think is the case here.

richardm-stripe: So the fact that we had it before was a mistake?

richardm-stripe: (Re: our convo, in https://github.com/stripe-archive/stripe-api-codegen/pull/143 we made all string params emptyable to support the empty string. The empty string isn't a valid input here -- the API will reject it.

ob-stripe: Final conclusion: we'll keep both the `String` and `EmptyParam` overloads for now to avoid a breaking change, and remove those in the next major version.

PR #926, opened 2020-01-11
[WIP][DONOTMERGE] Credit not line item preview

remi-stripe:


src/main/java/com/stripe/model/CreditNote.java
```diff
@@ -49,6 +50,10 @@
   @Setter(lombok.AccessLevel.NONE)
   ExpandableField<CustomerBalanceTransaction> customerBalanceTransaction;
 
+  /** The integer amount in **%s** representing the amount of the discount that was credited. */
+  @SerializedName("discount")
+  Long discount;
```
remi-stripe: This should be `discount_amount`, bug in the spec

src/main/java/com/stripe/param/CreditNoteCreateParams.java
```diff
@@ -38,6 +40,10 @@
   @SerializedName("invoice")
   String invoice;
 
+  /** Line items that make up the credit note. */
+  @SerializedName("lines")
+  List<Line> lines;
```
remi-stripe: This is the wrong name. Shouldn't be called Line like that

remi-stripe: Change of mind, `Line` makes sense now that I know what it is :)

remi-stripe:
closing now that the code is ready to be codegen-ed

PR #931, opened 2020-01-14
Refactor HTTP headers handling

ob-stripe:
cc @stripe/api-libraries 

This PR refactors the way HTTP headers are handled in the library. I recommend reviewing it commit by commit. Here's what each commit does:

1. Adds a `CaseInsensitiveMap<V>` class. This is a generic map that allows for case-insensitive lookups, but still maintains the original case of keys. It takes a single type parameter `<V>` for values, since keys must be strings.

2. Adds a `HttpHeaders` class for handling HTTP headers, using `CaseInsensitiveMap` under the hood. This class mimics the API of the [`java.net.http.HttpHeaders`](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpHeaders.html) class introduced in Java 11 (which we can't use, because we're still supporting Java 8), although it's a simplified version.

3. Use `HttpHeaders` to handle HTTP headers in `StripeRequest`. Because `HttpHeaders` is a read-only view and we sometimes need to modify headers (e.g. to inject the telemetry header), this forces us to create a new `HttpHeaders` instance with the new headers. It's a bit cumbersome, but I think ultimately it's a better/safer design.

4. Use `HttpHeaders` to handle HTTP headers in `StripeResponse`, and delete the old `StripeHeaders` class.

This is a breaking API change, because `StripeResponse.headers()` and now returns the `HttpHeaders` instance (previously it returned a `StripeHeaders`).


src/main/java/com/stripe/util/CaseInsensitiveMap.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * A case-insensitive {@link Map}.
+ *
+ * <p>All keys are expected to be strings (though {@code null} is a valid key). The map remembers
+ * the case of the last key to be set, and {@link Map#keySet()} or {@link Map#entrySet()} will
```
brandur-stripe: Just OOC, any particular reason to remember the case that was used? Just noting because the Ruby version that I wrote definitely does not do that :)

src/main/java/com/stripe/net/HttpHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+@EqualsAndHashCode
+public class HttpHeaders {
+  private CaseInsensitiveMap<List<String>> headerMap;
+
+  private HttpHeaders(CaseInsensitiveMap<List<String>> headerMap) {
+    this.headerMap = headerMap;
+  }
+
+  /**
+   * Returns an HTTP headers from the given map.
```
brandur-stripe: Tiny nit, but I wonder if we should say "Returns an HTTPHeaders ..." (i.e. the name of the object). The grammar doesn't really make sense unless it's implied that this is the name of a class.

src/main/java/com/stripe/net/HttpURLConnectionClient.java
```diff
@@ -78,18 +79,13 @@ public StripeResponse request(StripeRequest request) throws ApiConnectionExcepti
...
   }
 
-  static Map<String, String> getHeaders(StripeRequest request) {
-    Map<String, String> headers = new HashMap<>();
+  static HttpHeaders getHeaders(StripeRequest request) {
+    Map<String, List<String>> headerMap = new HashMap<>(request.headers().map());
 
-    headers.putAll(request.headers());
+    headerMap.put("User-Agent", Arrays.asList(buildUserAgentString()));
+    headerMap.put("X-Stripe-Client-User-Agent", Arrays.asList(buildXStripeClientUserAgentString()));
```
brandur-stripe: Just calling out that it strikes me as a little weird that you can't use `HttpHeaders` to build a set of headers, and instead have to manipulate a `Map` and then create `HttpHeaders`.

I think based on your PR description that it's supposed to be a safety feature to prevent reuse of an `HttpHeaders` object, but the limitation feels a little over-restrictive IMO.

ob-stripe:
> is the idea that we can migrate to `java.net.http.HttpHeaders` once Java 8 is dropped?

Unfortunately I don't see us dropping Java 8 anytime soon (~85% of stripe-java users are still on Java 8), but at least this way Java 11+ devs are faced with a familiar API. And yes, in the far future where we can drop Java 8, it should be very easy to switch to `java.net.http.HttpHeaders` :)

src/main/java/com/stripe/util/CaseInsensitiveMap.java
ob-stripe: I was influenced by Python's requests' implementation: https://github.com/psf/requests/blob/a9ee0eef5a70acb9ed35622b3675574b11f92cb4/requests/structures.py#L15. I agree it's probably unnecessary, but I thought it might be occasionally useful, for e.g. debugging.

src/main/java/com/stripe/net/HttpHeaders.java
ob-stripe: Haha, I copied the docstring straight from Java 11's `HttpHeader's` [docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpHeaders.html#of(java.util.Map,java.util.function.BiPredicate)), but I agree the grammar doesn't make sense. I'll update it 👍

src/main/java/com/stripe/net/HttpURLConnectionClient.java
ob-stripe: Yeah, I was on the fence about it too. I'll add a helper to `HttpHeaders`.

src/main/java/com/stripe/util/CaseInsensitiveMap.java
brandur-stripe: Fair enough! Seems like it could occasionally be useful, so +1.

src/main/java/com/stripe/net/HttpHeaders.java
brandur-stripe: lol! They probably generated it with some docstring tool that deforms camel case as that is The Java Way ;) Thanks for the change though.

brandur-stripe:
> Unfortunately I don't see us dropping Java 8 anytime soon (~85% of stripe-java users are still on Java 8), but at least this way Java 11+ devs are faced with a familiar API. And yes, in the far future where we can drop Java 8, it should be very easy to switch to `java.net.http.HttpHeaders` :)

85%?! Wow, haha. Okay, that sounds good though. Thanks.

ob-stripe:
@brandur-stripe Okay, I pushed a new commit that adds `withAdditionalHeader()` / `withAdditionalHeaders()` helpers to `HttpHeaders` to easily add headers, as well as `withAdditionalHeader()` on `StripeRequest` itself.

I've also added a `toString()` implementation to `HttpHeaders` to easily view the names and values of the headers.

ptal @brandur-stripe

PR #934, opened 2020-01-14
Add `maxNetworkRetries` as a global and per-request setting

ob-stripe:
cc @stripe/api-libraries 

I added support for automatic request retries back in #900, but the setting for the max number of retries was set on the `HttpClient` instance, which users won't be initializing themselves directly in most cases.

In order to make the feature more accessible, this PR changes the setting so that it can be set in one of two ways:
- per request, via `RequestOptions.builder().setMaxNetworkRetries(2).build()`
- globally, via `Stripe.setMaxNetworkRetries(2)`

(The former takes precedence over the latter.)

### DIFF
--- a/src/main/java/com/stripe/Stripe.java
+++ b/src/main/java/com/stripe/Stripe.java
@@ -26,6 +26,8 @@ public abstract class Stripe {
   private static volatile int connectTimeout = -1;
   private static volatile int readTimeout = -1;
 
+  private static volatile int maxNetworkRetries = 0;
+
   private static volatile String apiBase = LIVE_API_BASE;
   private static volatile String connectBase = CONNECT_API_BASE;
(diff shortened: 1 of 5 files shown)


ob-stripe:
Thanks Brandur!

I agree that setting this property per-request is a bit unorthodox, but it's consistent with what we're doing already for e.g. connect and read timeouts. The main benefit is that `RequestOptions` is something that stripe-java users are already familiar with, so this makes the feature a lot more discoverable (as opposed to creating a custom `HttpClient`, which nobody is familiar with because this was introduced in this integration branch).

I plan on reintroducing this setting as a property on the client when we make more progress towards the client + services architecture, and it becomes more commonplace for users to instantiate clients directly.

brandur-stripe:
Works for me. Thanks OB.

PR #935, opened 2020-01-15
Add `StreamUtils` class

ob-stripe:
cc @stripe/api-libraries 

Amazingly, Java doesn't provide a standard way of extracting the contents of an `InputStream` into a `String`. There are many ways of doing so, [this StackOverflow answer](https://stackoverflow.com/a/35446009/5307473) shows and compares 11 (!) different ways, some of them relying on third-party libraries.

Previously we were using #3, i.e. using `Scanner` to tokenize the stream with `\A` as the delimiter to tokenize the entire stream into a single string. This 1. feels hacky and 2. has pretty bad performance.

In this PR I added a new utility method `StreamUtils.readToEnd()` that uses #6, i.e. using `StringBuilder` and `InputStreamReader`, which feel like the right tools for the job (and should have better performance too).

Tagged as breaking because I changed the type of `ApiResource.CHARSET` from `String` to `Charset`, as well as the constructor of `MultipartProcessor`, though this shouldn't impact many users.

### DIFF
--- a/src/main/java/com/stripe/net/ApiResource.java
+++ b/src/main/java/com/stripe/net/ApiResource.java
@@ -23,12 +23,13 @@ import com.stripe.model.StripeRawJsonObjectDeserializer;
 import com.stripe.util.StringUtils;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Objects;
 
 public abstract class ApiResource extends StripeObject {
-  public static final String CHARSET = StandardCharsets.UTF_8.name();
+  public static final Charset CHARSET = StandardCharsets.UTF_8;
 
   private static StripeResponseGetter stripeResponseGetter = new LiveStripeResponseGetter();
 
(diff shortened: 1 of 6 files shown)


ob-stripe:
(Also I have vague hopes that this will help with some of the transient build failures we're observing on Travis from time to time, but I'm not holding my breath.)

src/main/java/com/stripe/net/ApiResource.java
```diff
@@ -23,12 +23,13 @@
...
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Objects;
 
 public abstract class ApiResource extends StripeObject {
-  public static final String CHARSET = StandardCharsets.UTF_8.name();
+  public static final Charset CHARSET = StandardCharsets.UTF_8;
```
brandur-stripe: Why'd they `String` when they didn't have to?? Types are good! :)

ob-stripe:
Thanks Brandur!

And yes, good point about this being potentially dangerous. Gson does have [support for streams](https://sites.google.com/site/gson/streaming), so I think in theory we could deserialize the stream directly into a model, without first loading the entire stream into memory as a string. In practice I don't think we need to bother with this for the reasons you listed.

PR #940, opened 2020-01-16
Add `StripeObjectInterface` interface

ob-stripe:
cc @stripe/api-libraries 

This PR adds a new `StripeObjectInterface` interface that is implemented by all `StripeObject`s. This allows for better type safety, because we can now constrain generic methods that return a `StripeObject` class with `<T extends StripeObjectInterface>`.

The reason we couldn't simply do `<T extends StripeObject>` is because of polymorphic resources. Those resources are represented by interfaces (e.g. `ExternalAccount` or `PaymentSource`), and interfaces cannot extend classes, only other interfaces.

This is ready to be reviewed, I'm just tagging it as WIP to avoid merging it before I've had a chance to update the codegen accordingly, since some of the files I've modified here (namely, the interfaces and type adapter factories for polymorphic resources) are generated.

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -18,6 +18,7 @@ import com.stripe.exception.oauth.UnsupportedGrantTypeException;
 import com.stripe.exception.oauth.UnsupportedResponseTypeException;
 import com.stripe.model.StripeError;
 import com.stripe.model.StripeObject;
+import com.stripe.model.StripeObjectInterface;
 import com.stripe.model.oauth.OAuthError;
 import java.util.Map;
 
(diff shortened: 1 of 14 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -67,16 +68,13 @@ public LiveStripeResponseGetter(HttpClient httpClient) {
       raiseMalformedJsonError(responseBody, responseCode, requestId);
     }
 
-    if (resource instanceof StripeObject) {
-      StripeObject obj = (StripeObject) resource;
-      obj.setLastResponse(response);
-    }
+    resource.setLastResponse(response);
```
ob-stripe: Because of the type constraint, we no longer need to do an ugly check and cast at runtime here.

ob-stripe:
Thanks both!

The internal PR to update codegen is ready so I'll merge this now.

PR #943, opened 2020-01-17
Add support for `metadata` on Checkout `Session`

remi-stripe:
Codegen for openapi fd2f212

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/checkout/Session.java
+++ b/src/main/java/com/stripe/model/checkout/Session.java
@@ -91,6 +91,13 @@ public class Session extends ApiResource implements HasId {
   @SerializedName("locale")
   String locale;
 
+  /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/model/BalanceTransactionSource.java
```diff
@@ -1,3 +1,3 @@
 package com.stripe.model;
 
-public interface BalanceTransactionSource extends StripeObjectInterface, HasId {}
+public interface BalanceTransactionSource extends HasId {}
```
remi-stripe: are all those changes normal?

ob-stripe: No, you'll need to regenerate with the latest codgen.

PR #947, opened 2020-01-23
Changes for custom HTTP clients

### LINKED ISSUE #946: Custom HTTP clients could be easier
Hi there,

A few days ago we implemented our custom HTTP client for our play application. It's great we can now use the stock Play-WS service provided by Play Framework.

But it could be easier. For request with body, we only have access to the byte array. Isn't it possible to get access to the Map of key/values (since it's a multipart/form-data) directly so we don't have to do weird (and useless) conversions to make it compatible with Play Framework?

Thanks,

ob-stripe:
cc @stripe/api-libraries @Flo354

In the v17 release, I did a large overhaul of the HTTP layer that among other things allows users to provide their own custom HTTP clients. However, the objects that are currently exposed to users are a bit limited. In particular, request bodies are only exposed as a byte array on the `HttpContent` object.

This is fine for our own usecase because our HTTP client (`HttpURLConnectionClient`) only expects a byte array for bodies, but many third-party HTTP libraries provide higher-level abstractions. In order to help users leverage these features, this PR exposes a bit more of our HTTP infrastructure. Specifically:

- `FormEncoder` is now public
- `FormEncoder` has a new public method `flattenParams()` to "pre-encode" a request by flattening nested parameters and encoding all values as strings (except for `File` and `Stream` values, which stay untouched and must be encoded as `multipart` requests)
- `HttpContent` has a new public method `stringContent()` that exposes the body as a `String` (so users don't have to convert the byte array themselves, possibly using the wrong charset)
- `KeyValuePair` is now public (necessary to expose `FormEncoder.flattenParams()` as it returns a list of `KeyValuePair`s)
- `StripeRequest` has a new public method `params()` that returns the raw (unflattened) map of request parameters as an unmodifiable map

Fixes #946.

### DIFF
--- a/src/main/java/com/stripe/net/FormEncoder.java
+++ b/src/main/java/com/stripe/net/FormEncoder.java
@@ -15,7 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-final class FormEncoder {
+public final class FormEncoder {
   public static HttpContent createHttpContent(Map<String, Object> params) throws IOException {
     // If params is null, we create an empty HttpContent because we still want to send the
     // Content-Type header.
@@ -23,7 +23,7 @@ final class FormEncoder {
       return HttpContent.buildFormURLEncodedContent(new ArrayList<KeyValuePair<String, String>>());
     }
 
-    Collection<KeyValuePair<String, Object>> flatParams = flattenParamsValue(params, null);
+    Collection<KeyValuePair<String, Object>> flatParams = flattenParams(params);
 
     // If all parameters have been encoded as strings, then the content can be represented
     // with application/x-www-form-url-encoded encoding. Otherwise, use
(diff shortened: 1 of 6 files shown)


src/main/java/com/stripe/net/FormEncoder.java
```diff
@@ -78,6 +78,16 @@ public static String createQueryString(
             .collect(Collectors.toList()));
   }
 
+  /**
+   * Returns a list of flattened parameters for the given map of parameters.
```
brandur-stripe: Would it be possible to explain what "flatten" in this context means?

ob-stripe: 👍 I added a longer comment.

ob-stripe:
ptal @brandur-stripe

Flo354:
Thanks for the quick PR.

Just one thing about Map. You send over an UnmodifiableMap, but be aware
that even though the Map cannot be modified, object within the Map can
still be modified if they are not immutables.

ob-stripe:
Released as 17.2.0.

PR #951, opened 2020-01-28
Tweak latency telemetry behavior

ob-stripe:
cc @stripe/api-libraries 

Small tweak to the latency telemetry behavior: we now check whether `Stripe.enableTelemetry` is set first, which avoids consuming a latency metrics entry from the queue if we're not going to use it.

I also made the `RequestTelemetry` class public and added a `resetMetrics()` method to clear the queue, mostly to make the tests more consistent, otherwise the results can depend on the order in which tests are executed / whether a single telemetry test is executed vs. the entire test suite.

This PR targets the `integration-telemetry` branch. I plan on making a few more changes with the eventual goal of adding non-latency telemetry (such as whether typed parameters are used or not).

### DIFF
--- a/src/main/java/com/stripe/net/RequestTelemetry.java
+++ b/src/main/java/com/stripe/net/RequestTelemetry.java
@@ -8,8 +8,8 @@ import java.util.Optional;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import lombok.Data;
 
-/** Helper class used by {@link LiveStripeResponseGetter} to manage request telemetry. */
-class RequestTelemetry {
+/** Helper class used by {@link HttpClient} to manage request telemetry. */
+public class RequestTelemetry {
   /** The name of the header used to send request telemetry in requests. */
   public static final String HEADER_NAME = "X-Stripe-Client-Telemetry";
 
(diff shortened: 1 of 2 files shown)


ob-stripe:
> Are you sure that kind of thing wouldn't belong in `X-Stripe-Client-User-Agent` instead? I say that because currently telemetry is just about latency, but `X-Stripe-Client-User-Agent` is a big grab bucket of a lot of informational points.

That's certainly open for debate, but my reasoning was that `X-Stripe-Client-User-Agent` contains general information about the user's environment and does not change across requests, while `X-Stripe-Client-Telemetry` contains information about a specific request.

brandur-stripe:
> That's certainly open for debate, but my reasoning was that `X-Stripe-Client-User-Agent` contains general information about the user's environment and does not change across requests, while `X-Stripe-Client-Telemetry` contains information about a specific request.

Cool. Kind of make sense except that in practice, we always send both with every request, and when we're sampling `X-Stripe-Client-User-Agent` on the server-side, it's sampled per-request as well (and none of that's likely to change). You also have the complication that telemetry is more of a measure of the _last_ request rather the current, and I think you could also argue with use of typed parameters is more of a per-request or a per-integration concern (i.e. likely integrations will favor one direction or another). Anyway, I guess not a big deal either way.

PR #957, opened 2020-01-31
Add additional verification file purpose

lukaszwelnicki:
Creating a Account of business type Company requires company.verification file to be submitted. Based on the API docs, the document's purpose should be `additional_verification` . This type of purpose is missing in `FileCreateParams.Purpose`


remi-stripe:
@lukaszwelnicki Thanks a lot for doing the work and opening a PR with the changes!

We're automatically generating stripe-java internally so I'm working on getting this `purpose` value public in our openapi spec first. I can't merge your change immediately otherwise it would be reverted in the next release but I'll merge this early next week!

PR #964, opened 2020-02-11
Convert Markdown to HTML in Javadoc

ob-stripe:
r? @remi-stripe @richardm-stripe 
cc @stripe/api-libraries 

Convert Markdown to HTML in Javadoc comments.


src/main/java/com/stripe/model/Account.java
```diff
@@ -116,7 +118,10 @@
   @SerializedName("tos_acceptance")
   TosAcceptance tosAcceptance;
 
-  /** The Stripe account type. Can be `standard`, `express`, or `custom`. */
+  /**
+   * The Stripe account type. Can be <code>standard</code>, <code>express</code>, or <code>custom
+   * </code>.
```
remi-stripe: Minor but I'm assuming it's okay that this goes to the next line right?

ob-stripe: It's not perfect because it does cause an extra space to be rendered. I'll see if I can fix it in codegen.

image

ob-stripe: Looked into it, it's actually a bug with google-java-format: https://github.com/google/google-java-format/issues/84. Unfortunately I don't think we can work around it.

ob-stripe: We can work around it :) by using `{@code ...}` instead of `<code>...</code>`. This avoids the bug with google-java-format, and is rendered fine by Javadoc:
image

ob-stripe:
You can view the output of this branch at https://stripe.dev/stripe-java/

Issue #770, opened 2019-05-04
README -> Documentation does not point to Javadoc


goldbach00009:
The link in README does not point to any Javadoc, but to the generic guidance for Stripe web API. A google search for Javadoc associated with a specific version brings up websites not related to stripe.

It could be that my expectation is wrong and there is no Javadoc associated to this API, to describe the various objects and their responsibilities. If that is the case, it should be briefly mentioned, perhaps.

README.md

## Documentation

Please see the [Java API docs](https://stripe.com/docs/api/java) for the most up-to-date documentation.

remi-stripe:
@goldbach00009 Today the library contains the javadoc library so that the documentation appears in your IDE but we don't generate or host a separate javadoc just yet. I do see how that could be useful though so tagging as future! cc @mickjermsurawong-stripe

mickjermsurawong-stripe:
Hi @goldbach00009,
Ah "Java API docs" could easily be understood as the standard java docs. As remi, pointed out, we don't host java docs yet. For now, do you think under the documentation rephrasing as: 
"Stripe API reference for Java" would have made it clearer to you?

goldbach00009:
Yes, it is clearer and perhaps, for the time being, the simple additional information that the standard Javadoc is available in the JAR could be useful, given that the https://stripe.com/docs/api/java is cross-language and, afaik, there is no Javadoc in there as well. I did not think of looking in the JAR, my first idea was to google for it instead...!

ob-stripe:
We could easily host the Javadoc HTML on GitHub Pages or similar, but I'm not convinced it'd be terribly useful -- most of the Javadoc is automatically generated from the OpenAPI spec and the information is going to be the same as / a subset of the API reference. The Javadoc is definitely helpful for IDEs, but if you're looking for an external reference, then you're probably better off looking at the "real" API reference on Stripe's site than the HTML version of Javadoc.

On the other hand, it wouldn't cost us much to host the Javadoc HTML somewhere, and maybe Java devs can navigate the Javadoc HTML more efficiently than the Stripe's API reference.

ob-stripe:
Just wanted to let you know that we're now hosting the library's Javadoc at https://stripe.dev/stripe-java. The README has been updated to mention it.

Closing, but feel free to reply to let us know what you think!

PR #965, opened 2020-02-11
Add Gradle plugin for publishing JavaDoc to GitHub Pages

ob-stripe:
cc @stripe/api-libraries 

This PR adds a Gradle plugin (`org.ajoberstar.git-publish`) that adds support for publishing the JavaDoc HTML output to GitHub pages.

The plugin can be used like this:
```
$ GRGIT_USER=<GitHub API token> ./gradlew gitPublishPush`
```

This command will build the JavaDoc and push the results to the `gh-pages` branch. I've configured GitHub pages to use the contents of this branch. Apparently GitHub Pages even knows about the stripe.dev domain, so the results are available at https://stripe.dev/stripe-java.

The intent is that we would run this command automatically in the release pipeline, so the JavaDoc stays up to date. We could then link to it from the README, as requested by a user some time ago (#770).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,6 +18,7 @@ plugins {
     id "net.ltgt.errorprone" version "1.1.1"
     id "com.github.kt3k.coveralls" version "2.8.4"
     id "biz.aQute.bnd.builder" version "4.2.0"
+    id "org.ajoberstar.git-publish" version "3.0.0-rc.1"
 }
 
 sourceCompatibility = 1.8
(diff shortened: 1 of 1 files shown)


ob-stripe:
> Although, "github pages knows about stripe.dev" seems backwards to me...
> stripe.github.io appears to 301 to stripe.dev, do you happen to know the details of how stripe.dev uses github pages as a backend? Or is it mirrored?

I'm unsure, but I got the stripe.dev URL directly from GitHub:
image

ob-stripe:
> Also consider adding it as a GitHub Action — their system for secrets management works pretty well.

I'm still trying to clarify the status of GitHub Actions on our public repos -- it just got reenabled, but I'd like to make sure it won't be disabled again before we start using it for anything important.

PR #994, opened 2020-03-26
Add support for `spending_controls` on Issuing `Card` and `Cardholder`

remi-stripe:
Codegen for openapi 595f9a2

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/issuing/Card.java
+++ b/src/main/java/com/stripe/model/issuing/Card.java
@@ -33,8 +33,11 @@ public class Card extends ApiResource implements HasId, MetadataStore<Card> {
   String brand;
 
   /**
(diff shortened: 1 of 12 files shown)


src/main/java/com/stripe/model/issuing/Card.java
```diff
@@ -122,6 +128,9 @@
   @SerializedName("shipping")
   Shipping shipping;
 
+  @SerializedName("spending_controls")
+  AuthorizationControls spendingControls;
```
remi-stripe: The class name **should** be `SpendingControls` but internally it's shared as one resource with `AuthorizationControls`.

While it's not great, we're doing a fast follow with a major version that removes all deprecated features and that major version will rename that class too.

PR #1010, opened 2020-04-16
Multiple API changes

remi-stripe:
Multiple API changes:
  * Add `institution_number` and `transit_number` in `payment_method_details[acss]` on `Charge`
  * Add `cvc` and `number` as properties that can be included when retrieving an issuing `Card`.

Codegen for openapi 14a5272

r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Charge.java
+++ b/src/main/java/com/stripe/model/Charge.java
@@ -1137,9 +1137,17 @@ public class Charge extends ApiResource implements MetadataStore<Charge>, Balanc
       @SerializedName("fingerprint")
       String fingerprint;
 
+      /** Institution number of the bank account. */
+      @SerializedName("institution_number")
+      String institutionNumber;
+
       /** Last four digits of the bank account number. */
       @SerializedName("last4")
       String last4;
+
+      /** Transit number of the bank account. */
+      @SerializedName("transit_number")
+      String transitNumber;
     }
 
     @Getter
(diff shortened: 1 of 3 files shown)


totof3110:
@remi-stripe : this might not be the best place to report this but I just tried the new way of loading the card details like:
```java
RequestOptions requestOptions = RequestOptions.builder()
		.setStripeAccount(accountId)
		.build();
CardRetrieveParams params = CardRetrieveParams.builder()
		.addExpand("number")
		.addExpand("cvc")
		.build();
Card card = Card.retrieve(cardId, params, requestOptions);
```
And in the card I get back:
- the number starts with the correct 6-BIN prefix and ends with the correct last 4, but in between are 13 alphanumeric characters (mix of numbers and upper / lower case letters).
- the CVC is a very large number (like 938395385493788174) and seems to change every time I call the API.

Any idea why that is?

remi-stripe:
@totof3110 Thanks for the flag, I'm investigating

remi-stripe:
We're working on a fix and for now we recommend continuing to use the `details()` API method until we get this fixed.

totof3110:
Thanks @remi-stripe. Will this require an update to the SDK or is this a "behind-the-scene" kind of fix? And how will we know to try again?

remi-stripe:
I'll reply here once it's fixed and it won't require a change to the library. Without going too much into the details, the issue is due to sending `expand[0]=number` instead of `expand[]=number`. Both are correct but we only handle the latter (which is incorrect) and are working on a fix! Sorry for the confusion!

remi-stripe:
@totof3110 The issue has been resolved in production and you should be all set. Thanks again for flagging early :)

totof3110:
@remi-stripe : confirmed! It works now. Thanks for the quick fix.

PR #1012, opened 2020-04-17
[WIP][DONOTMERGE]

remi-stripe:
cc @areagle-stripe as an FYI

### DIFF
--- a/src/main/java/com/stripe/model/Price.java
+++ b/src/main/java/com/stripe/model/Price.java
@@ -1,0 +1,370 @@
+package com.stripe.model;
+
+import com.google.gson.annotations.SerializedName;
+import com.stripe.Stripe;
+import com.stripe.exception.StripeException;
+import com.stripe.net.ApiResource;
+import com.stripe.net.RequestOptions;
+import com.stripe.param.PriceCreateParams;
+import com.stripe.param.PriceListParams;
+import com.stripe.param.PriceRetrieveParams;
+import com.stripe.param.PriceUpdateParams;
+import java.math.BigDecimal;
+import java.util.List;
+import java.util.Map;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class Price extends ApiResource implements HasId, MetadataStore<Price> {
+  /** Whether the price can be used for new purchases. */
+  @SerializedName("active")
+  Boolean active;
+
+  /**
+   * Describes how to compute the price per period. Either {@code per_unit} or {@code tiered}.
+   * {@code per_unit} indicates that the fixed amount (specified in {@code unit_amount} or {@code
+   * unit_amount_decimal}) will be charged per unit in {@code quantity} (for prices with {@code
+   * usage_type=licensed}), or per unit of total usage (for prices with {@code usage_type=metered}).
+   * {@code tiered} indicates that the unit pricing will be computed using a tiering strategy as
+   * defined using the {@code tiers} and {@code tiers_mode} attributes.
+   *
+   * <p>One of {@code per_unit}, or {@code tiered}.
+   */
+  @SerializedName("billing_scheme")
+  String billingScheme;
+
+  /** Time at which the object was created. Measured in seconds since the Unix epoch. */
+  @SerializedName("created")
+  Long created;
+
+  /**
+   * Three-letter <a href="https://www.iso.org/iso-4217-currency-codes.html">ISO currency code</a>,
+   * in lowercase. Must be a <a href="https://stripe.com/docs/currencies">supported currency</a>.
+   */
+  @SerializedName("currency")
+  String currency;
+
+  /** Always true for a deleted object. */
+  @SerializedName("deleted")
+  Boolean deleted;
+
+  /** Unique identifier for the object. */
+  @Getter(onMethod_ = {@Override})
+  @SerializedName("id")
+  String id;
+
+  /**
+   * Has the value {@code true} if the object exists in live mode or the value {@code false} if the
+   * object exists in test mode.
+   */
+  @SerializedName("livemode")
+  Boolean livemode;
+
+  /** A lookup key used to retrieve prices dynamically from a static string. */
+  @SerializedName("lookup_key")
+  String lookupKey;
+
+  /**
+   * Set of key-value pairs that you can attach to an object. This can be useful for storing
+   * additional information about the object in a structured format.
+   */
+  @Getter(onMethod_ = {@Override})
+  @SerializedName("metadata")
+  Map<String, String> metadata;
+
+  /** A brief description of the plan, hidden from customers. */
+  @SerializedName("nickname")
+  String nickname;
+
+  /**
+   * String representing the object's type. Objects of the same type share the same value.
+   *
+   * <p>Equal to {@code price}.
+   */
+  @SerializedName("object")
+  String object;
+
+  /** The ID of the product this price is associated with. */
+  @SerializedName("product")
+  @Getter(lombok.AccessLevel.NONE)
+  @Setter(lombok.AccessLevel.NONE)
+  ExpandableField<Product> product;
+
+  /** The recurring components of a price such as {@code interval} and {@code usage_type}. */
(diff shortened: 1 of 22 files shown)


src/main/java/com/stripe/model/Price.java
```diff
@@ -0,0 +1,328 @@
...
+  /** The recurring components of a price such as {@code interval} and {@code usage_type}. */
+  @SerializedName("recurring")
+  Recurring recurring;
+
+  /**
+   * Each element represents a pricing tier. This parameter requires {@code billing_scheme} to be
+   * set to {@code tiered}. See also the documentation for {@code billing_scheme}.
+   */
+  @SerializedName("tiers")
+  List<Plan.Tier> tiers;
```
remi-stripe, line 106: This is problematic and we need to fix it internally, possibly by duplicating the class/resource. `Price` is a new resource and it shouldn't reference the deprecated one. And we can't change `Plan` to use the `Price.Tier` either since it's a breaking change.

src/main/java/com/stripe/model/Price.java
```diff
@@ -0,0 +1,328 @@
...
+   */
+  @SerializedName("tiers_mode")
+  String tiersMode;
+
+  /**
+   * Apply a transformation to the reported usage or set quantity before computing the amount
+   * billed. Cannot be combined with {@code tiers}.
+   */
+  @SerializedName("transform_quantity")
+  Plan.TransformUsage transformQuantity;
```
remi-stripe, line 123: Same as above

src/main/java/com/stripe/model/SubscriptionItem.java
```diff
@@ -71,6 +71,15 @@
   @SerializedName("plan")
   Plan plan;
 
+  /**
+   * The price object defines the unit cost (or tiered cost) for both recurring and one-time prices.
+   * For example, you might have a &lt;currency&gt;5&lt;/currency&gt;/month recurring price for
```
remi-stripe, line 76: There are encoding issues here and likely in similar docstrings that we need to look into

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -467,6 +467,38 @@ public SubscriptionSchedule release(
         ApiResource.RequestMethod.POST, url, params, SubscriptionSchedule.class, options);
   }
 
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class AddInvoiceItemsEntry extends StripeObject {
```
remi-stripe, line 473: This class shouldn't be here I think and instead should be inside `Subscription` which is the core resource. You will need to change the `client_metadata`. Also the name should be `AddInvoiceItem` with no S or extra word

src/main/java/com/stripe/model/SubscriptionSchedule.java
```diff
@@ -467,6 +467,38 @@ public SubscriptionSchedule release(
...
+    /** ID of the price used to generate the invoice item. */
+    @SerializedName("price")
+    @Getter(lombok.AccessLevel.NONE)
+    @Setter(lombok.AccessLevel.NONE)
+    ExpandableField<Plan> price;
```
remi-stripe, line 478: There's a bug here, it should be a `Price` not a `Plan`

src/main/java/com/stripe/model/checkout/Session.java
```diff
@@ -77,6 +78,10 @@
   @SerializedName("id")
   String id;
 
+  /** The line items purchased by the customer. */
+  @SerializedName("line_items")
+  List<Session.LineItem> lineItems;
```
remi-stripe, line 83: I think this needs to be separate from the Price changes. The problem is that Price can ship but the LineItem can't ship. Right now this is an inner class in Java but it's supposed to point to a top-level resource called `LineItem` and it will be a breaking change

src/main/java/com/stripe/param/InvoiceItemUpdateParams.java
```diff
@@ -300,6 +318,24 @@ public Builder setPeriod(Period period) {
...
     }
 
+    /** The ID of the price object. */
+    public Builder setPrice(String price) {
+      this.price = price;
+      return this;
+    }
+
+    /** The ID of the price object. */
+    public Builder setPrice(EmptyParam price) {
```
remi-stripe, line 328: Is it expected that you can unset a `price` on an existing `InvoiceItem`?

src/main/java/com/stripe/param/InvoiceItemUpdateParams.java
```diff
@@ -300,6 +318,24 @@ public Builder setPeriod(Period period) {
...
+      this.price = price;
+      return this;
+    }
+
+    /** Data used to generate a new price object inline. */
+    public Builder setPriceData(PriceData priceData) {
```
remi-stripe, line 334: Same here, should you be allowed to change this information after creation?

src/main/java/com/stripe/param/InvoiceUpcomingParams.java
```diff
@@ -1046,6 +1076,392 @@ public Builder setStart(Long start) {
         }
       }
     }
+
+    @Getter
+    public static class PriceData {
```
remi-stripe, line 1081: Is it safe to pass `price_data` to an upcoming endpoint? Do we create a fake price/product under the hood?

src/main/java/com/stripe/param/PriceListParams.java
```diff
@@ -0,0 +1,568 @@
...
+  /**
+   * A limit on the number of objects to be returned. Limit can range between 1 and 100, and the
+   * default is 10.
+   */
+  @SerializedName("limit")
+  Long limit;
+
+  /** Only return the price with these lookup_keys, if any exist. */
+  @SerializedName("lookup_key")
+  List<String> lookupKey;
```
remi-stripe, line 62: Is this expected to be a list of strings? And if so should it be called `lookupKeys` instead?

src/main/java/com/stripe/param/SubscriptionItemUpdateParams.java
```diff
@@ -72,6 +73,14 @@
   @SerializedName("plan")
   Object plan;
 
+  /** The ID of the price object. */
+  @SerializedName("price")
+  Object price;
```
remi-stripe, line 78: Same as for InvoiceItem, I don't understand why we would let you unset a `Price`

src/main/java/com/stripe/param/SubscriptionItemUpdateParams.java
```diff
@@ -72,6 +73,14 @@
...
+
+  /** Data used to generate a new price object inline. */
+  @SerializedName("price_data")
+  PriceData priceData;
```
remi-stripe, line 82: I think this is needed to move to a different price, but I'm curious if we should forbid this.

Also feels like if you use Checkout it's okay to do `price_data` but if you use SubscriptionItems you really should have Product and Price in your account already. Like how we never added `plan_data`

src/main/java/com/stripe/model/Price.java
remi-stripe: @areagle-stripe This one still seems to be incorrect.

src/main/java/com/stripe/model/Price.java
remi-stripe: @areagle-stripe same as above, not fixed

src/main/java/com/stripe/model/SubscriptionItem.java
remi-stripe: Same as above, not resolved

src/main/java/com/stripe/model/SubscriptionSchedule.java
remi-stripe: So the name is still incorrect, it shouldn't be `AddInvoiceItemsEntry` it should be `AddInvoiceItem`

src/main/java/com/stripe/model/SubscriptionSchedule.java
remi-stripe: Same as above, still a `Plan` and not a `Price`

remi-stripe:
this has been released, closing

PR #1017, opened 2020-04-27
Changelog change porposal

thorsten-stripe:
r? @remi-stripe 

Proposal to use this format for all changelog headings going forward to make it easier to spot which library version is pinning which API version.

Wdyt?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 19.5.0 - 2020-04-24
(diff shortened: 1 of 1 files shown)


remi-stripe:
Thanks for the PR @thorsten-stripe. I am not a fan of the change but I'll let others internally decide.

richardm-stripe:
I think it's potentially confusing to have two dates in the header. I think it would be enough if we described the change consistently (i.e. picked between "Pin to API version" and "Update API version to") and consistently made this the row below the header.

thorsten-stripe:
Thanks both, totally open to a different format, I just think the current format is confusing as folks might mistake the date as the API version (I know I have). I like @richardm-stripe 's proposal as it would allow folks to easily know which API version they'll be running on 👍 

I'll close this PR, I think it's fine to not do this retroactively but rather just going forward.

PR #1026, opened 2020-05-08
Support checkout.Session line items

richardm-stripe:
r? @remi-stripe 
cc @areagle-stripe

### DIFF
--- a/src/main/java/com/stripe/model/checkout/Session.java
+++ b/src/main/java/com/stripe/model/checkout/Session.java
@@ -17,7 +17,7 @@ import com.stripe.model.Subscription;
 import com.stripe.net.ApiResource;
 import com.stripe.net.RequestOptions;
 import com.stripe.param.checkout.SessionCreateParams;
-import com.stripe.param.checkout.SessionLineItemCollectionListParams;
+import com.stripe.param.checkout.SessionListLineItemsParams;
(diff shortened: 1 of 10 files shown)


src/main/java/com/stripe/model/checkout/Session.java
```diff
@@ -323,6 +333,43 @@ public static Session create(SessionCreateParams params, RequestOptions options)
...
+            "%s%s",
+            Stripe.getApiBase(),
+            String.format(
+                "/v1/checkout/sessions/%s/line_items", ApiResource.urlEncodeId(this.getId())));
+
+    return ApiResource.requestCollection(url, params, LineItemCollection.class, options);
+  }
+
+  /** Returns a list of Line Items */
+  public LineItemCollection listLineItems(SessionLineItemCollectionListParams params)
```
remi-stripe, line 355: I believe the name is backwards. When we confirm a `PaymentIntent` it's `PaymentIntentConfirmParams`. Similarly, this is a nested resource and so usually you do `<operation><nestedresource>` so in ruby it would be `listLineItems` and in the future `createLineItem` `deleteLineItem` etc.

So I believe the correct name for the class should be `SessionListLineItemsParams` instead.

remi-stripe:
new PR replacing it

PR #1036, opened 2020-05-29
Add support for BACS Debit as a `PaymentMethod`

remi-stripe:
Add support for BACS Debit as a `PaymentMethod`
  * Add `settings[bacs_debit_payments]` on `Account`
  * Add `bacs_debit` as a `type` on `PaymentMethod` 
  * Add `payment_method_details[bacs_debit]` on `Charge`
  * Add `bacs_debit` on `Mandate`

Codegen for openapi 7a853ce

r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -1025,6 +1025,9 @@ public class Account extends ApiResource implements MetadataStore<Account>, Paym
   @Setter
   @EqualsAndHashCode(callSuper = false)
   public static class Settings extends StripeObject {
+    @SerializedName("bacs_debit_payments")
+    BacsDebitPayments bacsDebitPayments;
+
     @SerializedName("branding")
     SettingsBranding branding;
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/stripe/model/Account.java
```diff
@@ -1025,6 +1025,9 @@ public void setFrontObject(File expandableObject) {
   @Setter
   @EqualsAndHashCode(callSuper = false)
   public static class Settings extends StripeObject {
+    @SerializedName("bacs_debit_payments")
+    BacsDebitPayments bacsDebitPayments;
```
remi-stripe, line 1029: This one doesn't have the `Settings` prefix because it's nested inside the `Settings` class as it should. It's the other ones we missed unfortunately and would need to change in a future major

src/main/java/com/stripe/model/Mandate.java
```diff
@@ -184,6 +198,9 @@ public static Mandate retrieve(
     @SerializedName("au_becs_debit")
     AuBecsDebit auBecsDebit;
 
+    @SerializedName("bacs_debit")
```
remi-stripe, line 201: This should be _inside_ `PaymentMethodDetails` but... the others are not, so I'll just ignore and we'll fix in a major.

PR #1048, opened 2020-06-23
Multiple API changes

remi-stripe:
Multiple API changes:
  * Add `discounts` on `LineItem`
  * Add `document_provider_identity_document` as a `purpose` on `File`
  * Support nullable `metadata` on Issuing `Dispute`
  * Add `klarna[shipping_delay]` on `Source`

Codegen for openapi 75c2f60

r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/LineItem.java
+++ b/src/main/java/com/stripe/model/LineItem.java
@@ -32,6 +32,10 @@ public class LineItem extends StripeObject implements HasId {
   @SerializedName("description")
   String description;
 
+  /** The discounts applied to the line item. */
+  @SerializedName("discounts")
+  List<LineItem.Discount> discounts;
+
   /** Unique identifier for the object. */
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/model/LineItem.java
```diff
@@ -73,6 +77,25 @@
...
+
+    /**
+     * A discount represents the actual application of a coupon to a particular customer. It
+     * contains information about when the discount began and when it will end.
+     *
+     * <p>Related guide: <a href="https://stripe.com/docs/billing/subscriptions/discounts">Applying
+     * Discounts to Subscriptions</a>.
+     */
+    @SerializedName("discount")
+    Discount discount;
```
cjavilla-stripe, line 96: Is this supposed to be recursive like this? Like a Discount has a Discount? I feel like i'm missing something.

PR #1077, opened 2020-08-05
Multiple API changes

remi-stripe:
Multiple API changes:
  * Add support for the `PromotionCode` resource and APIs
  * Add support for `allow_promotion_codes` on Checkout `Session`
  * Add support for `applies_to[products]` on `Coupon`
  * Add support for `promotion_code` on `Customer` and `Subscription`
  * Add support for `promotion_code` on `Discount`

Codegen for openapi f71053e + an urgent change to fix the `AppliesTo` name which was `Restrictions` incorrectly

r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/Coupon.java
+++ b/src/main/java/com/stripe/model/Coupon.java
@@ -10,6 +10,7 @@ import com.stripe.param.CouponListParams;
 import com.stripe.param.CouponRetrieveParams;
 import com.stripe.param.CouponUpdateParams;
 import java.math.BigDecimal;
+import java.util.List;
 import java.util.Map;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
@@ -25,6 +26,9 @@ public class Coupon extends ApiResource implements HasId, MetadataStore<Coupon> 
    */
   @SerializedName("amount_off")
   Long amountOff;
+
+  @SerializedName("applies_to")
+  Restrictions appliesTo;
 
   /** Time at which the object was created. Measured in seconds since the Unix epoch. */
   @SerializedName("created")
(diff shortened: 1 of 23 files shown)


src/test/java/com/stripe/functional/StripeObjectTest.java
```diff
@@ -22,15 +22,6 @@ public void testGetRawJsonObject() throws StripeException {
         subscription.getCreated(),
         subscription.getRawJsonObject().getAsJsonPrimitive("created").getAsLong());
 
-    // Access `plan[id]`, a nested string element
```
remi-stripe, line 25: `plan` top-level has been undocumented on `Subscription` and there's no equivalent hash that is set. I'm assuming that since we look deeper in the hash below (for `items`) then we don't need this test. @ob-stripe are you okay with this test being removed?

src/main/java/com/stripe/model/Coupon.java
```diff
@@ -26,6 +27,9 @@
   @SerializedName("amount_off")
   Long amountOff;
 
+  @SerializedName("applies_to")
+  Restrictions appliesTo;
```
cjavilla-stripe, line 31: Wonder why this is called Restrictions and not CouponAppliesTo ?

cjavilla-stripe: or maybe just an inner class `AppliesTo`?

remi-stripe: OH NO ugh I must have named it wrong and that wasn't caught in the PR. I guess I have to fix it internally now. sadness.

PR #1088, opened 2020-08-28
Multiple API changes

remi-stripe:
Multiple API changes
  * Pin to API version `2020-08-27`
  * Removed `authenticated` and `succeeded` on `payment_method_details[card][three_d_secure]` on `Charge`
  * Removed `tax_percent` and `prorate` across all Billing APIs
  * Renamed `plans` to `items` on `SubscriptionSchedule`
  * Removed `display_items` on Checkout `Session`
  * Removed `save_payment_method` and `source` on `PaymentIntent` confirm, create and update. Those parameters still work in the API but are removed from the library
  * Removed `payment_method_details[bitcoin]` on `Charge`
  * Removed `unified_proration` on Billing API as this is a deprecated feature that never shipped publicly
  * Removed `plan` and `quantity` from `Subscription`, use `items` instead
  * Removed `requested_capabilities` on `Account` creation, use `capabilities` instead
  * Removed `failure_url` and `success_url` from `AccountLink`, use `refresh_url` and `return_url` instead
  * Removed `invoice.payment_succeeded` and `payment_method.card_automatically_updated` events.

Codegen for openapi 6c45bc9

cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/param/PaymentIntentConfirmParams.java
+++ b/src/main/java/com/stripe/param/PaymentIntentConfirmParams.java
@@ -1517,7 +1517,7 @@ public class PaymentIntentConfirmParams extends ApiRequestParams {
     public static class BillingDetails {
       /** Billing address. */
       @SerializedName("address")
-      Address address;
+      Object address;
 
       /** Email address. */
       @SerializedName("email")
@@ -1541,7 +1541,7 @@ public class PaymentIntentConfirmParams extends ApiRequestParams {
       String phone;
 
       private BillingDetails(
-          Address address,
+          Object address,
           String email,
           Map<String, Object> extraParams,
           String name,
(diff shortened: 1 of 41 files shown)


src/main/java/com/stripe/param/PaymentIntentConfirmParams.java
```diff
@@ -1517,7 +1447,7 @@ public Builder putAllExtraParam(Map<String, Object> map) {
     public static class BillingDetails {
       /** Billing address. */
       @SerializedName("address")
-      Address address;
+      Object address;
```
richardm-stripe, line 1450: I don't think I noticed this change in the changelog.

remi-stripe: yeah I don't usually document when we support unsetting hashes

PR #1095, opened 2020-09-08
OAuth.deauthorize no longer mutates params

### LINKED ISSUE #1091: Don't assume that a provided Map<String, String> is mutable
There are places in stripe-java where arguments are provided as a `Map`, and it modifies the map content. For example, in [OAuth.java:29](https://github.com/stripe/stripe-java/blob/ec82d87509fd04fad930ba9e0cee94a3ff633cca/src/main/java/com/stripe/net/OAuth.java#L29)

```Java
params.put("client_id", getClientId(params, options));
```

In general modifying objects provided as params to your API is not considered good form, as it makes this use case impossible:

```
OAuth.authorizeUrl(
  Map.of(
    "client_id", "foobar",
    "response_type", "code",
   // ... other parameters...
  ),
  url
)
```

In this example, `Map.of()` returns an immutable Map, which should be a valid parameter to pass to Stripe. Instead, the current implementation tries to `put()` and throws an Exception. 

A common workaround would be to use a forwarding map instead.

Java version: 1.11
stripe-java version: 19.35.0

suz-stripe:
:wave:

Fixes #1091 🎉 

r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/net/OAuth.java
+++ b/src/main/java/com/stripe/net/OAuth.java
@@ -58,13 +58,13 @@ public final class OAuth {
    * @param options the request options.
    * @return the DeauthorizedAccount instance containing the response from the OAuth API.
    */
-  public static DeauthorizedAccount deauthorize(
-      Map<String, Object> initialParams, RequestOptions options) throws StripeException {
-    Map<String, Object> params = new HashMap<>();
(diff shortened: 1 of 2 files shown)


richardm-stripe (approved):
LGTM!

PR #1118, opened 2020-09-30
Add support for the `SetupAttempt` resource and List API

remi-stripe:
Codegen for openapi 3854362

r? @paulasjes-stripe
cc @stripe/api-libraries

### DIFF
--- a/src/main/java/com/stripe/model/SetupAttempt.java
+++ b/src/main/java/com/stripe/model/SetupAttempt.java
@@ -1,0 +1,247 @@
+// File generated from our OpenAPI spec
+package com.stripe.model;
+
+import com.google.gson.annotations.SerializedName;
+import com.stripe.Stripe;
+import com.stripe.exception.StripeException;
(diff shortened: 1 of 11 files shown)


paulasjes-stripe (approved):

src/main/java/com/stripe/model/SetupAttempt.java
```diff
@@ -0,0 +1,247 @@
...
+    @SerializedName("type")
+    String type;
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class Card extends StripeObject {
+      /** Populated if this authorization used 3D Secure authentication. */
+      @SerializedName("three_d_secure")
+      Charge.PaymentMethodDetails.Card.ThreeDSecure threeDSecure;
```
remi-stripe, line 244: Flagging this should not be shared but for now Java does still share so deferring to a future major

PR #1139, opened 2020-10-22
Upgrade Lombok plugin

ob-stripe:
r? @richardm-stripe 

Upgrade Lombok Gradle plugin, and bump Lombok to the latest version.

The official Gradle plugin for Lombok changed, so this was a bit of a tricky update, but it does simplify the Gradle configuration and I've verified that the generated sources and Javadoc artifacts are unchanged (save for some minor changes brought by the Lombok upgrade itself and unrelated to the plugin).


deploy.gradle
```diff
@@ -104,24 +104,18 @@ afterEvaluate { project ->
         onlyIf { isReleaseBuild() && project.hasProperty("signing.gnupg.keyName") }
     }
 
-    task makeJavadocs(type: Javadoc, dependsOn: delombok) {
-        source = delombok.outputDir
-        classpath = configurations.compileClasspath + configurations.annotationProcessor
-        failOnError = true
+    task sourcesJar(type: Jar) {
+        classifier = "sources"
+        from delombok
```
ob-stripe, line 109: This ensure the sources JAR uses the delombok'd sources.

deploy.gradle
```diff
@@ -104,24 +104,18 @@ afterEvaluate { project ->
...
     }
 
-    task makeJavadocsJar(type: Jar, dependsOn: makeJavadocs) {
+    task javadocJar(type: Jar, dependsOn: javadoc) {
```
ob-stripe, line 112: Since the default `javadoc` task is updated by the plugin to use the delombok'd sources, we simply need to reference it and no longer need the `makeJavadocs` task.

richardm-stripe (approved):
Looks good! I think some of your PR comments might be useful as inline code comments, too.

PR #1158, opened 2020-11-24
Better log the error causing the JSON parsing to fail on deserialization

remi-stripe:
We're now logging the error message associated with the `JsonSyntaxException` exception raised on parsing, if any. We also record that exception as the cause of the `ApiException` we throw.

r? @richardm-stripe 
cc @stripe/api-libraries


richardm-stripe (approved):
LGTM

remi-stripe:
Pushed a brand new version based on @ob-stripe's work just adding an extra details in the default log line so that we can debug this more easily. He also helped write an explicit test for this

ob-stripe:
Note that `StripeObjectTest` is a terrible place for this test as it has nothing to do with `StripeObject` -- I just used it as a quick and dirty launch ramp. This test should be in a new `LiveStripeResponseGetterTest` class.

richardm-stripe (approved):
This looks good to me.

I took the liberty of moving the test per @ob's comment in the latest commit.

PR #1166, opened 2020-12-17
MetadataStore extends HasId

totof3110:
We're getting a bit of a "generics headache" re. `MetadataStore` vs `HasId`, mainly due to the fact that `MetadataStore` does not extend `HasId` (even though all impls that I found do). This makes it hard to write generic code that expects an object that has both interfaces.

Would this change to the `MetadataStore` make sense?

Before change we have to write:
```java
private <T extends MetadataStore<T>> T updateMetadata(T t, Map<String, String> metadata) throws StripeException {
	log.info("Adding metadata to {} in account {}: {}.", ((HasId) t).getId(), metadata);
	Map<String, Object> params = new HashMap<>();
	params.put("metadata", metadata);
	T updated = (T) t.update(params);
	t = updated;
	return t;
}
```

After we can do:
```java
private <T extends MetadataStore<T>> T updateMetadata(T t, Map<String, String> metadata) throws StripeException {
	log.info("Adding metadata to {} in account {}: {}.", t.getId(), metadata);
	Map<String, Object> params = new HashMap<>();
	params.put("metadata", metadata);
	return t.update(params);
}
```

### DIFF
--- a/src/main/java/com/stripe/model/MetadataStore.java
+++ b/src/main/java/com/stripe/model/MetadataStore.java
@@ -5,11 +5,10 @@ import com.stripe.net.RequestOptions;
 import java.util.Map;
 
 /** Common interface for Stripe objects that can store metadata. */
-public interface MetadataStore<T> {
+public interface MetadataStore<T extends MetadataStore<T>> extends HasId {
   Map<String, String> getMetadata();
 
-  MetadataStore<T> update(Map<String, Object> params) throws StripeException;
+  T update(Map<String, Object> params) throws StripeException;
 
-  MetadataStore<T> update(Map<String, Object> params, RequestOptions options)
-      throws StripeException;
+  T update(Map<String, Object> params, RequestOptions options) throws StripeException;
 }


richardm-stripe:
Hi @totof3110, thanks for the submission. I hesitate to merge this -- in general good practice is to keep interfaces as unrestrictive as possible, and while by happenstance all implementations of MetadataStore happen to also implement HasId, there is no logical reason why it would be so, and you can imagine situations (especially in tests) where you would want to be able to implement MetadataStore without implementing HasId.

Out of curiosity, wouldn't multiple type parameter bounds work for this use case?
```java
private <T extends HasId & MetadataStore<T>> T updateMetadata {
// ...
```

totof3110:
Hi @richardm-stripe ! With this signature it's unfortunately not possible to cast an object to pass to this method. Here's a simplified version of what I'm trying to do:
```java
private void processBalanceTransaction(BalanceTransaction balanceTransaction) {
    BalanceTransactionSource sourceObject = balanceTransaction.getSourceObject();
    if (sourceObject instanceof MetadataStore<?>) {
        MetadataStore<?> metadataStore = (MetadataStore<?>) sourceObject; // <- it's not possible to cast sourceObject to something that works with updateMetadata (other than trying instanceof / cast for every class that implements both MetadataStore and HasId.
        updateMetadata(metadataStore, Map.of("foo", "bar")); // <- doesn't compile.
    }
}

private <T extends HasId & MetadataStore<T>> T updateMetadata(T t, Map<String, String> metadata) {
    // ...
}
```

totof3110:
Actually I just discovered that Java supports casting to multiple interfaces...
```java
private void processBalanceTransaction(BalanceTransaction balanceTransaction) {
    BalanceTransactionSource sourceObject = balanceTransaction.getSourceObject();
    if (sourceObject instanceof MetadataStore<?>) {
        MetadataStore<?> metadataStore = (MetadataStore<?>) sourceObject;
        updateMetadata((HasId & MetadataStore<?>) sourceObject, Map.of("foo", "bar"));
    }
}

private <T extends HasId & MetadataStore<?>> T updateMetadata(T t, Map<String, String> metadata) {
    // ...
}
```
I guess that works! :P

PR #1180, opened 2021-02-06
Accept all InputStreams for FileCreateParams.setFile

### LINKED ISSUE #1170: FileCreateParams.setFile does not accept an input stream
In my use case, I  have to save an input stream generated from the client side request file payload to a temporary file in order to then upload that file to Stripe. The file is then immediately deleted.

I'd like to avoid this performance overhead and pass an input stream to the setFile method. Is this possible?

ob-stripe:
r? @richardm-stripe 
cc @stripe/api-libraries 

Change `FileCreateParams.setFile` to accept any `InputStream`. Since `InputStream` is a supertype of `FileInputStream`, this is not a breaking change.

Fixes #1170.


richardm-stripe (approved):
Looks good! I'll take the release.

PR #1188, opened 2021-02-25
add docs to readme about JVM caching risk

suz-stripe:
r? @richardm-stripe 

Some of our users have reported intermittent timeouts / issues and we were able to identify them as DNS related when the Stripe API's IP address can be subject to change.

Updating the docs to make a recommendation around this since we already advise best practice for other network related mitigations.


richardm-stripe (commented):
LGTM: although it might be worth spelling out "availability interruptions" explicitly: viz. "otherwise, when Stripe's IP address changes, your application's requests to Stripe will all fail until the JVM restarts."

PR #1192, opened 2021-03-26
API Updates

richardm-stripe:
Codegen for openapi 06ec3f8.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `card_issuing` on `AccountCreateParams.settings`, `AccountUpdateParams.settings` and `Account.settings`

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -1208,12 +1208,12 @@ public class Account extends ApiResource implements MetadataStore<Account>, Paym
     @EqualsAndHashCode(callSuper = false)
     public static class SettingsCardIssuing extends StripeObject {
       @SerializedName("tos_acceptance")
-      CardIssuingTermsOfService tosAcceptance;
+      TosAcceptance tosAcceptance;
 
       @Getter
       @Setter
       @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/model/Account.java
```diff
@@ -1199,6 +1202,37 @@ public void setFrontObject(File expandableObject) {
       @SerializedName("creditor_id")
       String creditorId;
     }
+
+    @Getter
+    @Setter
+    @EqualsAndHashCode(callSuper = false)
+    public static class SettingsCardIssuing extends StripeObject {
```
remi-stripe, line 1209: Every time I see those I get sad, they should have been nested inside `Settings` instead but we missed fixing them originally and then got stuck

src/main/java/com/stripe/model/Account.java
```diff
@@ -1199,6 +1202,37 @@ public void setFrontObject(File expandableObject) {
...
+      @SerializedName("tos_acceptance")
+      CardIssuingTermsOfService tosAcceptance;
```
remi-stripe, line 1211: The name is wrong here right? It should be TosAcceptance no?

remi-stripe (changes requested):

remi-stripe (approved):

PR #1203, opened 2021-05-05
Fix flaky test: no telemetry by default in test suite

richardm-stripe:
Our CI in github actions has ~consistently been failing tests like [this](https://github.com/stripe/stripe-java/runs/2513156266?check_suite_focus=true#step:7:675). This hasn't been consistent across time, and it doesn't happen in Travis CI (which we are trying to deprecate).

This happens in `RequestTelemetry.java`, which in my opinion shouldn't be enabled when we are running tests.

This disables telemetry throughout tests, which causes CI to pass. 

I do want to better understand the cause of why the NullPointerException is triggering in this environment but not locally, but have not yet identified it.

### DIFF
--- a/src/test/java/com/stripe/BaseStripeTest.java
+++ b/src/test/java/com/stripe/BaseStripeTest.java
@@ -117,6 +117,7 @@ public class BaseStripeTest {
     Stripe.overrideUploadBase("http://localhost:" + port);
     Stripe.apiKey = "sk_test_123";
     Stripe.clientId = "ca_123";
+    Stripe.enableTelemetry = false;
 
     networkSpy = Mockito.spy(new LiveStripeResponseGetter());
     ApiResource.setStripeResponseGetter(networkSpy);


remi-stripe (approved):
fine with the change but that means we could break telemetry and not notice right? Or do we have enough coverage in tests specifically for this?

richardm-stripe:
Good point, we have [some tests](https://github.com/stripe/stripe-java/blob/master/src/test/java/com/stripe/functional/TelemetryTest.java) that explicitly test enabled/disabled Telemetry. Still, I don't really want to merge this until I actually understand the error.

richardm-stripe:
Ok, I now understand what is going on.

In `RequestTelemetry.java` these is a [piece of global state](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/net/RequestTelemetry.java#L20-L21):
```
private static ConcurrentLinkedQueue<RequestMetrics> prevRequestMetrics
```
which can affect the behavior of `RequestMetrics.getHeaderValue`. Sometimes it returns empty, sometimes not, depending on previous/concurrent requests that have been made throughout the test suite. (This is why I can never reproduce the failure running the test in isolation).

When it's not empty, then `requestWithTelemetry` [adds a header](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/net/HttpClient.java#L49-L52) to the `request` argument before passing it to `this.request`.

This prevents the [Mockito matcher](https://github.com/stripe/stripe-java/blob/master/src/test/java/com/stripe/net/HttpClientTest.java#L46) from applying, and so the mocket client just returns null instead of providing a response. Attempting to call a method on the null response triggers the null pointer exception.

---

Given the role of Mockito, this edge case can't arise in production. Disabling Telemetry by default for requests is a reasonable solution, because that removes the global state/non-determinism here.

PR #1207, opened 2021-05-12
Streaming requests

richardm-stripe:
## Notify
r? @dcr-stripe 
cc @stripe/api-libraries 

## Summary

Adds `ApiResource.requestStream`, which is like `ApiResource.request` except does not buffer the response body into memory and makes no attempt to deserialize the body. Instead, returns an `InputStream`.

Motivation: a `/pdf` endpoint is coming soon to the Stripe API, which returns the bytes of a pdf. Users will want to e.g. pipe these bytes to a file, or pipe them to an email API. This is unlike all previous Stripe methods, which return JSON and are buffered into memory and parsed.

Similar to https://github.com/stripe/stripe-node/pull/1157/files

I ran [japi-compliance-checker](https://github.com/lvc/japi-compliance-checker) and got the following
[report](https://gitcdn.xyz/cdn/richardm-stripe/6d775af7cde1d3ddee39fdd68412fc27/raw/af5149016aee5fee1f6e715f2cf7790739a0585c/compat.html) -- no incompatible changes.

image


src/main/java/com/stripe/net/StripeResponseGetter.java
```diff
@@ -20,4 +21,14 @@
...
       throws StripeException;
+
+  default InputStream requestStream(
+      ApiResource.RequestMethod method,
+      String url,
+      Map<String, Object> params,
+      RequestOptions options)
+      throws StripeException {
+    throw new UnsupportedOperationException(
+        "requestStream; is unimplemented for this StripeResponseGetter");
```
richardm-stripe, line 32:
```suggestion
        "requestStream is unimplemented for this StripeResponseGetter");
```

src/main/java/com/stripe/net/StripeResponseStream.java
```diff
@@ -0,0 +1,98 @@
...
+  public StripeResponseStream(int code, HttpHeaders headers, InputStream bodyStream) {
+    requireNonNull(headers);
+    requireNonNull(bodyStream);
+
+    this.code = code;
+    this.headers = headers;
+    this.bodyStream = bodyStream;
+  }
+
+  public StripeResponse unstream() throws IOException {
```
richardm-stripe, line 59:
```suggestion
  /**
   * Buffers the entire response body into a string, constructing the appropriate StripeResponse
   * @return the StripeResponse
   */
  public StripeResponse unstream() throws IOException {
```

richardm-stripe, line 63: Bikeshedding on this name encouraged.

src/main/java/com/stripe/net/HttpClient.java
```diff
@@ -38,13 +38,24 @@ protected HttpClient() {}
...
+   * Sends the given request to Stripe's API, streaming the response body.
    *
    * @param request the request
    * @return the response
    * @throws StripeException If the request fails for any reason
    */
-  public StripeResponse requestWithTelemetry(StripeRequest request) throws StripeException {
+  public StripeResponseStream requestStream(StripeRequest request) throws StripeException {
+    throw new UnsupportedOperationException(
+        "streamingRequest is unimplemented for this HttpClient");
```
dcr-stripe, line 49: streamingRequest -> requestStream?

src/main/java/com/stripe/net/HttpClient.java
```diff
@@ -63,23 +74,40 @@ public StripeResponse requestWithTelemetry(StripeRequest request) throws StripeE
...
-   * problems.
+   * Sends the given request to Stripe's API, handling telemetry if not disabled.
    *
    * @param request the request
    * @return the response
    * @throws StripeException If the request fails for any reason
    */
-  public StripeResponse requestWithRetries(StripeRequest request) throws StripeException {
+  public StripeResponse requestWithTelemetry(StripeRequest request) throws StripeException {
+    return sendWithTelemetry(request, (r) -> this.request(r));
```
dcr-stripe, line 84: Nit: Can use method references here throughout if you want to save some typing:

```
(r) -> this.request(r)
```

can become

```
this::request
```

src/main/java/com/stripe/net/StripeResponseGetter.java
```diff
@@ -20,4 +21,14 @@
       Class<T> clazz,
       RequestOptions options)
       throws StripeException;
+
+  default InputStream requestStream(
```
dcr-stripe, line 25: We do lose some type safety here since it's just a generic InputStream, but I'm not sure what we can really do about that... presumably the few callers that use this will be aware of how they should process the stream. If we wanted, we could make some typed wrapper around InputStream, but then the issue becomes compatibility with functions take operate on streams so I don't think it's worth it.

src/main/java/com/stripe/net/StripeResponseStream.java
dcr-stripe: `readStream`? `readFromStream`?

src/main/java/com/stripe/net/StripeResponseInterface.java
```diff
@@ -0,0 +1,37 @@
+package com.stripe.net;
+
+import java.time.Instant;
+
+/** Common interface representing an HTTP response from Stripe. */
+interface StripeResponseInterface {
```
dcr-stripe, line 6: It seems like the majority of this implementation would be shared right (ie. all the header extraction)? If so, I think we could make some abstract class with the shared implementation to avoid duplicating the code and have both implementations extend this instead of (or along with) the interface. This might be nice for testing as well.

dcr-stripe (commented):
I think this makes sense overall - thanks for putting this together! My biggest feedback would be around the duplicated implementation for the response interface. Probably worth sharing that for now. Otherwise I think this all looks good!

src/main/java/com/stripe/net/StripeResponseStream.java
```diff
@@ -0,0 +1,30 @@
+package com.stripe.net;
+
+import com.stripe.util.StreamUtils;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class StripeResponseStream extends AbstractStripeResponse<InputStream> {
```
dcr-stripe, line 7: Nit: StripeResponseStream -> StripeStreamResponse for consistency with other libraries?

dcr-stripe (approved):
LG! Just one naming nit!

PR #1208, opened 2021-05-14
API Updates

richardm-stripe:
Codegen for openapi abeda00.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* New method: `PaymentIntent#verify_microdeposits`
* New method: `SetupIntent#verify_microdeposits`


remi-stripe (approved):
This one looks good to me and we don't always add tests in java so I'm fine without one.

PR #1227, opened 2021-06-29
API Updates

dcr-stripe:
Codegen for openapi 3237fde.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Added support for `boleto_payments` on `Account.capabilities`
* Added support for `boleto` and `oxxo` on `SessionCreateParams` and `Session`

### DIFF
--- a/src/main/java/com/stripe/model/checkout/Session.java
+++ b/src/main/java/com/stripe/model/checkout/Session.java
@@ -490,10 +490,10 @@ public class Session extends ApiResource implements HasId {
     AcssDebit acssDebit;
 
     @SerializedName("boleto")
-    PaymentIntent.PaymentMethodOptions.Boleto boleto;
+    Boleto boleto;
 
     @SerializedName("oxxo")
-    PaymentIntent.PaymentMethodOptions.Oxxo oxxo;
+    Oxxo oxxo;
 
     @Getter
     @Setter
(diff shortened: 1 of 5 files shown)


src/main/java/com/stripe/model/checkout/Session.java
```diff
@@ -489,6 +489,12 @@ public LineItemCollection listLineItems(SessionListLineItemsParams params, Reque
     @SerializedName("acss_debit")
     AcssDebit acssDebit;
 
+    @SerializedName("boleto")
+    PaymentIntent.PaymentMethodOptions.Boleto boleto;
+
+    @SerializedName("oxxo")
+    PaymentIntent.PaymentMethodOptions.Oxxo oxxo;
+
```
remi-stripe, line 497: Those two are wrong. This should not be shared with the `PaymentIntent` class' options. It should be a separate resource internally in pay-server so that when they diverge over time we don't cause breaking changes. This will require changes internally first.

remi-stripe (changes requested):

src/main/java/com/stripe/model/checkout/Session.java
dcr-stripe: Great catch. Thanks Remi!

dcr-stripe (commented):
Thanks @remi-stripe  - PTAL?

remi-stripe (approved):

PR #1231, opened 2021-07-08
API Updates

richardm-stripe:
Codegen for openapi 780fbc3.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new resource `Quote`
* Add support for `quote` on `Invoice`

### DIFF
--- a/src/main/java/com/stripe/param/QuotePdfParams.java
+++ b/src/main/java/com/stripe/param/QuotePdfParams.java
@@ -1,0 +1,98 @@
+// File generated from our OpenAPI spec
+package com.stripe.param;
+
+import com.google.gson.annotations.SerializedName;
+import com.stripe.net.ApiRequestParams;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Getter;
+
+@Getter
+public class QuotePdfParams extends ApiRequestParams {
+  /** Specifies which fields in the response should be expanded. */
+  @SerializedName("expand")
+  List<String> expand;
+
+  /**
+   * Map of extra parameters for custom features not available in this client library. The content
+   * in this map is not serialized under this field's {@code @SerializedName} value. Instead, each
+   * key/value pair is serialized as if the key is a root-level field (serialized) name in this
+   * param object. Effectively, this map is flattened to its parent instance.
+   */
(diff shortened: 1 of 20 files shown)


richardm-stripe:
This LGTM now! r? @dcr-stripe

src/main/java/com/stripe/param/QuotePdfParams.java
```diff
@@ -0,0 +1,98 @@
...
+    public QuotePdfParams build() {
+      return new QuotePdfParams(this.expand, this.extraParams);
+    }
+
+    /**
+     * Add an element to `expand` list. A list is initialized for the first `add/addAll` call, and
+     * subsequent calls adds additional elements to the original list. See {@link
+     * QuotePdfParams#expand} for the field documentation.
+     */
+    public Builder addExpand(String element) {
```
dcr-stripe, line 51: I wonder if we want these expand properties on the /pdf endpoint? It doesn't really make sense but I suppose the implementation could differ based on the expand request...

src/test/java/com/stripe/functional/QuoteTest.java
```diff
@@ -0,0 +1,114 @@
...
+import com.stripe.model.Quote;
+import com.stripe.model.QuoteCollection;
+import com.stripe.net.ApiResource;
+import com.stripe.util.StreamUtils;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class QuoteTest extends BaseStripeTest {
```
dcr-stripe, line 17: Do we want to add tests for `Computed`?

dcr-stripe (approved):
Overall LGTM! Thanks Richard!

src/main/java/com/stripe/param/QuotePdfParams.java
richardm-stripe: The problem originates with the OpenAPI spec which too eagerly defines "expand" params. Out of scope for now.

src/test/java/com/stripe/functional/QuoteTest.java
richardm-stripe: I've done some testing manually. The [fixture](https://github.com/stripe/openapi/blob/master/openapi/fixtures3.yaml#L1846) that we have from stripe-mock doesn't include line_items, so the regression test wouldn't be a very useful one really.

PR #1251, opened 2021-08-30
Test Java 14 and 15

ob-stripe:
cc @stripe/api-libraries 

Test Java 14 and 15, and bump all dependencies.

Note that Java 16 is out, and Java 17 will be out soonish. However these versions require Gradle 7, and upgrading to Gradle 7 will necessitate rewriting `deploy.gradle` to use the `maven-publish` plugin instead of `maven` (cf. https://docs.gradle.org/6.9.1/userguide/upgrading_version_5.html#legacy_publication_system_is_deprecated_and_replaced_with_the_publish_plugins). This is a non-trivial change that will be somewhat annoying to test, so I'm leaving it to the client libraries team 😛


seang-stripe (approved):
Looks good!

PR #1254, opened 2021-08-31
API Updates

richardm-stripe:
Codegen for openapi 944e56f.
r? @remi-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `after_expiration`, `consent_collection`, and `expires_at` on `CheckoutSessionCreateParams` and `Checkout.Session`
* Add support for `consent` and `recovered_from` on `Checkout.Session`

### DIFF
--- a/src/main/java/com/stripe/model/checkout/Session.java
+++ b/src/main/java/com/stripe/model/checkout/Session.java
@@ -452,6 +452,37 @@ public class Session extends ApiResource implements HasId {
     /** When set, configuration used to recover the Checkout Session on expiry. */
     @SerializedName("recovery")
     Recovery recovery;
+
+    @Getter
+    @Setter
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/model/checkout/Session.java
```diff
@@ -578,6 +640,37 @@ public LineItemCollection listLineItems(SessionListLineItemsParams params, Reque
     }
   }
 
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class Recovery extends StripeObject {
```
remi-stripe, line 646: This should be inside `AfterExpiration`

remi-stripe (changes requested):

remi-stripe (approved):

PR #1255, opened 2021-09-02
API Updates

richardm-stripe:
Codegen for openapi 9e4d34e.
r? @remi-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `future_requirements` on `Account`, `Capability`, and `Person`
* Add support for `alternatives` on `Account.requirements`, `Capability.requirements`, and `Person.requirements`

### DIFF
--- a/src/main/java/com/stripe/model/Account.java
+++ b/src/main/java/com/stripe/model/Account.java
@@ -1096,7 +1096,7 @@ public class Account extends ApiResource implements MetadataStore<Account>, Paym
      * verification failed.
      */
     @SerializedName("errors")
-    List<Account.FutureRequirements.Error> errors;
+    List<Account.FutureRequirements.Errors> errors;
 
     /**
      * Fields that need to be collected assuming all volume thresholds are reached. As they become
@@ -1141,7 +1141,7 @@ public class Account extends ApiResource implements MetadataStore<Account>, Paym
     @Getter
     @Setter
     @EqualsAndHashCode(callSuper = false)
-    public static class Error extends StripeObject {
+    public static class Errors extends StripeObject {
       /**
        * The code for the type of error.
        *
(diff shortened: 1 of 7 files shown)


src/main/java/com/stripe/model/Account.java
```diff
@@ -1117,7 +1261,7 @@ public void setFrontObject(File expandableObject) {
      * verification failed.
      */
     @SerializedName("errors")
-    List<Account.Requirements.Errors> errors;
+    List<Account.Requirements.Error> errors;
```
remi-stripe, line 1264: I think this change is a breaking change and is incorrect though you're right that it should have been the name of the class :(

remi-stripe (changes requested):

richardm-stripe:
@remi-stripe should be fixed now

src/main/java/com/stripe/model/Capability.java
```diff
@@ -154,7 +298,7 @@ public Capability update(CapabilityUpdateParams params, RequestOptions options)
      * verification failed.
      */
     @SerializedName("errors")
-    List<Account.Requirements.Errors> errors;
+    List<Capability.Requirements.Errors> errors;
```
remi-stripe, line 301: That's also a breaking change right? It's correct but you can't change this without breaking people

src/main/java/com/stripe/model/Person.java
```diff
@@ -343,7 +477,7 @@ public Person delete(Map<String, Object> params, RequestOptions options) throws
      * verification failed.
      */
     @SerializedName("errors")
-    List<Account.Requirements.Errors> errors;
+    List<Person.Requirements.Errors> errors;
```
remi-stripe, line 480: Same here

remi-stripe (changes requested):

remi-stripe (approved):

PR #1261, opened 2021-09-10
Upgrade Gradle

ob-stripe:
r? @richardm-stripe
cc @stripe/api-libraries 

- Upgrade to Gradle 7
- Replace `enabled` with `required`: https://docs.gradle.org/7.2/dsl/org.gradle.api.reporting.Report.html#org.gradle.api.reporting.Report:enabled
- Remove `config` values from `lombok` block. There is little context in the [PR that deprecates `config`](https://github.com/freefair/gradle-plugins/pull/312) but it looks like the plugin no longer generates the `lombok.config` file and users are supposed to commit the file directly (which we were already doing anyway).

Unfortunately we cannot yet test Java 16+ because google-java-format requires special configuration with Java 16: https://github.com/google/google-java-format#jdk-16. That configuration can be set in gradle via the [`org.gradle.jvmargs`](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties) parameter in `gradle.properties`, but of course that fails with Java 1.8... We need a way to conditionally set the parameter depending on Java version and I couldn't find one in 5 minutes of googling.


richardm-stripe:
Thank you @ob!

richardm-stripe:
Thank you @ob-stripe!

PR #1273, opened 2021-10-15
API Updates

dcr-stripe:
Codegen for openapi 1f5d304.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Change type of `UsageRecordCreateParams.timestamp` from `integer` to `literal('now') | integer`
* Change `UsageRecordCreateParams.timestamp` to be optional


dcr-stripe:
Yep agreed - closing this as it's a no-op otherwise.

PR #1290, opened 2021-11-24
Update Javadoc task to not use module directories.

dcr-stripe:
r? @richardm-stripe 
cc @stripe/api-libraries 

### Summary
Updates the `javadoc` task to explicitly specify we're not using module directories. By default the task attempts to use them, searching for `module-info.java` files.

Since we don't use these currently, the end result is we had broken links in our Javadoc where clicking on a search result tried going to the `undefined` module, resulting in a URL such as: https://stripe.dev/stripe-java/undefined/com/stripe/model/Account.html

### Test plan

I build Javadoc before and after this change using `./gradlew javadoc` and then tried opening the locally built index.html (`build/docs/javadoc/index.html`) in my browser and searching.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,11 @@ coveralls {
 
 javadoc {
   doFirst {
+    // We should remove this once we start using module-info.java.
+    // Without that file, Javadoc will fall back to "undefined" as the module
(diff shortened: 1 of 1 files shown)


richardm-stripe (approved):
Thank you!

PR #1291, opened 2021-12-02
Test Java 16 and 17

ob-stripe:

Bump Gradle and dependencies, and start testing Java 16 and 17.

Due to https://github.com/google/google-java-format#jdk-16 I had to add some `--add-exports` arguments to the JVM via the `org.gradle.jvmargs` variable in `gradle.properties`.

However Java 1.8 does not support `--add-exports`, so what I did is add a conditional step in `.github/workflows/ci.yml` to comment out these arguments when testing Java 1.8. It's kinda gross but I could not find a better way to do this.

Note that this does mean that running the test suite locally with Java 1.8 will fail unless you go and manually comment out the `--add-exports` lines in `gradle.properties`.


dcr-stripe (approved):
LGTM! Thanks Olivier!

PR #1297, opened 2021-12-17
API Updates

dcr-stripe:
Codegen for openapi 26c0399.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `processing` on `PaymentIntent`

### DIFF
--- a/src/main/java/com/stripe/model/PaymentIntent.java
+++ b/src/main/java/com/stripe/model/PaymentIntent.java
@@ -1073,11 +1073,6 @@ public class PaymentIntent extends ApiResource implements HasId, MetadataStore<P
   @Getter
   @Setter
   @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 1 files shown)


src/main/java/com/stripe/model/PaymentIntent.java
```diff
@@ -1066,6 +1070,11 @@ public PaymentIntent capture(PaymentIntentCaptureParams params, RequestOptions o
         ApiResource.RequestMethod.POST, url, params, PaymentIntent.class, options);
   }
 
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class CardProcessingResource extends StripeObject {}
```
dcr-stripe, line 1076: I don't think this is actually the desired name here...

remi-stripe (approved):

PR #1300, opened 2021-12-23
API Updates

yejia-stripe:
Codegen for openapi d70de34.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `mandate_options` on `SubscriptionCreateParams.payment_settings.payment_method_options.card`, `SubscriptionUpdateParams.payment_settings.payment_method_options.card`, and `Subscription.payment_settings.payment_method_options.card`

### DIFF
--- a/src/main/java/com/stripe/model/Invoice.java
+++ b/src/main/java/com/stripe/model/Invoice.java
@@ -1592,6 +1592,30 @@ public class Invoice extends ApiResource implements HasId, MetadataStore<Invoice
        */
       @SerializedName("request_three_d_secure")
       String requestThreeDSecure;
+
+      @Getter
+      @Setter
+      @EqualsAndHashCode(callSuper = false)
+      public static class MandateOptions extends StripeObject {
+        /** Amount to be charged for future payments. */
+        @SerializedName("amount")
+        Long amount;
+
+        /**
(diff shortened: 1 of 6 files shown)


adityatandon007:
@yejia-stripe can we also pass this `mandate_options` while creating the payment intent. I couldn't find any parameters to enable eMandate in the documentation regarding eMandate. Its great that with this PR we can have eMandate setup for recurring payments for INR but since I am using a different service for handling my subscriptions and dunning, it would be great if you can add the support for eMandate while creating intent. 
P.S. I was gonna raise an issue for this but then saw this PR which adds support for eMandate but only in subscription api not in the intent api.

richardm-stripe (approved):

src/main/java/com/stripe/model/Invoice.java
```diff
@@ -1592,6 +1592,30 @@ public void setDiscountObject(Discount expandableObject) {
        */
       @SerializedName("request_three_d_secure")
       String requestThreeDSecure;
+
+      @Getter
+      @Setter
+      @EqualsAndHashCode(callSuper = false)
+      public static class MandateOptions extends StripeObject {
```
richardm-stripe, line 1599: Where is this referenced?

dcr-stripe (approved):
LGTM!

PR #1302, opened 2022-01-19
API Updates

dcr-stripe:
Codegen for openapi 7734045.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `bacs_debit` and `eps` on `PaymentIntentCreateParams.payment_method_options`, `PaymentIntentUpdateParams.payment_method_options`, `PaymentIntentConfirmParams.payment_method_options`, and `PaymentIntent.payment_method_options`
* Add support for `image_url_png` and `image_url_svg` on `PaymentIntent.next_action.wechat_pay_display_qr_code`

### DIFF
--- a/src/main/java/com/stripe/model/PaymentIntent.java
+++ b/src/main/java/com/stripe/model/PaymentIntent.java
@@ -1141,6 +1141,20 @@ public class PaymentIntent extends ApiResource implements HasId, MetadataStore<P
       /** The base64 image data for a pre-generated QR code. */
       @SerializedName("image_data_url")
       String imageDataUrl;
+
+      /**
+       * The image_url_png string used to render QR code, can be used as &lt;img src=&quot;…&quot;
+       * /&gt;.
+       */
+      @SerializedName("image_url_png")
+      String imageUrlPng;
+
+      /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/stripe/model/PaymentIntent.java
```diff
@@ -1141,6 +1141,20 @@ public PaymentIntent capture(PaymentIntentCaptureParams params, RequestOptions o
       /** The base64 image data for a pre-generated QR code. */
       @SerializedName("image_data_url")
       String imageDataUrl;
+
+      /**
+       * The image_url_png string used to render QR code, can be used as &lt;img src=&quot;…&quot;
```
yejia-stripe, line 1146: Is the `<`/`>`/`"` replacement easy to fix? Or is this expected?

yejia-stripe (approved):

src/main/java/com/stripe/model/PaymentIntent.java
dcr-stripe: Similar to .NET, we need to encode these since otherwise the Javadoc interpreter will try to interpret this comment and we're giving this as a code snippet (ie. we want users to explicitly see `` and not a rendered image).

PR #1319, opened 2022-03-01
API Updates

richardm-stripe:
Codegen for openapi 07922e0.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `deletes_after` on `TestHelpers.TestClock`


src/main/java/com/stripe/model/SetupIntentTypeSpecificPaymentMethodOptionsClient.java
```diff
@@ -0,0 +1,11 @@
...
+package com.stripe.model;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class SetupIntentTypeSpecificPaymentMethodOptionsClient extends StripeObject {}
```
dcr-stripe, line 11: What is this class - do we need to add something in the changelog?

dcr-stripe (changes requested):

src/main/java/com/stripe/model/SetupIntentTypeSpecificPaymentMethodOptionsClient.java
richardm-stripe: It should not be generated. There's a bug in the spec.

dcr-stripe (approved):
LGTM!

PR #1321, opened 2022-03-04
Delete PaymentIntentTypeSpecificPaymentMethodOptionsClient.java

yejia-stripe:
This type should not have been generated and slipped in accidentally. It corresponds to a type that appears on client-side-only responses and so is not appropriate for stripe-java.

This type is not referenced by any other type, and won't be returned by any method or included on any event payload, so removing it should be a safe change. Removing would be a breaking change for any user referencing these types, however we are extremely certain that nobody would have the need or occasion to do this, so we will not be bumping the major version.

r? @richardm-stripe


richardm-stripe (approved):
LGTM! Took the liberty of adding a bit more detail to the PR description.

PR #1326, opened 2022-03-16
Update search results types and remove beta warning

pakrym-stripe:
Adjusts the field name and removes beta comments.

https://jira.corp.stripe.com/browse/DEVSDK-524

r? @dcr-stripe


dcr-stripe (commented):
This looks good to me - thanks @pakrym-stripe !

Could we add a test for the iterator if it's not too tricky? I know we should've done this before :( But it would be awesome to be able to test this.

pakrym-stripe:
Added tests.

PR #1327, opened 2022-03-18
API Updates

dcr-stripe:
Codegen for openapi ba66006.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `status` on `Card`


src/main/java/com/stripe/model/PaymentIntent.java
```diff
@@ -2338,25 +2362,25 @@ public void setPackage(String package_) {
...
-         * Whether customer approval has been requested for this payment. For payments greater than
-         * INR 5000 or mandate amount, the customer must provide explicit approval of the payment
-         * with their bank.
-         */
-        @SerializedName("approval_requested")
-        Boolean approvalRequested;
+  @Getter
+  @Setter
+  @EqualsAndHashCode(callSuper = false)
+  public static class ProcessingCustomerNotification extends StripeObject {
```
dcr-stripe, line 2372: This is a breaking change - we likely need an override here.

yejia-stripe (approved):

PR #1330, opened 2022-03-25
API Updates

pakrym-stripe:
Codegen for openapi 2956fca.
r? @dcr-stripe 
cc @stripe/api-libraries

## Changelog
* Add support for Search API
  * Add support for `search` method on resources `Charge`, `Customer`, `Invoice`, `PaymentIntent`, `Price`, `Product`, and `Subscription`

### DIFF
--- a/src/main/java/com/stripe/model/StripeSearchResult.java
+++ b/src/main/java/com/stripe/model/StripeSearchResult.java
@@ -18,8 +18,12 @@ public abstract class StripeSearchResult<T> extends StripeObject
   @Getter(onMethod_ = {@Override})
   List<T> data;
 
+  /**
+   * Get the total count of search records in the result. The value is present when `total_count` is
+   * added to the `expand` search parameter.
+   */
   @Getter(onMethod_ = {@Override})
-  Integer totalCount;
+  Long totalCount;
 
   @Getter(onMethod_ = {@Override})
   Boolean hasMore;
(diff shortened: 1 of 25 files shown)


src/main/java/com/stripe/model/StripeSearchResult.java
```diff
@@ -18,6 +18,9 @@
   @Getter(onMethod_ = {@Override})
   List<T> data;
 
+  @Getter(onMethod_ = {@Override})
+  Integer totalCount;
```
dcr-stripe, line 22: I believe this should be a `Long` and not `Integer` (https://github.com/stripe/stripe-java/commit/b656eac45ab1e843b508dc5dbb080e916441b0ae for what it used to be in lists).

Also we should add a comment indicating that it's only present if the user explicitly expands `total_count`.

dcr-stripe (changes requested):
Tests are failing for formatting, so I think you might need to run the formatter.

Also - we should specify an accurate Changelog describing what this is doing.

src/test/java/com/stripe/functional/ChargeTest.java
```diff
@@ -148,6 +151,14 @@ public void testMarkSafeWithTypedParams() throws StripeException {
         fraudDetailsUntypedParam("safe"));
   }
 
+  @Test
+  public void testSearch() throws Exception {
+    ChargeSearchResult result =
+        Charge.search(ChargeSearchParams.builder().setQuery("currency=USD").build());
```
dcr-stripe, line 157: Nit: Let's do `currency:"USD"`. This isn't a valid query as is and you never know if someone will use his test to understand how to use the API.

dcr-stripe (approved):
LGTM - just one nit!

dcr-stripe (approved):
LGTM!

PR #1331, opened 2022-03-28
Check JDK version

pakrym-stripe:


.github/workflows/ci.yml
```diff
@@ -57,7 +77,7 @@ jobs:
         run: docker run -d -p 12111-12112:12111-12112 stripe/stripe-mock && sleep 5
 
       - name: Run test suite
-        run: ./gradlew check
```
pakrym-stripe, line 60: Check runs spotless that we already do in a separate config.

.github/workflows/ci.yml
```diff
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
pakrym-stripe, line 7: makes it so we don't run all configurations twice

src/test/java/com/stripe/functional/ChargeTest.java
```diff
@@ -152,11 +153,8 @@ public void testMarkSafeWithTypedParams() throws StripeException {
   }
 
   @Test
-  public void testSearch() throws Exception {
-    ChargeSearchResult result =
-        Charge.search(ChargeSearchParams.builder().setQuery("currency:\"USD\"").build());
-    assertEquals(1, result.getData().size());
-    assertEquals(1, result.getTotalCount());
+  public void testSearch() {
+    assertNull(System.getProperty("java.version"));
```
pakrym-stripe, line 157: This is to verify that we are actually running on the selected java version.

dcr-stripe: Thanks for doing this to illustrate the change! Let's make sure to revert the test.

One thing that's confusing to me is the `Display version` step in GH also uses Gradle but always shows JDK 17.

Maybe we could change it to show the accurate runtime version here? Or print this as part of the test suite?

dcr-stripe (changes requested):
LGTM overall, just the comment around Display version. Thanks for investigating Pavel!

Let's make sure to update this PR description to capture:
1. The GH action changes we're doing as part of this.
2. Why this is not a breaking change for existing users.

src/test/java/com/stripe/functional/ChargeTest.java
pakrym-stripe: Added test_home to the display version output:
image

pakrym-stripe:
PTAL @dcr-stripe ?

dcr-stripe (approved):
LGTM.

Let's make sure everyone on the team is updated.

What might also be handy is to do a check that a jar built before and after this PR has no real diff.

dcr-stripe (commented):
Oh let's also update the README's Development section to document this requirement.

pakrym-stripe:
Updated to Java 18, added 18 as a test target, pinned JDK to an exact version, fixed spotless.


PTAL @dcr-stripe ?

build.gradle
```diff
@@ -22,20 +22,26 @@ plugins {
     id "signing"
     id "jacoco"
     id "io.freefair.lombok" version "6.3.0"
-    id "com.diffplug.spotless" version "6.0.1"
+    id "com.diffplug.spotless" version "6.4.0"
```
pakrym-stripe, line 25: The old version didn't like Java 18.

PR #1347, opened 2022-05-04
API Updates

dcr-stripe:
Codegen for openapi 158cf6c.
r? @kamil-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `registered_address` on `AccountCreateParams.individual`, `AccountUpdateParams.individual`, `PersonCreateParams`, `PersonUpdateParams`, `Person`, `TokenCreateParams.account.individual`, and `TokenCreateParams.person`
* Add support for `payment_method_data` on `SetupIntentConfirmParams`, `SetupIntentCreateParams`, and `SetupIntentUpdateParams`

### DIFF
--- a/src/main/java/com/stripe/param/PersonUpdateParams.java
+++ b/src/main/java/com/stripe/param/PersonUpdateParams.java
@@ -134,6 +134,10 @@ public class PersonUpdateParams extends ApiRequestParams {
   @SerializedName("political_exposure")
   Object politicalExposure;
 
+  /** The person's registered address. */
+  @SerializedName("registered_address")
+  RegisteredAddress registeredAddress;
+
   /** The relationship that this person has with the account's legal entity. */
   @SerializedName("relationship")
   Relationship relationship;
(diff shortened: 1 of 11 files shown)


src/main/java/com/stripe/param/PersonUpdateParams.java
```diff
@@ -1804,6 +1819,198 @@ public Builder addAllFile(List<String> elements) {
     }
   }
 
+  @Getter
+  public static class RegisteredAddress {
+    /** City, district, suburb, town, or village. */
+    @SerializedName("city")
+    Object city;
```
kamil-stripe, line 1826: Should that all be objects? Should we share Address here?

dcr-stripe: Good question!

> Should that all be objects? 

We use Object here since it's emptyable. Below we have overloaded methods to set one or the other:

```java

        /** City, district, suburb, town, or village. */
        public Builder setCity(String city) {
          this.city = city;
          return this;
        }

        /** City, district, suburb, town, or village. */
        public Builder setCity(EmptyParam city) {
          this.city = city;
          return this;
        }

```

This is consistent with other uses of Address eg. https://github.com/stripe/stripe-java/blob/41530ea4/src/main/java/com/stripe/param/AccountUpdateParams.java#L721

> Should we share Address here?

We don't share Address in params, just in the response type.

kamil-stripe (approved):

PR #1351, opened 2022-05-11
API Updates

richardm-stripe:
Codegen for openapi 7789931.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `description` on `CheckoutSessionCreateParams.subscription_data`, `SubscriptionCreateParams`, `SubscriptionUpdateParams`, and `Subscription`
* Add support for `consent_collection`, `payment_intent_data`, `shipping_options`, `submit_type`, and `tax_id_collection` on `PaymentLinkCreateParams` and `PaymentLink`
* Add support for `customer_creation` on `PaymentLinkCreateParams`, `PaymentLinkUpdateParams`, and `PaymentLink`
* Add support for `metadata` on `SubscriptionSchedule.phases[]`, `SubscriptionScheduleCreateParams.phases[]`, and `SubscriptionScheduleUpdateParams.phases[]`
* Add support for new value `billing_portal.session.created` on enums `WebhookEndpointCreateParams.enabled_events[]` and `WebhookEndpointUpdateParams.enabled_events[]`

### DIFF
--- a/src/main/java/com/stripe/model/PaymentLink.java
+++ b/src/main/java/com/stripe/model/PaymentLink.java
@@ -61,6 +61,18 @@ public class PaymentLink extends ApiResource implements HasId, MetadataStore<Pay
   @SerializedName("billing_address_collection")
   String billingAddressCollection;
 
+  /** When set, provides configuration to gather active consent from customers. */
+  @SerializedName("consent_collection")
+  ConsentCollection consentCollection;
+
+  /**
+   * Configuration for Customer creation during checkout.
+   *
+   * <p>One of {@code always}, or {@code if_required}.
+   */
+  @SerializedName("customer_creation")
+  String customerCreation;
+
   /** Unique identifier for the object. */
   @Getter(onMethod_ = {@Override})
   @SerializedName("id")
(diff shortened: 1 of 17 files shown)


src/main/java/com/stripe/model/PaymentLink.java
```diff
@@ -119,13 +135,29 @@ public class PaymentLink extends ApiResource implements HasId, MetadataStore<Pay
   @SerializedName("shipping_address_collection")
   ShippingAddressCollection shippingAddressCollection;
 
+  /** The shipping rate options applied to the session. */
+  @SerializedName("shipping_options")
+  List<PaymentLink.ShippingOption> shippingOptions;
```
remi-stripe, line 140: Why is there a prefix here? Isn't it the default to use the local inner class?

richardm-stripe, line 140: This is fine even though it's not dequalified.

remi-stripe (approved):
Left one question but I don't think it prevents merging

src/main/java/com/stripe/model/PaymentLink.java
richardm-stripe: The generator apparently derps when the reference is inside a type parameter. Harmless but worth fixing because it makes the eyeball test more reliable.

PR #1355, opened 2022-05-20
Use github.actor as GRGIT_USER

pakrym-stripe:
Attempt to fix docs publishing error. 

We were using GITHUB_TOKEN as GRGIT_USER but I think we need to have both the username and the password.


dcr-stripe (approved):
LGTM!

PR #1361, opened 2022-05-23
Fix version update script and the version in README.md

pakrym-stripe:
We were accidentally stripping the quote.


dcr-stripe (approved):
LGTM, thank you for fixing!

Issue #1358, opened 2022-05-20
invalid_request_error mapped to CardException not InvalidRequestException


jstafford406:
### Describe the bug

[handleApiError](https://github.com/stripe/stripe-java/blob/ecb791f567f85377f60b272eb9f7a4d91bfb67b7/src/main/java/com/stripe/net/LiveStripeResponseGetter.java#L203) maps all 402 status code responses to `CardException`. We have received responses that in the Stripe dashboard are "type": "invalid_request_error", but result in a `CardException`. (The CardException has a null `code` which causes us problems.) An example of the Response body is shown further below.

I have been in touch with Stripe support who say "our error documentation doesn't map status codes to specific error types, so the invalid_request_error may not always go hand in hand with the 404 status code.". So I think `handleApiError` needs to allow for `invalid_request_error` type in the `case 402:` branch.

Example of Response body for `invalid_request_error` type which results in `CardException`:

```Response body
{
  "error": {
    "message": "The payment is blocked because the used IBAN has been suspended. You can learn more about this at https://support.stripe.com/questions/sepa-debit-chargebacks-faq and if you need assistance processing a payment please contact us via https://support.stripe.com/contact/ for further information.",
    "type": "invalid_request_error"
  }
}```

### To Reproduce

Using the Charges API create a Charge with an underlying payment source where the associated IBAN has been disputed 3 times in row. Such that Stripe block the Charge and return an invalid_request_error response. See https://support.stripe.com/questions/sepa-debit-chargebacks-faq

### Expected behavior

Expected the Stripe Java SDK to translate the invalid_request_error with HTTP status code 402 into an InvalidRequestException. 
Actual behavior: CardException where only the `message` property is populated, notably not the `code`.

### Code snippets

_No response_

### OS

macOS

### Java version

Java 11

### stripe-java version

20.121.0

### API version

2020-08-27

### Additional context

_No response_

jstafford406:
> Is there a way you can workaround this problem for the time being?

Thanks for following up! Yes we have a workaround for the most pressing case - we catch the `CardException` and look in the `message` for `"The payment is blocked because the used IBAN has been suspended"` to identify when we should have got an `InvalidRequestException`.

richardm-stripe:
We didn't get to this in https://github.com/stripe/stripe-java/releases/tag/v22.0.0, unfortunately. We will consider again for the next major version.

PR #1362, opened 2022-05-24
Determine exception type by error type first

pakrym-stripe:
Right now we are using status codes to determine the exception type which causes issues like https://github.com/stripe/stripe-java/issues/1358 where `invalid_request_error` was returned with `402` status code.

Change the error-handling logic to look at the error type first but keep the old status code-based logic as a fallback.

Add a bunch of tests.

Fixes: https://github.com/stripe/stripe-java/issues/1358


pakrym-stripe:
This will have to be done consistently across all languages.

Issue #1364, opened 2022-05-26
Missing test helpers methods in ReceivedCredits and ReceivedDebits


kamil-stripe:
### Describe the bug

Currently released stripe-java library misses two test helper methods matching following URLs:
* POST /v1/test_helpers/treasury/received_credits
* POST /v1/test_helpers/treasury/received_debits

Objects and methods missing in our Java code:
* ReceivedCredit.getTestHelpers()
* ReceivedCredit.TestHelpers
* ReceivedDebit.getTestHelpers()
* ReceivedDebit.TestHelpers

We're still working to fix that omission.

### To Reproduce

```
ReceivedCreditCreateParams params =
  ReceivedCreditCreateParams.builder().build();

ReceivedCredit receivedCredit = ReceivedCredit.create(params);
```

### Expected behavior

The call above simulates a test mode received credit initiated by a third party.

### OS

N/A

### Java version

Java 8+

### stripe-java version

20.123.0

### API version

2020-08-27

PR #1367, opened 2022-06-01
Add treasury test helpers

pakrym-stripe:
Fixes: https://github.com/stripe/stripe-java/issues/1364

## Changelog
* Add support for `create` test helper on `ReceivedCredit`.
* Add support for `create` test helper on `ReceivedDebit`.
* Deprecates the ability to directly create instances of inner `TestHelper` classes.


dcr-stripe (approved):
LGTM, though let's put more in this PR description using the standard changelog format. Specifically, we're adding support for new test helpers + deprecating an access pattern.

PR #1374, opened 2022-06-14
API Updates for beta branch

pakrym-stripe:
Codegen for openapi v154.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
Specs are identical


.github/workflows/ci.yml
```diff
@@ -3,11 +3,15 @@ name: CI
 on:
   workflow_dispatch: {}
   push:
-    branches: [master]
+    branches:
```
kamil-stripe, line 6: Weird that this got pulled into diff

pakrym-stripe: It's not, every time we update beta branch codegen we also merge the master in.

pakrym-stripe: This way we get latest fixes from master in case they are needed for newly generated code.

kamil-stripe: makes sense

dcr-stripe (approved):
LGTM!

Issue #1371, opened 2022-06-12
The source code for 20.128.0 is not what is compiled into classes


geniot:
### Describe the bug

When trying to debug EventDataObjectDeserializer public Optional<StripeObject> getObject() I hit comment lines.

### To Reproduce

1. Using Maven add dependency 20.128.0
2. Create a Java snippet:
Event event = Webhook.constructEvent(json, header, endpointSecret);
Optional<StripeObject> optional = event.getDataObjectDeserializer().getObject();
3. Using Stripe CLI hit a webhook endpoint.
4. Step into getDataObjectDeserializer.

### Expected behavior

I expect to step through getObject method code.

### Code snippets

```Java
@RequestMapping(value = "/stripeCheckoutSessionCompleted", method = POST, consumes = APPLICATION_JSON_VALUE, produces = APPLICATION_JSON_VALUE)
    @ResponseBody
    public String stripeWebhookEndpoint(HttpServletRequest request) {
        String header = request.getHeader("Stripe-Signature");
        try {
            String json = IOUtils.toString(request.getInputStream(), StandardCharsets.UTF_8);
            Event event = Webhook.constructEvent(json, header, endpointSecret);
            Optional<StripeObject> optional = event.getDataObjectDeserializer().getObject();
...
```

### OS

Windows 10

### Java version

11

### stripe-java version

20.128.0

### API version

2020-08-27

dcr-stripe:
Hi @geniot ,

Thanks for the report. Sorry you're having problems here! Our team will investigate and report back!

geniot:
@dcr-stripe , https://youtu.be/4dlLgrf9E1s Windows 10, Intellij 2021.2.4 Ultimate, default debug settings. Extract from Manifest:
Manifest-Version: 1.0
Bnd-LastModified: 1654809243609
Bundle-ManifestVersion: 2
Bundle-Name: stripe-java
Bundle-SymbolicName: stripe-java
Bundle-Version: 20.128.0
Created-By: 18.0.1 (Azul Systems, Inc.)
Export-Package: com.stripe;version="20.128.0"

dcr-stripe:
Interesting, this isn't reproducing for me.

I've seen this happen before if a package is accidentally pulling in multiple versions of stripe-java accidentally. Just confirming that you've done a clean rebuild (Build > Rebuild Project) here.

[...]

PR #1382, opened 2022-06-28
chore: Update gradle lombok plugin.

dcr-stripe:
Updates the lombok plugin to get the most recent Lombok stable in an attempt to fix https://github.com/stripe/stripe-java/issues/1371#issuecomment-1168977414. This will include some Lombok changes which fix delombok output not matching.

Note that this is only rolling this out in the beta SDKs until we've confirmed the fix.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ plugins {
     id "maven-publish"
     id "signing"
     id "jacoco"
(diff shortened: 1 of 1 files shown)


ramya-stripe:
Closing this PR as we got a confirmation that this was not enough to fix the issue

PR #1386, opened 2022-07-05
Support serializing nested objects in map

pakrym-stripe:
r? @kamil-stripe 

Currently we call `out.setSerializeNulls(true);` before serializing `Map`s. This worked well when values were strings but causes unexpected issues now that values can have full Params classes. One of the issues is that `extraParams` dictionary gets serialized with null value.

This PR changes our custom serializer to manually serialize items and enable writing null only when explicitly serializing a null map value.

### DIFF
--- a/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
+++ b/src/main/java/com/stripe/net/ApiRequestParamsConverter.java
@@ -13,6 +13,8 @@ import com.google.gson.stream.JsonWriter;
 import com.stripe.Stripe;
 import com.stripe.param.common.EmptyParam;
 import java.io.IOException;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/net/ApiRequestParamsConverter.java
```diff
@@ -86,34 +88,71 @@ private static void validateDuplicateKey(
...
+      for (Map.Entry<String, V> entry : value.entrySet()) {
+        out.name(entry.getKey());
+        V entryValue = entry.getValue();
+        if (entryValue == null) {
+          boolean oldSerializeNullsValue = out.getSerializeNulls();
+          try {
+            out.setSerializeNulls(true);
+            out.nullValue();
+          } finally {
+            out.setSerializeNulls(oldSerializeNullsValue);
```
kamil-stripe, line 144: Why is it done this way?

kamil-stripe (approved):

src/main/java/com/stripe/net/ApiRequestParamsConverter.java
pakrym-stripe: We want to restore the original value after we are done writing null. Using finally is the most reliable way to do it.

PR #1396, opened 2022-07-20
API Updates for beta branch

pakrym-stripe:
Codegen for openapi v167.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
- Updated stable APIs to the latest version
- Add `QuotePhase` resource.


src/main/java/com/stripe/model/Quote.java
```diff
@@ -940,193 +940,6 @@ public static class InvoiceSettings extends StripeObject {
     Long daysUntilDue;
   }
 
-  @Getter
-  @Setter
-  @EqualsAndHashCode(callSuper = false)
-  public static class QuotePhaseConfiguration extends StripeObject implements HasId {
```
dcr-stripe, line 946: Double checking we haven't published a beta version with this embedded class right?

dcr-stripe (approved):
LGTM!

dcr-stripe (changes requested):
un-LGTM.

Why does `QuotePhaseRetrieveParams` have pagination params?

src/main/java/com/stripe/model/Quote.java
pakrym-stripe: Nope.

PR #1401, opened 2022-07-22
Check for API key in autoPagingIterable

yejia-stripe:
r? @dcr-stripe @kamil-stripe 

## Summary
Check that `apiKey` is set (either globally or via request options) in `StripeCollection.autoPagingIterable` and `StripeSearchResult.autoPagingIterable`, and throw an exception if it is not.

### DIFF
--- a/src/main/java/com/stripe/exception/ApiKeyMissingException.java
+++ b/src/main/java/com/stripe/exception/ApiKeyMissingException.java
@@ -8,11 +8,6 @@ package com.stripe.exception;
 public class ApiKeyMissingException extends RuntimeException {
   private static final long serialVersionUID = 1L;
 
-  /** Initializes a new instance of the {@link ApiKeyMissingException}. */
-  public ApiKeyMissingException() {
-    super();
-  }
-
   /** Initializes a new instance of the {@link ApiKeyMissingException} with a message. */
   public ApiKeyMissingException(String message) {
     super(message);
--- a/src/main/java/com/stripe/exception/ApiKeyMissingException.java
+++ b/src/main/java/com/stripe/exception/ApiKeyMissingException.java
@@ -1,0 +1,20 @@
+package com.stripe.exception;
+
+/**
+ * {@link ApiKeyMissingException} is thrown when the API key is not set for a request. The API key
+ * for a request may be set either globally through {@link com.stripe.Stripe#apiKey} or through
+ * {@link com.stripe.net.RequestOptions}.
+ */
+public class ApiKeyMissingException extends RuntimeException {
+  private static final long serialVersionUID = 1L;
+
+  /** Initializes a new instance of the {@link ApiKeyMissingException}. */
+  public ApiKeyMissingException() {
+    super();
+  }
+
+  /** Initializes a new instance of the {@link ApiKeyMissingException} with a message. */
+  public ApiKeyMissingException(String message) {
+    super(message);
+  }
+}
--- a/src/main/java/com/stripe/model/StripeCollection.java
+++ b/src/main/java/com/stripe/model/StripeCollection.java
@@ -1,7 +1,7 @@
 package com.stripe.model;
 
 import com.stripe.Stripe;
-import com.stripe.exception.AuthenticationException;
+import com.stripe.exception.ApiKeyMissingException;
 import com.stripe.net.RequestOptions;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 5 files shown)


kamil-stripe (approved):
LGTM. @dcr-stripe double check java for me?

dcr-stripe (changes requested):
Hmm.... I don't think `AuthenticationException` is the right call here because it extends `StripeException` which extends `Exception`, and this is a checked exception and always needs to be handled. This would be a substantial breaking change because it means any use of these methods would now need to explicitly either:

```
try {

} catch (AuthenticationException e) {...}
```

or throw it up to a higher level using `throws`.

I think making a new exception type here is appropriate and having it extend `RuntimeException` so that it is unchecked. This is similar to https://github.com/stripe/stripe-java/blob/34a2a978/src/main/java/com/stripe/net/RequestOptions.java#L375

Also ideally let's add a test to this behavior to make sure we don't regress in the future.

src/main/java/com/stripe/exception/ApiKeyMissingException.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * {@link ApiKeyMissingException} is thrown when the API key is not set for a request. The API key
+ * for a request may be set either globally through {@link com.stripe.Stripe#apiKey} or through
+ * {@link com.stripe.net.RequestOptions}.
+ */
+public class ApiKeyMissingException extends RuntimeException {
+  private static final long serialVersionUID = 1L;
+
+  /** Initializes a new instance of the {@link ApiKeyMissingException}. */
```
dcr-stripe, line 11: Do we need this constructor? Maybe we should always pass a message explaining the fix?

src/main/java/com/stripe/model/StripeCollection.java
```diff
@@ -74,6 +83,10 @@ public Iterable<T> autoPagingIterable(Map<String, Object> params) {
    * @param options request options (will override the options from the initial list request)
    */
   public Iterable<T> autoPagingIterable(Map<String, Object> params, RequestOptions options) {
+    String apiKey = options.getApiKey();
+    if (apiKey == null) {
```
dcr-stripe, line 87: Should this also check the global `Stripe.apiKey`?

src/main/java/com/stripe/model/StripeSearchResult.java
```diff
@@ -60,6 +68,10 @@ public Iterable<T> autoPagingIterable(Map<String, Object> params) {
    * @param options request options (will override the options from the initial list request)
    */
   public Iterable<T> autoPagingIterable(Map<String, Object> params, RequestOptions options) {
+    String apiKey = options.getApiKey();
```
dcr-stripe, line 71: Same comment here.

dcr-stripe (commented):
LG overall, just some small ergonomics comments!

r? @yejia-stripe

src/main/java/com/stripe/model/StripeCollection.java
yejia-stripe: I think if `RequestOptionsBuilder` (which it should?) is used, then `options.getApiKey()` defaults to `Stripe.apiKey`

src/main/java/com/stripe/exception/ApiKeyMissingException.java
yejia-stripe: Do you mean in the constructer, do something like `super('API key missing')`, or do you mean when we throw the exception?

src/main/java/com/stripe/model/StripeCollection.java
dcr-stripe: Good call - that sounds good to me!

src/main/java/com/stripe/exception/ApiKeyMissingException.java
dcr-stripe: We shouldn't provide an empty constructor, and the call sites which new up an exception should pass an error message explaining the issue + fix.

src/main/java/com/stripe/model/StripeCollection.java
```diff
@@ -57,10 +59,19 @@
   private Map<String, Object> requestParams;
 
   public Iterable<T> autoPagingIterable() {
+    if (Stripe.apiKey == null) {
+      throw new ApiKeyMissingException(
+          "API key is not set for autoPagingIterable. You can set the API key globally using Stripe.ApiKey, or through RequestOptions in with autoPagingIterable(params, options).");
```
dcr-stripe, line 64: "in with" -> "with" throughout

dcr-stripe (approved):
Thanks Yejia, this looks great! Just one comment and then good to go!

PR #1403, opened 2022-07-22
Delete deprecated files.

dcr-stripe:
r? @kamil-stripe 

## Changelog
* Delete the `CashBalance.retrieve` instance method and rename the `CashBalance.retrieveCashBalance` static method to `CashBalance.retrieve`.
  * Previously if you had:
     ```
     CashBalance.retrieveCashBalance(customerId)
     ```
     this would become:
     ```
     CashBalance.retrieve(customerId)
     ```
* Delete `treasury.InitiatingPaymentMethodDetails` and `treasury.BillingDetails` top-level classes. These were unused and each treasury class has its own version of it.
* Delete `IssuerFraudRecord` resource. This resource is obsolete and has been superseded by `EarlyFraudWarning`.


src/main/java/com/stripe/model/CashBalance.java
```diff
@@ -49,54 +48,19 @@ public class CashBalance extends ApiResource {
   BalanceSettings settings;
 
   /** Retrieves a customer’s cash balance. */
-  public static CashBalance retrieveCashBalance(String customer) throws StripeException {
```
kamil-stripe, line 52: This seems to not be a straight-forward removal

kamil-stripe:
> Delete CashBalance.retrieve instance method. The CashBalance.**retrieve** static method is still available and should be used instead.

In the changelog.

src/main/java/com/stripe/model/CashBalance.java
kamil-stripe: I see the description. Please update the PR's title :)

kamil-stripe (approved):

src/main/java/com/stripe/model/CashBalance.java
dcr-stripe: Good catch! Updated.

PR #1424, opened 2022-08-19
Allow setting Stripe API Version on global configuration.

dcr-stripe:
r? @pakrym-stripe 

## Summary

Allows configuring `Stripe.stripeVersion` on a global-basis to specify beta headers.

For per-requests, `setStripeVersionOverride` can be used.

This also tweaks webhook version checking to ignore beta headers.

## Note 

I tried to stick to the nomenclature used within stripe-java + pre-existing methods rather than making breaking changes to name this apiVersion across the board. WDYT?

### DIFF
--- a/src/main/java/com/stripe/model/EventDataObjectDeserializer.java
+++ b/src/main/java/com/stripe/model/EventDataObjectDeserializer.java
@@ -155,10 +155,7 @@ public class EventDataObjectDeserializer {
                     + "consider transforming the raw JSON data object to be compatible with this "
                     + "current model class schemas using `deserializeUnsafeWith`. "
                     + "Original error message: %s",
-                Stripe.stripeVersion,
-                this.apiVersion,
-                Stripe.stripeVersion,
-                e.getMessage());
+                Stripe.stripeVersion, this.apiVersion, Stripe.stripeVersion, e.getMessage());
       } else {
         errorMessage =
             String.format(
--- a/src/main/java/com/stripe/model/EventDataObjectDeserializer.java
+++ b/src/main/java/com/stripe/model/EventDataObjectDeserializer.java
@@ -183,7 +183,6 @@ public class EventDataObjectDeserializer {
         transformer.transform(rawJsonObject.deepCopy(), apiVersion, eventType));
   }
 
-  @SuppressWarnings("StringSplitter")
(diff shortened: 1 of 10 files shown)


src/main/java/com/stripe/model/EventDataObjectDeserializer.java
```diff
@@ -185,13 +182,14 @@ public StripeObject deserializeUnsafeWith(CompatibilityTransformer transformer)
...
   }
 
+  @SuppressWarnings("StringSplitter")
   private boolean apiVersionMatch() {
-    return getIntegrationApiVersion().equals(this.apiVersion);
-  }
+    // Test that only the actual API version component of the version strings match. The integration
+    // API version might have beta headers that are not present in the event payload.
+    String integrationApiVersionStableComponent = Stripe.stripeVersion.split(";")[0];
+    String payloadApiVersionStableComponent = this.apiVersion.split(";")[0];
```
pakrym-stripe, line 190: Is this required?  I haven't done it in my PRs.

src/main/java/com/stripe/model/EventDataObjectDeserializer.java
```diff
@@ -185,13 +182,14 @@ public StripeObject deserializeUnsafeWith(CompatibilityTransformer transformer)
         transformer.transform(rawJsonObject.deepCopy(), apiVersion, eventType));
   }
 
+  @SuppressWarnings("StringSplitter")
```
pakrym-stripe, line 185: Should we use substring and avoid the warning?

pakrym-stripe (approved):

src/main/java/com/stripe/model/EventDataObjectDeserializer.java
dcr-stripe: Good call. Done.

src/main/java/com/stripe/model/EventDataObjectDeserializer.java
dcr-stripe: You're right, I don't think so. Updated.

dcr-stripe (commented):
ptal @pakrym-stripe , updated!

pakrym-stripe (approved):

PR #1429, opened 2022-08-24
API Updates for beta branch

dcr-stripe:
Codegen for openapi v185.
r? @pakrym-stripe
cc @stripe/api-libraries

## Changelog
- Updated stable APIs to the latest version


src/main/java/com/stripe/model/giftcards/Transaction.java
```diff
@@ -0,0 +1,312 @@
...
+  @SerializedName("confirmed_at")
+  Long confirmedAt;
+
+  /** Time at which the object was created. Measured in seconds since the Unix epoch. */
+  @SerializedName("created")
+  Long created;
+
+  /** The related Stripe objects that created this gift card transaction. */
+  @SerializedName("created_by")
+  Card.CreatedBy createdBy;
```
dcr-stripe, line 51: This needs unsharing

src/main/java/com/stripe/model/Order.java
```diff
@@ -82,6 +85,10 @@ public class Order extends ApiResource implements HasId, MetadataStore<Order> {
   @SerializedName("created")
   Long created;
 
+  /** The credits applied to the Order. At most 10 credits can be applied to an Order. */
+  @SerializedName("credits")
+  List<Order.Credit> credits;
```
kamil-stripe, line 90: Should that be just `Credit`? Not sure if there is any difference.

kamil-stripe (approved):

src/main/java/com/stripe/model/Order.java
dcr-stripe: Not sure why it generated the fully qualified version, but `Order.Credit` is fine here! Good question.

pakrym-stripe (approved):

PR #1431, opened 2022-08-26
Update coveralls to only run for one java version

anniel-stripe:
### Summary
r? @dcr-stripe 

Update coveralls CI step to run for one Java version like in other client libraries.

See related discussion thread: https://github.com/stripe/stripe-php/pull/1360#discussion_r955290668


dcr-stripe (approved):
LGTM!

PR #1434, opened 2022-09-06
API Updates

yejia-stripe:
Codegen for openapi v189.
r? @pakrym-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new value `terminal_reader_splashscreen` on enum `FileListParams.purpose`
* Fix `LineItem.Discount.discount` to reference the correct `Discount` class (`com.stripe.model.Discount`)


pakrym-stripe (approved):

davidbigham:
The change log mentions "Add support for new value terminal_reader_splashscreen on enum FileListParams.purpose", but terminal_reader_splashscreen has not been added to FileListParams.Purpose enum (only a mention in Javadoc comment)

anniel-stripe:
Hi @davidbigham ! In the stripe-java library, we represent enums on the response side as strings, so this Javadoc-only change is expected (I can see why the Changelog entry is confusing, though).

davidbigham:
Hi @anniel-stripe 
Sure you use Strings - but there is already an enum in a builder that has all (?) other types in it:
`com.stripe.param.FileCreateParams.Purpose` ~ line 209

Seems all that is missing is a line something like:
```java
    @SerializedName("terminal_reader_splashscreen")
    TERMINAL_READER_SPLASHSCREEN("terminal_reader_splashscreen");
```

I would prefer to use builders when possible, rather than Maps, so instead of:

```java
final FileCreateParams fileCreateParams = FileCreateParams.builder()
                 .setPurpose(FileCreateParams.Purpose.TERMINAL_READER_SPLASHSCREEN) // TODO API is missing "terminal_reader_splashscreen" Purpose (as at version 22.13.0)
                 .setFile(splashFileInputStream)
                 .build();
```

I need to do the following (the above builder does not have the `addExtraParam` method mentioned on the GitHub site https://github.com/stripe/stripe-java#how-to-use-undocumented-parameters-and-properties):

```java
final Map<String, Object> fileCreateParams = new HashMap<>();
         fileCreateParams.put("purpose", "terminal_reader_splashscreen");
         fileCreateParams.put("file", splashFileInputStream);
```

I guess it doesn't matter - there are work arounds - but still, it would be nice to use the supplied builders.

Cheers,
David

davidbigham:
Thanks, will open an issue future. Cheers, David

PR #1446, opened 2022-09-21
React to EventDataClassLookup generation changes

pakrym-stripe:
Remove the generated method that can easily be replaced in hand-written code.


anniel-stripe (approved):
LGTM!

PR #1457, opened 2022-10-13
Bugfix: Don't require global API Key when API Key is propagated from list request

richardm-stripe:


src/main/java/com/stripe/model/StripeCollection.java
```diff
@@ -59,15 +59,17 @@
   private Map<String, Object> requestParams;
 
   public Iterable<T> autoPagingIterable() {
-    if (Stripe.apiKey == null) {
+    System.out.println("foo");
```
pakrym-stripe, line 62: rem

src/test/java/com/stripe/model/PagingIteratorTest.java
```diff
@@ -44,10 +46,26 @@ public String getId() {
...
+      return request(
+          ApiResource.RequestMethod.GET,
+          classUrl(ReferencesPageableModel.class),
+          new HashMap<String, Object>(),
+          ReferencesPageableModel.class,
+          options);
+    }
+  }
+
+  List<String> pages;
```
pakrym-stripe, line 64: Is this new field used outside of setUpMockPages ?

src/test/java/com/stripe/model/PagingIteratorTest.java
```diff
@@ -77,6 +95,56 @@ public PageableModelCollection answer(InvocationOnMock invocation) {
...
+        .when(networkSpy)
+        .request(
+            Mockito.any(ApiResource.RequestMethod.class),
+            Mockito.anyString(),
+            Mockito.<Map<String, Object>>any(),
+            Mockito.<Class<ReferencesPageableModel>>any(),
+            Mockito.<RequestOptions>any());
+    Stripe.apiKey = null;
+    ReferencesPageableModel model =
+        ReferencesPageableModel.retrieve(RequestOptions.builder().setApiKey("sk_test_xyz").build());
```
pakrym-stripe, line 115: Should we assert that RequestOptions flow/don't flow?

src/test/java/com/stripe/model/PagingIteratorTest.java
```diff
@@ -168,4 +236,28 @@ void testAutoPaginationWithoutApiKey() throws StripeException {
           collection.autoPagingIterable();
         });
   }
+
+  @Test
+  void testAutoPaginationRequestOptionsPropagation() throws StripeException {
```
pakrym-stripe, line 241:
```suggestion
  public void testAutoPaginationRequestOptionsPropagation() throws StripeException {
```

src/test/java/com/stripe/model/PagingIteratorTest.java
```diff
@@ -168,4 +236,28 @@ void testAutoPaginationWithoutApiKey() throws StripeException {
...
+    // set some arbitrary parameters so that we can verify that they're
+    // used for requests on ALL pages
+    final Map<String, Object> page0Params = new HashMap<>();
+    page0Params.put("foo", "bar");
+
+    Stripe.apiKey = null;
```
pakrym-stripe, line 247: nit: use the preserve/restore pattern for global settings https://github.com/stripe/stripe-java/blob/110ae567c68aa67cfe697a6309d35e3d13c5e419/src/test/java/com/stripe/net/StripeRequestTest.java#L125-L140

src/test/java/com/stripe/model/PagingIteratorTest.java
```diff
@@ -168,4 +236,28 @@ void testAutoPaginationWithoutApiKey() throws StripeException {
           collection.autoPagingIterable();
         });
   }
+
+  @Test
+  void testAutoPaginationRequestOptionsPropagation() throws StripeException {
+    // set some arbitrary parameters so that we can verify that they're
+    // used for requests on ALL pages
+    final Map<String, Object> page0Params = new HashMap<>();
+    page0Params.put("foo", "bar");
```
pakrym-stripe, line 245: Assert that the param is sent with all pages?

src/test/java/com/stripe/model/PagingIteratorTest.java
richardm-stripe: Ah good catch.

src/test/java/com/stripe/model/PagingIteratorTest.java
richardm-stripe: This is already handled in the parent class:
https://github.com/stripe/stripe-java/blob/ba64a45bccbdafae779c841e64802941f6243e22/src/test/java/com/stripe/BaseStripeTest.java#L136

src/test/java/com/stripe/model/PagingIteratorTest.java
richardm-stripe: I think a couple lines down we assert that they DON'T flow in this case, but I added a comment to make this more explicit, and I also added the assertion to the other test case where they DO flow.

src/test/java/com/stripe/model/PagingIteratorTest.java
richardm-stripe: The comment was copy pasta, that doesn't make sense with this test case imo.

PR #1464, opened 2022-10-25
Ephemeral key accepts Stripe-Version through params, not opts (next major)

richardm-stripe:
## Notify
cc @stripe/api-library-reviewers 

## Summary

This turns `RequestOptions.setStripeVersionOverride` into something completely internal, something that users should never use. There was only one supported use case for `setStripeVersionOverride`, for `EphemeralKey.create`. Now, users should

```java
EphemeralKeyCreateParams params = EphemeralKeyCreateParams.builder()
  .setStripeVersion("XXXX-YY-ZZ")
  .build();
```

to set the `Stripe-Version` header appropriately.

Unfortunately, because `com.stripe.model` and `com.stripe.net` are different packages, we can't actually change the visibility of `RequestOptions.setStripeVersionOverride`. Instead, we rename it to confront upgrading users with the fact that it is now internal-only and they should not use it.

## Changelog
* Add support for `StripeVersion` as a required value on `EphemeralKeyCreateParams`.
* Add support for `EphemeralKey.create` without specifying `RequestOptions`.
* `RequestOptions.setStripeVersionOverride` is renamed to `_setStripeVersionOverride` to confront upgrading users with the fact that this is now internal-only and should not be used. The docs now indicate it is internal-only, instead of outlining circumstances it might be appropriate.
* `RequestOptions.getStripeVersionOverride` is similarly changed to `_getStripeVersionOverride`.
* `RequestOptions.clearStripeVersionOverride` is removed entirely. There is no reason to use this.
* `RequestOptions.toBuilder` is now marked as deprecated, in favor of `toBuilderFullCopy`, which has clearer semantics.

### DIFF
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -291,7 +291,7 @@ public class RequestOptions {
     }
 
     /** For internal use only. */
-    public String _getStripeVersionOverride() {
+   String getStripeVersionOverride() {
       return this.stripeVersionOverride;
     }
 
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -84,8 +84,8 @@ public class RequestOptions {
     return stripeVersion;
   }
 
-  public String getStripeVersionOverride() {
-    return stripeVersionOverride;
+  public static String unsafeGetStripeVersionOverride(RequestOptions options) {
+    return options.stripeVersionOverride;
   }
 
   public int getReadTimeout() {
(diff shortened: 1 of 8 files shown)


richardm-stripe:
Assigned you @pakrym-stripe as you are the "reporter" on our internal ticket for this ;)

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -267,31 +289,24 @@ public RequestOptionsBuilder clearStripeAccount() {
...
+     * RequestOptions)}. Setting this yourself will result in a version mismatch between your
+     * request and this library's types, which can result in missing data and deserialization
+     * errors.
      *
-     * @param stripeVersionOverride stripe version override which belongs to the client to make
-     *     request on behalf of.
      * @return option builder
      */
-    public RequestOptionsBuilder setStripeVersionOverride(String stripeVersionOverride) {
+    public RequestOptionsBuilder _setStripeVersionOverride(String stripeVersionOverride) {
```
pakrym-stripe, line 305: can we make this a static method so it doesn't show up in IDE code completion. And maybe name it `unsafeSetStripeVersionOverride`?

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -115,12 +115,34 @@ public static RequestOptionsBuilder builder() {
...
   public RequestOptionsBuilder toBuilder() {
     return new RequestOptionsBuilder().setApiKey(this.apiKey).setStripeAccount(this.stripeAccount);
   }
 
+  /**
+   * Convert request options to builder, copying all options.
+   *
+   * @return option builder.
+   */
+  public RequestOptionsBuilder toBuilderFullCopy() {
```
pakrym-stripe, line 132: Do we even need this one?

richardm-stripe: Yes, because `com.stripe.net` is not the same package as `com.stripe.model`

richardm-stripe: There might be a solution involving [JPMS](https://www.oracle.com/corporate/features/understanding-java-9-modules.html) but since we still support Java 8 I haven't done much research there.

pakrym-stripe: Can we inline the implementation into the place it's currently used at?

pakrym-stripe (approved):
LGTM overall. Wonder if we can hide _setStripeVersionOverride in a way that doesn't pollute code completion.

src/main/java/com/stripe/net/RequestOptions.java
richardm-stripe: No, because these fields are private

pakrym-stripe: Don't all of them have a getter?

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -267,29 +290,25 @@ public RequestOptionsBuilder clearStripeAccount() {
       return setStripeAccount(null);
     }
 
-    public String getStripeVersionOverride() {
+    /** For internal use only. */
+    public String _getStripeVersionOverride() {
```
pakrym-stripe, line 294:
```suggestion
   String getStripeVersionOverride() {
```

same package as test so doesn't need to be public

pakrym-stripe (approved):
One nit, let's hide the getVersionOverride

PR #1465, opened 2022-10-25
Create CONTRIBUTING.md

nob788:


richardm-stripe:
I think we should wait to add CONTRIBUTORS.md until we have content for it.

PR #1473, opened 2022-11-09
API Version 2022-11-09

richardm-stripe:
## Changelog
* Remove support for `charges` on `PaymentIntent` (breaking)
* Add support for `latest_charge` on `PaymentIntent`
* Add support for `reason_message` on `issuing.Authorization`
* Remove support for `amount`, `currency`, `description`, `images` and `name` from `line_items` on `checkout.SessionCreateParams` (breaking)
* Remove support for `PayNow.tos_shown_and_accepted` from `checkout.SessionCreateParams` (breaking)

### DIFF
--- a/src/main/java/com/stripe/ApiVersion.java
+++ b/src/main/java/com/stripe/ApiVersion.java
@@ -2,5 +2,5 @@
 package com.stripe;
 
 final class ApiVersion {
-  public static final String CURRENT = "2022-08-01";
+  public static final String CURRENT = "2022-11-09";
 }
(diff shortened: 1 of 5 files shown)


pakrym-stripe (approved):

cmboult:
Hi - I'm seeing some weird behaviour with the latest_charge field
we receive a webhook for a payment_intent.succeeded event and do a lookup using PaymentIntent.retrieve but latest_charge is NULL. If I do the same call manually a few seconds later latest_charge is populated... is this expected behaviour?

Issue #1181, opened 2021-02-07
Regarding ability to override stripe api url per API request


devansh-dalal:
Hi

We are using [stripe-mock](https://github.com/stripe/stripe-mock) for our e2e tests requiring stripe API calls. There we required to change the stripe api base url per request as:
```
public PaymentMethod createPaymentMethod(Card card) {
      var paymentMethodParams =
          PaymentMethodCreateParams.builder()
              .setType(Type.CARD)
              .setCard(
                  CardDetails.builder()
                      .setNumber(card.getNumber())
                      .setExpMonth((long) card.getExpiryMonth())
                      .setExpYear((long) card.getExpiryYear())
                      .setCvc(card.getCvv())
                      .build())
              .build();
      var response = testMode? "Make API call to mock stripe with params" : PaymentMethod.create(paymentMethodParams);
      return response;
  }
```

This could be done globally via static method as 
```
Stripe.overrideApiBase();
```

but this changes the url of stripe api globally, It could have been usable if I could have overridden the api url per request, maybe via `RequestOptions`. Please do let if know if there are any problems, this should be easy, or not?

remi-stripe:
@devansh-dalal Thanks for the report! We'll track this internally but I think it will only happen once we ship support for the client/services infrastructure we're working on!

richardm-stripe:
This was made possible in https://github.com/stripe/stripe-java/pull/1476 and was released in https://github.com/stripe/stripe-java/releases/tag/v22.5.0.

PR #1476, opened 2022-11-14
Solution For "Regarding ability to override stripe api url per API re…

yasirshabbir44:
…quest #1181".

https://github.com/stripe/stripe-java/issues/1181


richardm-stripe:
Hello @yasirshabbir44. Thank you for this change. It makes sense to me!

Can you sign the CLA per CLAassistant's comment above?

yasirshabbir44:
Hello @richardm-stripe 
Done

richardm-stripe:
Ah, I see you have changed `src/main/java/com/stripe/model/PaymentMethod.java` to replace `Stripe.getApiBase()` with `options.getApiBase()` but

- This file is generated, so we have to change the generator (which alas, is private) otherwise the change to this file will be reverted.
- Many other models also need to be changed to use `options.getApiBase()`.
- We need to account for the possibility that `options` is null, since passing `null` RequestOptions is allowed into methods like `PaymentMethod.create`.

So we can't merge as-is. I'll see if I can get a generator change in this week.

yasirshabbir44:
Waiting for to get merge if you all get well please merge it.

thanks

richardm-stripe:
Hello @yasirshabbir44, I'm working on this and made a number of changes:

* I changed `apiBase` to `baseUrl`. I wanted to give it a different name from `Stripe.apiBase` because `Stripe.connectBase` (which defaults to `connect.stripe.com`) and `Stripe.uploadBase` (which defaulst to `files.stripe.com`) also exist, but specifying `RequestOptions.baseUrl` will override *all* of these, not just `Stripe.apiBase`.
* I changed it so that defaulting to Stripe.getApiBase() happens in a new method `ApiResource.fullUrl` which is called from each resource method. It needs to be this way, since we need to handle null `RequestOptions`, and methods that use `connectBase` or `uploadBase` need a way to specify that they have different defaults from other methods.
* I changed our generator so that all methods in the SDK respect `RequestOptions.baseUrl` and committed the resulting changes.

src/main/java/com/stripe/net/ApiResource.java
```diff
@@ -55,6 +55,14 @@ public ReflectionAccessFilter.FilterResult check(Class<?> rawClass) {
     return builder.create();
   }
 
+  public static String fullUrl(String defaultBaseUrl, RequestOptions options, String relativeUrl) {
```
pakrym-stripe, line 58:
```suggestion
  protected static String fullUrl(String defaultBaseUrl, RequestOptions options, String relativeUrl) {
```

richardm-stripe: Alas, StripeCollection is not a subclass of ApiResource

pakrym-stripe (approved):

yasirshabbir44:
Could you please merge it

richardm-stripe:
Sure, we can merge! We won't cut a release until the new year, though.

richardm-stripe:
Released in https://github.com/stripe/stripe-java/releases/tag/v22.5.0

yasirshabbir44:
Hello Richard,

I did changes and raised PR and that PR also got merged, but in the
contributor list I cannot see my name.

Why it’s not showing up?


Thanks
Yasir Shabbir

PR #1490, opened 2022-12-09
Fix flaky encoding test

pakrym-stripe:
I tried to fix it. But at thins point I'm out of options and the coverage seems to be enough without Java 10.


src/test/java/com/stripe/net/FormEncoderTest.java
```diff
@@ -395,7 +398,6 @@ class TestCase {
                     Collections.singletonMap("map", ImmutableMap.of("one", 1, "two", 2)),
                     "map[one]=1&map[two]=2"));
 
-            // --- URL-encoding of both keys and values ---
```
anniel-stripe, line 398: Was changing this comment intentional?

anniel-stripe (approved):

PR #1496, opened 2022-12-20
Replace ReflectionCheckingTypeAdapterFactory with a ReflectionAccessFilter

### LINKED ISSUE #1494: Library accesses GSON internal API, breaks with gson on the module path
### Describe the bug

The `ReflectionCheckingTypeAdapterFactory` class accesses a GSON internal class directly. This breaks when GSON is placed on the module path due to encapsulation.

`java.lang.IllegalAccessError: class com.stripe.net.ReflectionCheckingTypeAdapterFactory (in unnamed module @0x69a3d1d) cannot access class com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter (in module com.google.gson) because module com.google.gson does not export com.google.gson.internal.bind to unnamed module @0x69a3d1d @ com.stripe.net.ReflectionCheckingTypeAdapterFactory.create(Unknown Source)`

### To Reproduce

Place gson on the module path and then try to use the Stripe API.

### Expected behavior

There are a few possible fixes that spring to mind:

1. This class looks like a utility / assertion checking class, so it could be disabled in production code paths.
2. Instead of using `instanceof`, the class name could maybe be checked using reflection?
3. The `IllegalAccessError` could be caught, and that used to disable the check at runtime.

### Code snippets

_No response_

### OS

macOS

### Java version

Java 17

### stripe-java version

22.1.0

### API version

2022-11-15

### Additional context

A workaround is to place gson on the classpath instead of the module path. One could probably also use `--add-opens`.

richardm-stripe:
## Notify

## Summary
Fixes #1494
- Updates GSON by a patch version.
- Removes `ReflectionCheckingTypeAdapterFactory` -- this should be non breaking as the visibility is package-private.
- Adds a `ReflectionAccessFilter` (new in Gson 2.9.1) as a replacement. 

I have NOT actually set up a project to reproduce #1494, but I am still highly confident this will fix it as it is completely removing the class with the suspect import statement.

## Context 
Basically, we want to throw an exception when GSON is asked to use reflection and deserialize a class that isn't beneath `com.stripe.`, usually this happens when the class has an "infrastructure" field that doesn't represent data, but the field is not marked "transient", e.g. in [#1197](https://github.com/stripe/stripe-java/pull/1197/files#diff-513f6068c3bea2b70aeb2db8161819632c0aa9efccd400fa7852e44cd61675f3L53) the issue was that `StripeCollection` had a `private RequestOptions requestOptions` that should have been `private transient RequestOptions requestOptions`. ("transient" is a Java keyword that means "shouldn't be serialized"). The `ReflectionCheckingTypeAdapter` was added to give a better error message and help prevent us from making mistakes like this again (the error message in #1190 is pretty hard to interpret), but the `ReflectionAccessFilter` should do the same thing without causing the module path issues mentioned in #1494.


pakrym-stripe:
This is such an amazing find! There's barely a mention of `addReflectionAccessFilter` on the internet. Do we worry that this feature is too new and will break customers that have older GSON versions?

richardm-stripe:
All credit goes to [Marcono1234](https://github.com/Marcono1234) who commented on #1197. 

Stripe-java presently needs GSON >= 2.8.2 (when `JsonElement.deepCopy` [was added](https://github.com/google/gson/blob/master/CHANGELOG.md#version-282) [used here](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/EventDataDeserializer.java#L26)), from 2017.

By SemVer, 2.9.1 should be completely backwards compatible with 2.8.2 so in theory any user upgrading stripe-java should also be able to upgrade GSON without problems. 

This is five years of GSON versions we lose compatibility with, but I don't think we should default to assuming SemVer is untrustworthy and users won't be able to update. If we ever get any feedback about a versioning conflict, we can always release a newer version with a workaround.

anniel-stripe (approved):

PR #1501, opened 2023-01-03
Add fix for per-request ApiBase

richardm-stripe:
## Notify
r? @pakrym-stripe 
## Summary
Add tests for the feature introduced in #1476 and fixes two bugs (these bugs did not get released!)

* One bug broke `.toBuilderFullCopy()`. It didn't copy over the new field `baseUrl`.
* The other bug was that `new RequestOptions()` defaulted to `baseUrl = Stripe.getApiBase()` when really it should have defaulted to `null`. This means that code like in `Quote.java`

```java
  public InputStream pdf(Map<String, Object> params, RequestOptions options)
      throws StripeException {
    String url =
        ApiResource.fullUrl(
            Stripe.getUploadBase(),
            options,
            String.format("/v1/quotes/%s/pdf", ApiResource.urlEncodeId(this.getId())));
    return ApiResource.requestStream(ApiResource.RequestMethod.GET, url, params, options);
  }
```
would have regressed and been sending requests to api.stripe.com instead of files.stripe.com as intended.

### DIFF
--- a/src/main/java/com/stripe/model/File.java
+++ b/src/main/java/com/stripe/model/File.java
@@ -92,7 +92,7 @@ public class File extends ApiResource implements HasId {
    */
   public static File create(FileCreateParams params, RequestOptions options)
       throws StripeException {
-    checkNullTypedParams(String.format("%s/v1/files", Stripe.getUploadBase()), params);
+    checkNullTypedParams(ApiResource.fullUrl(Stripe.getUploadBase(), options, "/v1/files"), params);
     return create(params.toMap(), options);
   }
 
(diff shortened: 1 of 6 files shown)


pakrym-stripe (approved):

richardm-stripe:
r? @pakrym-stripe added another commit: https://github.com/stripe/stripe-java/pull/1501/commits/770bc13f46bb3de40d4b3fbe632168bab2cc2069

It looks like I missed replacing `String.format` with `Stripe.fullUrl` in a couple non-codegenned methods.

I'm now confident I got them all. I did
`grep -r Stripe.getApiBase -B1` (same for uploadBase and connectBase) to get all the places this is called. Then I did some vim magic to join any line that ended with "Stripe.fullUrl(\n" with the next line, then I deleted any line that contained "Stripe.fullUrl" then I looked at all the remaining instances of Stripe.apiBase and they were all in tests or error messages.

pakrym-stripe (approved):

PR #1553, opened 2023-04-25
Introduce Stripe.rawRequest as a canonical way to request APIs without definitions

anniel-stripe:


README.md
```diff
@@ -1,4 +1,4 @@
-# Stripe Java client library
+ # Stripe Java client library
```
anniel-stripe, line 1:
```suggestion
# Stripe Java client library
```

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -71,7 +71,55 @@ public StripeRequest(
...
+   * @param content the body of the request
+   * @param options the special modifiers of the request
+   * @throws StripeException if the request cannot be initialized for any reason
+   */
+  public StripeRequest(
+      ApiResource.RequestMethod method,
+      String url,
+      String content,
+      RequestOptions options)
+      throws StripeException {
```
anniel-stripe, line 109: This overload might be a breaking change. If for some reason a user were importing `StripeRequest` and passing in `null` for params, then they'd get an ambiguous method error

anniel-stripe: Should we instead do an overload that just accepts an additional string argument `content`? And call that with a `null` params for `rawRequest`?

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -71,7 +71,52 @@ public StripeRequest(
...
+      this.url = buildURL(method, url, params);
+
+      if (this.options instanceof RawRequestOptions) {
+        RawRequestOptions rawOptions = (RawRequestOptions) this.options;
+        this.content = buildContent(method, content, rawOptions.getApiMode());
+      } else {
+        this.content = buildContent(method, content, RawRequestOptions.ApiMode.STANDARD);
+      }
+
       this.headers = buildHeaders(method, this.options);
```
anniel-stripe, line 120: We should also generate the preview API version + set `Stripe-Version` header

src/main/java/com/stripe/net/StripeRequest.java
pakrym-stripe: we can do a static factory function like in .net https://github.com/stripe/stripe-dotnet/pull/2690/files#diff-ac904c400fdd4a5fb6b51487f1d8b4f8b5cc1076c67c44b590c9b0ce53a5ecc3R116

src/main/java/com/stripe/net/StripeResponseStream.java
```diff
@@ -22,7 +22,7 @@ public StripeResponseStream(int code, HttpHeaders headers, InputStream body) {
    *
    * @return the StripeResponse
    */
-  StripeResponse unstream() throws IOException {
+  public StripeResponse unstream() throws IOException {
```
pakrym-stripe, line 25: why?

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -71,7 +71,52 @@ public StripeRequest(
...
+
+      if (this.options instanceof RawRequestOptions) {
+        RawRequestOptions rawOptions = (RawRequestOptions) this.options;
+        this.content = buildContent(method, params, rawOptions.getApiMode());
+      } else {
+        this.content = buildContent(method, params, RawRequestOptions.ApiMode.STANDARD);
+      }
+
+      this.headers = buildHeaders(method, this.options);
+    } catch (IOException e) {
```
pakrym-stripe, line 83: unrelated but how can this exception even happen in the ctor?

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -71,7 +71,52 @@ public StripeRequest(
...
+  /**
+   * Initializes a new instance of the {@link StripeRequest} class.
+   *
+   * @param method the HTTP method
+   * @param url the URL of the request
+   * @param content the body of the request
+   * @param options the special modifiers of the request
+   * @throws StripeException if the request cannot be initialized for any reason
+   */
+  public StripeRequest(
```
pakrym-stripe, line 104: Should we create a protected ctor that takes `HttpContent` for content and have public ctors call it? That way we can avoid duplicating all the other logic.

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -129,14 +174,35 @@ private static URL buildURL(
...
+    }
+
     return FormEncoder.createHttpContent(params);
   }
 
+  private static HttpContent buildContent(
+      ApiResource.RequestMethod method, String content, RawRequestOptions.ApiMode apiMode)
+      throws IOException {
+    if (method != ApiResource.RequestMethod.POST) {
+      return null;
```
pakrym-stripe, line 196: Throw?

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -129,14 +174,35 @@ private static URL buildURL(
...
+    }
+
+    if (apiMode == RawRequestOptions.ApiMode.PREVIEW) {
+      return HttpContent.buildJsonContent(content);
```
pakrym-stripe, line 200: Is this correct? Will it not encode content as JSON string?

pakrym-stripe: Ah, we have an overload

src/main/java/com/stripe/net/StripeResponseStream.java
anniel-stripe: Because we need to change the visibility to use it here in rawRequest:
https://github.com/stripe/stripe-java/pull/1553/files#diff-8e4bf2c082d9f6d0e50bd100e2c612728dbcf75d230b1f4d33bd5adfcec4ef8aR265

src/main/java/com/stripe/net/StripeRequest.java
anniel-stripe: `buildURL` and `buildContent` throw `IOException`'s

src/main/java/com/stripe/net/StripeRequest.java
anniel-stripe: Oh yes I like that idea a lot

README.md
```diff
@@ -253,12 +253,38 @@ We highly recommend keeping an eye on when the beta feature you are interested i
...
+// Make the request using the Stripe.rawRequest() method.
+final StripeResponse response =
+  Stripe.rawRequest(
+    ApiResource.RequestMethod.POST, "/v1/beta_endpoint", "param=123", options);
+
+// (Optional) response.body() is a string. You can call
+// Stripe.deserialize() to get a StripeObject.
+StripeObject obj = Stripe.deserialize(response.body());
+
+// Handle response...
```
anniel-stripe, line 285:
```suggestion
```

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -51,6 +52,32 @@ public class StripeRequest {
...
+      HttpContent content,
+      Map<String, Object> params,
+      RequestOptions options)
+      throws StripeException {
+    this.content = content;
+    this.params = (params != null) ? Collections.unmodifiableMap(params) : null;
+    this.options = (options != null) ? options : RequestOptions.getDefault();
+    this.method = method;
+    this.url = buildURL(method, url, params);
+
```
anniel-stripe, line 77:
```suggestion
```

src/test/java/com/stripe/net/RawRequestOptionsTest.java
```diff
@@ -0,0 +1,158 @@
...
+        "{\"end_behavior\":\"release\",\"phases\":[{\"items\":[{\"quantity\":1,\"price\":\"price_123\"}],\"iterations\":12}],\"customer\":\"cus_123\",\"start_date\":1683338558}",
+        request.getBody().readUtf8());
+
+    assertNotNull(response);
+    assertEquals(200, response.code());
+    assertTrue(response.body().length() > 0);
+  }
+
+  @Test
+  public void testPreviewRequestWithComplexParams() throws StripeException, InterruptedException {
```
pakrym-stripe, line 101: Is this test different from `testPreviewRequest` ?

src/test/java/com/stripe/net/RawRequestOptionsTest.java
```diff
@@ -0,0 +1,158 @@
...
+    final StripeResponse response =
+        Stripe.rawRequest(
+            ApiResource.RequestMethod.POST,
+            "/v1/customers",
+            "{\"description\":\"test customer\"}",
+            options);
+
+    RecordedRequest request = server.takeRequest();
+    assertEquals("application/json", request.getHeader("Content-Type"));
+    assertNotEquals(Stripe.API_VERSION, request.getHeader("Stripe-Version"));
```
pakrym-stripe, line 121: SHouldn't this be preview?

pakrym-stripe, line 121: Let's check equality instead.

src/test/java/com/stripe/net/RawRequestOptionsTest.java
```diff
@@ -0,0 +1,158 @@
...
+    assertEquals(200, response.code());
+    assertTrue(response.body().length() > 0);
+
+    RecordedRequest request = server.takeRequest();
+    assertEquals(
+        "application/x-www-form-urlencoded;charset=UTF-8", request.getHeader("Content-Type"));
+    assertEquals(Stripe.API_VERSION, request.getHeader("Stripe-Version"));
+    assertEquals("description=test+customer", request.getBody().readUtf8());
+
+    Customer customer = (Customer) Stripe.deserialize(response.body());
```
pakrym-stripe, line 65: Should we split deserialize into it's own test?

src/main/java/com/stripe/Stripe.java
```diff
@@ -206,4 +212,82 @@ public static void setAppInfo(String name, String version, String url, String pa
...
+        String value = entry.getValue();
+        request = request.withAdditionalHeader(key, value);
+      }
+    }
+
+    if (options.getApiMode() == ApiMode.PREVIEW) {
+      request = request.withAdditionalHeader("Stripe-Version", ApiVersion.PREVIEW_CURRENT);
+    }
+
+    StripeResponseStream responseStream = new HttpURLConnectionClient().requestStream(request);
```
pakrym-stripe, line 267: Just noticed this. In all other cases of request sending we go through the `StripeResponseGetter`. It allows users to replace the HttpClient used and change it's settings. By going directly through the `HttpURLConnectionClient` here we are creating a separate way of making requests.

pakrym-stripe (changes requested):
Let's figure of if the request should go through the response getter abstraction.

src/main/java/com/stripe/Stripe.java
anniel-stripe: Very good point, just fixed.

src/main/java/com/stripe/Stripe.java
```diff
@@ -10,6 +14,7 @@ public abstract class Stripe {
   public static final int DEFAULT_READ_TIMEOUT = 80 * 1000;
 
   public static final String API_VERSION = ApiVersion.CURRENT;
+  public static final String PREVIEW_API_VERSION = ApiVersion.PREVIEW_CURRENT;
```
anniel-stripe, line 17: I'm a little wary about setting this as a public property on `Stripe`, but otherwise `PREVIEW_CURRENT` isn't accessible from `LiveStripeResponseGetter` and tests in the `net` package. Plus this is in the beta branch anyways, where users can expect volatility.

src/main/java/com/stripe/net/ApiResource.java
```diff
@@ -211,6 +211,12 @@ public static InputStream requestStream(
     return ApiResource.stripeResponseGetter.requestStream(method, url, params, options);
   }
 
+  public static StripeResponse rawRequestStream(
```
pakrym-stripe, line 214: Do we need the stream version of raw request?

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -104,37 +142,88 @@ public StripeRequest withAdditionalHeader(String name, String value) {
...
+      throws ApiConnectionException {
     if (method != ApiResource.RequestMethod.POST) {
       return null;
     }
 
-    return FormEncoder.createHttpContent(params);
+    HttpContent httpContent = null;
+    try {
+      if (apiMode == RawRequestOptions.ApiMode.PREVIEW) {
+        httpContent = JsonEncoder.createHttpContent(params);
```
pakrym-stripe, line 187: Is this ever used considering we take the string body?

src/main/java/com/stripe/net/JsonEncoder.java
```diff
@@ -0,0 +1,18 @@
...
+import com.google.gson.FieldNamingPolicy;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import java.io.IOException;
+import java.util.Map;
+
+final class JsonEncoder {
+  private static final Gson BODY_GSON =
+      new GsonBuilder()
+          .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
```
pakrym-stripe, line 12: Do we ever use JSON serializer?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -109,6 +110,50 @@ public InputStream requestStream(
     return responseStream.body();
   }
 
+  @Override
+  public StripeResponse rawRequestStream(
```
pakrym-stripe, line 114: should this be a non-streaming overload instead?

src/main/java/com/stripe/model/StripeError.java
```diff
@@ -205,4 +205,14 @@ public class StripeError extends StripeObject {
    */
   @SerializedName("type")
   String type;
+
+  @SerializedName("developer_message")
+  String developerMessage;
+
+  public String getMessage() {
```
pakrym-stripe, line 212: Does this override the getter for `message` property?

We might want to handle this logic closer to Exception creation. StripeError is more of a data class so it should reflect what we receive from the service.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -109,6 +110,50 @@ public InputStream requestStream(
...
+    if (additionalHeaders != null) {
+      for (Map.Entry<String, String> entry : additionalHeaders.entrySet()) {
+        String key = entry.getKey();
+        String value = entry.getValue();
+        request = request.withAdditionalHeader(key, value);
+      }
+    }
+
+    if (options.getApiMode() == ApiMode.PREVIEW) {
+      request = request.withAdditionalHeader("Stripe-Version", Stripe.PREVIEW_API_VERSION);
```
pakrym-stripe, line 130: Where is this value set for non-preview requests?

src/main/java/com/stripe/net/StripeRequest.java
anniel-stripe: Ah good point, we don't need this switch in regular `buildContent`. I'll remove it

src/main/java/com/stripe/net/ApiResource.java
anniel-stripe: Ah you're right, we don't need the stream version. Updated

src/main/java/com/stripe/net/JsonEncoder.java
anniel-stripe: Nope :)

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
anniel-stripe: It's set in `buildHeaders` when the `StripeRequest` is constructed: https://github.com/stripe/stripe-java/blob/beta/src/main/java/com/stripe/net/StripeRequest.java#L187

pakrym-stripe: Any reason not to handle this header in StripeRequest? We already have some apiMode-specific logic there.

anniel-stripe: Good point, moved to StripeRequest

pakrym-stripe (approved):
Great work!

PR #1650, opened 2023-09-13
Flow response getters into Event.data.object and models deserialized via ApiResource.GSON

richardm-stripe:
This PR fixes multiple cases of StripeResponseGetter not being set on objects:
1. When objects are deserialized using `ApiResources.GSON`. In this case we add a TypeAdapter that calls `setResponseGetter` on every deserialized objects that supports it using the global response getter.
2. When events are deserialized using `constructEvent`. To fix this we centralized StripeObject deserialization logic in `StripeObject.deserializeStripeObject` and require passing a response getter.


pakrym-stripe (approved):

richardm-stripe (commented):
LGTM

PR #1663, opened 2023-09-29
Flow type information further into autopagination requests

### LINKED ISSUE #1660: StripeClient#subscriptions().list() causes class com.google.gson.internal.LinkedTreeMap cannot be cast to class com.stripe.model.HasId
### Describe the bug

The following code throws an exception.
```java
for (final Subscription subscription : stripeClient.subscriptions().list(params).autoPagingIterable()) {
      System.out.println("subscription: " + subscription);
}
```

The exception is:
```
Exception in thread "main" java.lang.RuntimeException: Unable to lazy-load stripe objects
	at com.stripe.model.PagingIterator.next(PagingIterator.java:57)
	at com.stripe.model.PagingIterator.next(PagingIterator.java:9)
	at SubscriptionsFailJavaMain.main(SubscriptionsFailJavaMain.java:32)
Caused by: java.lang.ClassCastException: class com.google.gson.internal.LinkedTreeMap cannot be cast to class com.stripe.model.HasId (com.google.gson.internal.LinkedTreeMap and com.stripe.model.HasId are in unnamed module of loader 'app')
	at com.stripe.model.StripeCollection.setResponseGetter(StripeCollection.java:92)
	at com.stripe.model.StripeObject.deserializeStripeObject(StripeObject.java:131)
	at com.stripe.net.LiveStripeResponseGetter.request(LiveStripeResponseGetter.java:72)
	at com.stripe.model.PagingIterator.list(PagingIterator.java:79)
	at com.stripe.model.PagingIterator.next(PagingIterator.java:53)
	... 2 more
```

### To Reproduce

Here is a whole code to reproduce.
```java
// The public API key
final String API_KEY = "sk_test_<redacted>";
final RequestOptions requestOptions = RequestOptions.builder().setApiKey(API_KEY).build();

final StripeClient stripeClient = new StripeClient(API_KEY);

final SubscriptionListParams params = SubscriptionListParams.builder()
    .setLimit(100L)
    .setCreated(
    SubscriptionListParams.Created.builder()
        .setGt(Instant.now().getEpochSecond() - 3 * 86400 /* 3 days */)
        .build()
    ).build();

// // Works!
// for (final Subscription subscription : Subscription.list(params, requestOptions).autoPagingIterable()) {
//   System.out.println("subscription: " + subscription);
// }

// Not works :(
for (final Subscription subscription : stripeClient.subscriptions().list(params).autoPagingIterable()) {
    System.out.println("subscription: " + subscription);
}
```

### Expected behavior

List subscriptions without errors.

### Code snippets

_No response_

### OS

macOS

### Java version

corretto-11.0.16.1

### stripe-java version

23.6.0

### API version

23.6.0

### Additional context

_No response_

richardm-stripe:
Fixes #1660.

## Explanation
Whereas `Subscription.list()` returns `SubscriptionCollection` (which implements `StripeCollection<Subscription>`), we decided that `stripeClient.subscriptions().list()` would return simply `StripeCollection<Subscription>`, in order to (in the future) prevent cluttering the library with mostly empty `XCollection` classes, and helping users need to import fewer classes.

Unfortunately, this change makes deserializing collections a little trickier. You can't just do `GSON.fromJson(..., collection.getClass())` because the class is just `StripeCollection` now, it doesn't tell you the type of what the inner objects should be. You need to do something like `new TypeToken<StripeCollection<Subscription>>() {}.getType()` to hint to GSON that the type of the items inside is `Subscription`, and then you need to take this `Type` and pipe it around everywhere where more requests to fetch a collection might be invoked.

We failed to do this for the requests made by `PagingIterator.list()`. When you do `new PagingIterator(collection, ...)` it calls `collection.getClass()` in order to get the deserialization key (which corresponds just to a generic `StripeCollection`), which ultimately causes the error in #1660. This PR flows through the TypeToken from the initial `SubscriptionService.list(...)` down onto the collection where it can be used by PagingIterator.

I also add a test.


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -77,6 +77,7 @@ public <T extends StripeObjectInterface> T request(
     if (resource instanceof StripeCollectionInterface<?>) {
       ((StripeCollectionInterface<?>) resource).setRequestOptions(options);
       ((StripeCollectionInterface<?>) resource).setRequestParams(params);
+      ((StripeCollectionInterface<?>) resource).setTypeToken(typeToken);
```
pakrym-stripe, line 80: Will this work for nested collections?

richardm-stripe: Not 100% clear what you mean. If you mean
```java
var stripe = StripeClient(...);
FooResource foo = stripe.foos().retrieve(...);
for (Bar bar: foo.getNestedCollectionOfBars().autoPagingIter()) {
  System.out.println(bar);
}
```
Then it won't work but is not needed atm because the way resources are generated right now, the type of `.getNestedCollectionOfBars()` is `BarCollection`, not `StripeCollection<Bar>`.

richardm-stripe: Ah I see the test failure, yeah there's something that needs to be done here.

richardm-stripe: Ok, I added  [Fall back to class](https://github.com/stripe/stripe-java/pull/1663/commits/518c39414ed04619c05840421775d8cd8133dfb6) -- so that  `foo.getBars().autoPaginatingIter()` will continue to work as before.

This seemed like the minimal option but there are others:
* Generate specialized collection types as we do for resources and give up on being generic.
* Do a lookup into `EventDataClassLookup` of `page[0].object` and do dynamic class lookup instead of static class lookup.

I think I prefer what's implemented but LMK if you have other thoughts.

pakrym-stripe: I was thinking about custom type adapter for xxxCollection and xxxSearchResult that will set this field for all objects that are deserialized.

richardm-stripe: Using what information? `page[0].object`?

pakrym-stripe: You have the type toke in the deserializer: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/BalanceTransactionSourceTypeAdapterFactory.java#L22

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -77,6 +77,7 @@ public <T extends StripeObjectInterface> T request(
     if (resource instanceof StripeCollectionInterface<?>) {
       ((StripeCollectionInterface<?>) resource).setRequestOptions(options);
       ((StripeCollectionInterface<?>) resource).setRequestParams(params);
+      ((StripeCollectionInterface<?>) resource).setPageTypeToken(typeToken);
```
pakrym-stripe, line 80: do we still need this?

richardm-stripe: We absolutely do not
```suggestion
```

src/main/java/com/stripe/model/PagingIterator.java
```diff
@@ -21,7 +21,7 @@ public class PagingIterator<T extends HasId> extends ApiResource implements Iter
       final StripeCollectionInterface<T> stripeCollection, StripeResponseGetter responseGetter) {
     this.url = stripeCollection.getUrl();
 
-    this.collectionType = stripeCollection.getClass();
+    this.pageType = stripeCollection.getPageTypeToken();
```
pakrym-stripe, line 24: can we pass this as a parameter instead of exposing getter on StripeCollectionInterface?

richardm-stripe: Done!

src/main/java/com/stripe/model/StripeCollection.java
```diff
@@ -58,16 +59,20 @@ public class StripeCollection<T extends HasId> extends StripeObject
   @Setter(onMethod = @__({@Override}))
   private Map<String, Object> requestParams;
 
+  @Setter(onMethod = @__({@Override}))
```
pakrym-stripe, line 62: remove getter?

src/main/java/com/stripe/model/StripeSearchResult.java
```diff
@@ -44,15 +45,19 @@ public class StripeSearchResult<T> extends StripeObject
   @Setter(onMethod = @__({@Override}))
   private Map<String, Object> requestParams;
 
+  @Setter(onMethod = @__({@Override}))
+  @Getter(onMethod = @__({@Override}))
```
pakrym-stripe, line 49: remove getter?

src/test/java/com/stripe/functional/StripeCollectionTest.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+public class StripeCollectionTest extends BaseStripeTest {
+  @Test
+  public void testPaginationWithStripeClient() throws StripeException {
```
pakrym-stripe, line 28: move to SearchPagingIteratorTest ?

src/test/java/com/stripe/functional/StripeCollectionTest.java
```diff
@@ -0,0 +1,62 @@
...
+    HttpClient spy = Mockito.spy(new HttpURLConnectionClient());
+    StripeResponseGetter srg = new LiveStripeResponseGetter(spy);
+    StripeClient client = new StripeClient(srg);
```
pakrym-stripe, line 31: mockClient?

src/test/java/com/stripe/functional/StripeSearchResultTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+public class StripeSearchResultTest extends BaseStripeTest {
+  @Test
+  public void testPaginationWithStripeClient() throws StripeException {
```
pakrym-stripe, line 29: Move to SearchPagingIteratorTest?

src/test/java/com/stripe/functional/StripeSearchResultTest.java
```diff
@@ -0,0 +1,68 @@
...
+    HttpClient spy = Mockito.spy(new HttpURLConnectionClient());
+    StripeResponseGetter srg = new LiveStripeResponseGetter(spy);
+    StripeClient client = new StripeClient(srg);
```
pakrym-stripe, line 32: Use BaseStripeTest.mockClient?

richardm-stripe: Done

src/test/java/com/stripe/functional/StripeSearchResultTest.java
richardm-stripe: Done

src/test/java/com/stripe/functional/StripeCollectionTest.java
richardm-stripe: Done!

src/main/java/com/stripe/model/StripeSearchResult.java
richardm-stripe: Done!

src/main/java/com/stripe/model/StripeCollection.java
richardm-stripe: Done!

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
richardm-stripe: Done!

pakrym-stripe (approved):

Issue #1664, opened 2023-10-03
The StripeResponseGetter has not been set on this resource


MustafaSmesem:
### Describe the bug

In stripe java 23.7.0 when call event.getDataObjectDeserializer() I got this error message:
The StripeResponseGetter has not been set on this resource. This should not happen and is likely a bug in the Stripe Java library. 

I have checked the event, it has data and deprecated method event.getData().getObject() is work and can get the object.

### To Reproduce

1. create webhook endpoint
2. listen to webhook event  
3. call event.getDataObjectDeserializer()  

### Expected behavior

I should get deserialized object

### Code snippets

```Java
@PostMapping("/stripe")
public ResponseEntity<ApiErrorResponse> stripeWebhookHandler(@RequestHeader("Stripe-Signature") String stripeSignature, @RequestBody Event event) {
            var object = event.getDataObjectDeserializer().getObject();
....
}
```

### OS

macOS

### Java version

Java 17

### stripe-java version

v23.7.0

### API version

 2023-08-16

anniel-stripe:
Hi @MustafaSmesem ! Thanks the bug report. Despite the error message, you can fix this by changing the type of the request body to a string in your handler's method signature:
```
@PostMapping("/stripe")
public ResponseEntity<ApiErrorResponse> stripeWebhookHandler(@RequestHeader("Stripe-Signature") String stripeSignature, @RequestBody **String payload**) {
        Event event = null;
        try {
          event = Webhook.constructEvent(payload, sigHeader, "{{WEBHOOK_SIGNING_SECRET}}");
        } catch (SignatureVerificationException e) {
          ...
        }
        var object = event.getDataObjectDeserializer().getObject();
....
}
```

yglodt:
Can anyone tell the most recent version not suffering from this bug ?

I know for sure 23.2.0, but I would preferably upgrade to the lastest non-affected version.

pakrym-stripe:
The 23.4.0 is the last version before this bug was introduced. We'll ship a version with a fix early next week.

pakrym-stripe:
The [v23.10.0](https://github.com/stripe/stripe-java/releases/tag/v23.10.0) has the fix for this issue.

jasperroel:
Hi @pakrym-stripe, I also noticed this happening in the newer `v24.0.0`  release. Should/can this fix also be ported/merged into the v24 release line?

pakrym-stripe:
[...]

PR #1671, opened 2023-10-13
Disallow falback to global response getter only in tests

pakrym-stripe:
There are many existing scenarios where fallback to global response getter makes sense. 

For example: 
```java
Customer c = new Customer();
c.setId("ca_123");
c.delete();
```

We would still like to validate that we flow response getter correctly in our tests. 

I've added a `stripe.disallowGlobalResponseGetterFallback` system property and set it to `true` for tests. When `stripe.disallowGlobalResponseGetterFallback` is set resources with null `responseGetter` won't fall back to global.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,6 +119,8 @@ test {
         events "passed", "skipped", "failed"
         exceptionFormat "full"
     }
+
+    systemProperty "stripe.disallowGlobalResponseGetterFallback", "true"
 }
 
 spotless {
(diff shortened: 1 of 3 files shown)


richardm-stripe (approved):

richardm-stripe:
So can users set the setting? What behavior will users see?

pakrym-stripe:
Users will get fallback to global. They can set the setting but I don't see why would they.

pakrym-stripe:
Contributes to https://github.com/stripe/stripe-java/issues/1664

PR #1698, opened 2023-11-30
Report usage of StripeClient

richardm-stripe:
## Changelog
* Reports use of the new `StripeClient` in `X-Stripe-Client-Telemetry`. (You can disable telemetry via `Stripe.enableTelemetry = false;`, see the [README](https://github.com/stripe/stripe-java/blob/master/README.md#telemetry).)

## Details
* Gives `LiveStripeResponseGetter` a mutable "usage" that it will set on all requests it creates.
* Causes the `StripeClient` constructor to set usage on its `LiveStripeResponseGetter` to `["stripe_client"]`.
* Modifies `RequestTelemetry` to accept usage and set it appropriately on `X-Stripe-Client-Telemetry`. Had to add a parameter to a public method, so I made a new overload and deprecated the old one. Had to disable a linter rule to do this.

## Testing
[src/test/java/com/stripe/functional/StripeClientTest.java](https://github.com/stripe/stripe-java/pull/1698/files#diff-946bf1dd11f5207280536926023d0e93fbf7e9f1cd9a3ffcfe7d067ec1a5fab0) 
and [src/test/java/com/stripe/functional/LiveStripeResponseGetterTest.java](https://github.com/stripe/stripe-java/pull/1698/files#diff-dee2b343dc42c4d0a31ff9938ad453ea687ad7136d01d82b0dd09c91a7e5dff7) mock at the HttpClient level.


build.gradle
```diff
@@ -50,6 +50,10 @@ tasks.withType(JavaCompile) {
       // com.stripe.param.AccountUpdateParams.Individual.Address] within this file.)
       // We should fix this by having autogen use the fully-qualified class to eliminate ambiguity.
       check("SameNameButDifferent", net.ltgt.gradle.errorprone.CheckSeverity.OFF)
+
+      // InlineMe (https://errorprone.info/docs/inlineme) seems neat, but in order to add these annotations
+      // we would be imposing another dependency on `errorprone` to our users, not worth it.
+      check("InlineMeSuggester", net.ltgt.gradle.errorprone.CheckSeverity.OFF)
```
anniel-stripe, line 56: Why was this necessary?

richardm-stripe: "errorprone" is our linter. We use it just in CI and don't impose errorprone as a dependency to our users. One of the linter errors in "errorprone" will take `@Deprecated` annotations that you add (part of standard Java), and if it determines that the implementation of the deprecated thing is simple enough to be inlined, it will suggest that you use an "@InlineMe" annotation instead (not part of standard Java, only in errorprone).

Makes sense if you're Google (the authors of errorprone) and you can impose a standard of errorprone being installed everywhere in your vast Java codebase, but if you're an open-source library that wants to make the library as dependency-free and easy to install as possible, the linter rule doesn't make sense.

src/main/java/com/stripe/StripeClient.java
```diff
@@ -18,18 +19,35 @@ public class StripeClient {
...
+      //
+      // Unfortunately, mutation could technically cause some non-obvious behavior in the case
+      // of re-use; e.g.
+      //
+      // var lsrg = new LiveStripeResponseGetter();
+      // ApiResource.setStripeResponseGetter(lsrg);
+      // new StripeClient(lsrg);
+      //
+      // would cause e.g. Customer.create(...) to erroneously report "stripe_client" in usage
+      ((LiveStripeResponseGetter) this.responseGetter).setUsage(Arrays.asList("stripe_client"));
```
anniel-stripe, line 49: This makes sense to me. The re-use case seems like a far edge case (where a user has to subclass `LiveStripeResponseGetter`, and is mixing paradigms).

richardm-stripe: Yeah the thing that bothers me is that in our test infra we actually did this... but I still think in real-life code its unlikely.

src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -121,6 +121,9 @@ public void setUpStripeMockUsage() {
     httpClientSpy = Mockito.spy(new HttpURLConnectionClient());
     networkSpy = Mockito.spy(new LiveStripeResponseGetter(null, httpClientSpy));
     mockClient = new StripeClient(networkSpy);
+    // The StripeClient constructor calls .setUsage on networkSpy, let's reset that so
+    // in the test cases we can just verify the calls we care about.
+    Mockito.reset(networkSpy);
```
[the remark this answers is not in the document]
richardm-stripe: My worry with that, is that we would then need to make copies of all the other helper functions here that use `networkSpy` and it might be a bad experience when you're writing a test to have to choose between them and select the right one, and I hesitated to make the experience worse in 95% of cases for an edge case that only affects usage-related tests.

I was on the fence though, let me know what you prefer.

PR #1699, opened 2023-11-30
compatibility checking

richardm-stripe:
Runs `japi-compliance-checker` in CI.

The new "compat" CI job will fail if you've introduced any breaking changes (it shouldn't be required to merge, but it is a nice data point to have).

Also, if you go to the Action "workflow", a `.zip` file with the compatibility report inside will be available at the bottom under "Artifacts". See the second to last commit on this branch.

image
image
image
image
image


.github/workflows/ci.yml
```diff
@@ -161,3 +161,63 @@ jobs:
...
+        run: ./gradlew jar
+
+      - name: Copy JAR to japi-compliance-checker
+        run: |
+          mv build/libs/*.jar ../japi-compliance-checker/old.jar
+
+      - name: Run JAPI Compliance Checker
+        run: |
+          cd ../japi-compliance-checker
+          (perl japi-compliance-checker.pl -old old.jar -new new.jar --lib=stripe-java || echo "failed" > compliance_failure) | tee report_stdout
```
richardm-stripe, line 208:
```suggestion
          perl japi-compliance-checker.pl -old old.jar -new new.jar --lib=stripe-java || echo "failed" > compliance_failure
```

anniel-stripe (approved):
This looks great, thanks for adding!!

PR #1702, opened 2023-12-04
Change StripeResponseGetter to take a single APIRequest object

richardm-stripe:
## Background/Approach
We keep wanting to add more arguments to
`StripeResponseGetter.request` and `StripeResponseGetter.requestStream`. We added `ApiMode` when we introduced [StripeClient](https://github.com/stripe/stripe-java/pull/1622/files#diff-55b4e268c6f5bded4a1a2dbe940b86c9e22f6ef986a54e6a053041b4284a1e10), and now we want to introduce "usage" (pending in https://github.com/stripe/stripe-java/pull/1698)

Rather than keep accumulating overloads and defining default implementations, we can just define a single class to represent the arguments that `.request` accepts. We'll still have to add new constructors (or setters or whatnot) when we add to this object, so it doesn't make the problem completely go away, but it's generally easier to backwards-compatibly extend an object with new constructors and new fields than it is to extend a method on an interface.

## Summary

- This PR introduces a new class `APIRequest` (please bikeshed this name), and a new overload to `StripeResponseGetter.request` and `StripeResponseGetter.requestStream` that accepts `APIRequest` instead of all the arguments it currently accepts. 
- This PR also changes all internal usages to use APIRequest.
- The new overload has a default implementation, so this change can be non-breaking.
- I did not add the @Deprecated annotation to the old overload. But I can.


src/main/java/com/stripe/model/Account.java
```diff
@@ -214,15 +210,15 @@ public CapabilityCollection capabilities(AccountCapabilitiesParams params, Reque
     String path =
         String.format("/v1/accounts/%s/capabilities", ApiResource.urlEncodeId(this.getId()));
     ApiResource.checkNullTypedParams(path, params);
-    return getResponseGetter()
-        .request(
+    APIRequest request =
```
pakrym-stripe, line 213: Api is cased differently in `ApiResource`

src/main/java/com/stripe/net/APIRequest.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Map;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+public class ApiRequest {
+  private BaseAddress baseAddress;
+  private ApiResource.RequestMethod method;
+  private String path;
+  private Map<String, Object> params;
```
pakrym-stripe, line 13: Params as a map won't work for RawRequest where params is a string.

richardm-stripe: Do you think we should make the caller convert to a string or just provide multiple constructors?

pakrym-stripe: I'm not sure but having callers converting params to string will go agains our goal of centralizing the logic in ApiRequestor.

richardm-stripe: Here is what I want to do re: RawRequest: https://github.com/stripe/stripe-java/pull/1717/commits/80833c9cc9e1c05dfbd5016ab03c3513f828fdf7#diff-2b360bd267e306d7c356360cb81b1acf0198ab73539a4a46ea5308824866bc57

I don't think there is much advantage as it stands to trying to represent the arguments to `ResponseGetter.rawRequest` and `ResponseGetter.request` with a single type -- separate types will do.

I was allured for a bit by the idea of letting users who want to implement StripeResponseGetter to provide only one method and not have to implement both `.rawRequest` and `.request`, but I don't think this is actually workable.

src/main/java/com/stripe/net/ApiRequest.java
```diff
@@ -0,0 +1,33 @@
...
+  private RequestOptions options;
+  private ApiMode apiMode;
+
+  @Setter private List<String> usage;
+
+  public ApiRequest(
+      BaseAddress baseAddress,
+      ApiResource.RequestMethod method,
+      String path,
+      Map<String, Object> params,
```
pakrym-stripe, line 23: Taking into account https://github.com/stripe/stripe-java/pull/1717, it doesn't feel right that the base type has `private Map<String, Object> params` while RawRequest adds a `
  @Getter private String rawContent;` as well. We get a type with two things that can become a body.

Should we define a base type without params and then have a `Map` and Raw subtypes?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -45,26 +45,24 @@ public LiveStripeResponseGetter(StripeResponseGetterOptions options, HttpClient
...
-      String path,
-      Map<String, Object> params,
-      Type typeToken,
-      RequestOptions options,
-      ApiMode apiMode)
+  public <T extends StripeObjectInterface> T request(ApiRequest apiRequest, Type typeToken)
       throws StripeException {
-    String fullUrl = fullUrl(baseAddress, options, path);
+    String fullUrl =
+        fullUrl(apiRequest.getBaseAddress(), apiRequest.getOptions(), apiRequest.getPath());
```
pakrym-stripe, line 51: Change fullUrl to take the apiRequest?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -45,26 +45,24 @@ public LiveStripeResponseGetter(StripeResponseGetterOptions options, HttpClient
...
     StripeRequest request =
-        new StripeRequest(method, fullUrl, params, RequestOptions.merge(this.options, options));
+        new StripeRequest(
```
pakrym-stripe, line 53: `StripeRequest` is an unfortunate name in the context of this change :/

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -115,12 +111,39 @@ public InputStream requestStream(
...
-      handleError(response, apiMode);
+      handleError(response, apiRequest.getApiMode());
     }
 
     return responseStream.body();
   }
 
+  @Override
+  @SuppressWarnings({"TypeParameterUnusedInFormals", "unchecked"})
+  public <T extends StripeObjectInterface> T request(
```
pakrym-stripe, line 122: Why isn't the default implementation sufficient?

src/main/java/com/stripe/net/StripeResponseGetter.java
```diff
@@ -18,6 +18,19 @@ <T extends StripeObjectInterface> T request(
       ApiMode apiMode)
       throws StripeException;
 
+  @SuppressWarnings("TypeParameterUnusedInFormals")
+  default <T extends StripeObjectInterface> T request(ApiRequest request, Type typeToken)
+      throws StripeException {
+    return request(
```
pakrym-stripe, line 24: Should we deprecate the old request method?

src/main/java/com/stripe/net/ApiRequest.java
richardm-stripe: Works for me!

richardm-stripe: Done in [BaseApiRequest](https://github.com/stripe/stripe-java/pull/1702/commits/8c03e55ed2987214ec34066f8c13d83c7eea26f2)

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
richardm-stripe: :'(

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
richardm-stripe: Yeah this kind of sucks -- the default implementation actually goes the other way to avoid breaking changes. `StripeResponseGetter.request(ApiRequest req)` is the new one that has the default implementation that delegates to `StripeResponseGetter.request(baseAddress, method, path, params, ...)`

src/main/java/com/stripe/net/StripeResponseGetter.java
richardm-stripe: SGTM

richardm-stripe: Done in [Address codereview (deprecate, fullurl)](https://github.com/stripe/stripe-java/pull/1702/commits/a2262ab34877d7159e52618c98db992385815fff)

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
richardm-stripe: I don't think there's any action that can be taken now. Long term, StripeRequest is fairly "internal" and probably possible to rename or otherwise phase out in the future.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
richardm-stripe: Done in [Address codereview (deprecate, fullurl)](https://github.com/stripe/stripe-java/pull/1702/commits/a2262ab34877d7159e52618c98db992385815fff)

richardm-stripe:
re-requesting review, would just like a stamp on the naming of 
[BaseApiRequest](https://github.com/stripe/stripe-java/pull/1702/commits/8c03e55ed2987214ec34066f8c13d83c7eea26f2)

anniel-stripe (approved):
BaseApiRequest looks good to me!

PR #1721, opened 2024-01-15
Report usage of `stripe_client`

richardm-stripe:
## Changelog
* Reports use of the new `StripeClient` in `X-Stripe-Client-Telemetry`. (You can disable telemetry via `Stripe.enableTelemetry = false;`, see the [README](https://github.com/stripe/stripe-java/blob/master/README.md#telemetry).)

## Details
* A different approach to #1698. This one sets `usage` on each callsite to `StripeResponseGetter.request` from `StripeClient`, rather than attempting to wrap `LiveStripeResponseGetter`.
* Modifies `RequestTelemetry` to accept usage and set it appropriately on `X-Stripe-Client-Telemetry`. Had to add a parameter to a public method, so I made a new overload and deprecated the old one. Had to disable a linter rule to do this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,6 +50,10 @@ tasks.withType(JavaCompile) {
       // com.stripe.param.AccountUpdateParams.Individual.Address] within this file.)
       // We should fix this by having autogen use the fully-qualified class to eliminate ambiguity.
       check("SameNameButDifferent", net.ltgt.gradle.errorprone.CheckSeverity.OFF)
+
+      // InlineMe (https://errorprone.info/docs/inlineme) seems neat, but in order to add these annotations
+      // we would be imposing another dependency on `errorprone` to our users, not worth it.
+      check("InlineMeSuggester", net.ltgt.gradle.errorprone.CheckSeverity.OFF)
     }
 }
 
--- a/src/main/java/com/stripe/net/ApiRequest.java
+++ b/src/main/java/com/stripe/net/ApiRequest.java
@@ -42,6 +42,7 @@ public class ApiRequest extends BaseApiRequest {
         this.getPath(),
         this.getOptions(),
         this.getApiMode(),
-        newUsage, this.getParams());
+        newUsage,
+        this.getParams());
   }
 }
--- a/src/main/java/com/stripe/net/ApiRequest.java
+++ b/src/main/java/com/stripe/net/ApiRequest.java
@@ -31,7 +31,7 @@ public class ApiRequest extends BaseApiRequest {
     this(baseAddress, method, path, options, apiMode, null, params);
   }
 
-  public ApiRequest withAddedUsage(String usage) {
+  public ApiRequest addUsage(String usage) {
     List<String> newUsage = new ArrayList<>();
     if (this.getUsage() != null) {
       newUsage.addAll(this.getUsage());
--- a/src/main/java/com/stripe/net/ApiRequest.java
+++ b/src/main/java/com/stripe/net/ApiRequest.java
@@ -1,11 +1,25 @@
 package com.stripe.net;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import lombok.Getter;
 
 @Getter
 public class ApiRequest extends BaseApiRequest {
   private Map<String, Object> params;
+
+  private ApiRequest(
+      BaseAddress baseAddress,
+      ApiResource.RequestMethod method,
+      String path,
+      RequestOptions options,
+      ApiMode apiMode,
+      List<String> usage,
+      Map<String, Object> params) {
+    super(baseAddress, method, path, options, apiMode, usage);
+    this.params = params;
+  }
 
   public ApiRequest(
       BaseAddress baseAddress,
@@ -14,7 +28,20 @@ public class ApiRequest extends BaseApiRequest {
       Map<String, Object> params,
       RequestOptions options,
       ApiMode apiMode) {
-    super(baseAddress, method, path, options, apiMode);
-    this.params = params;
+    this(baseAddress, method, path, options, apiMode, null, params);
+  }
+
+  public ApiRequest withAddedUsage(String usage) {
+    List<String> newUsage = new ArrayList<>();
+    if (this.getUsage() != null) {
+      newUsage.addAll(this.getUsage());
+    }
+    return new ApiRequest(
+        this.getBaseAddress(),
+        this.getMethod(),
+        this.getPath(),
+        this.getOptions(),
+        this.getApiMode(),
+        newUsage, this.getParams());
   }
 }
--- a/src/main/java/com/stripe/net/ApiRequest.java
+++ b/src/main/java/com/stripe/net/ApiRequest.java
@@ -36,6 +36,7 @@ public class ApiRequest extends BaseApiRequest {
     if (this.getUsage() != null) {
       newUsage.addAll(this.getUsage());
     }
+    newUsage.add(usage);
     return new ApiRequest(
         this.getBaseAddress(),
         this.getMethod(),
--- a/src/main/java/com/stripe/net/BaseApiRequest.java
+++ b/src/main/java/com/stripe/net/BaseApiRequest.java
@@ -1,8 +1,6 @@
 package com.stripe.net;
 
-import java.util.ArrayList;
 import java.util.List;
-
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
--- a/src/main/java/com/stripe/net/BaseApiRequest.java
+++ b/src/main/java/com/stripe/net/BaseApiRequest.java
@@ -14,6 +14,11 @@ class BaseApiRequest {
 
   @Setter private List<String> usage;
 
+  /** Internal method. Adds a record of a tracked behavior identified by "used" to the request. */
+  public void addUsage(String used) {
+    this.usage.add(used);
+  }
+
   public BaseApiRequest(
       BaseAddress baseAddress,
       ApiResource.RequestMethod method,
--- a/src/main/java/com/stripe/net/BaseApiRequest.java
+++ b/src/main/java/com/stripe/net/BaseApiRequest.java
@@ -2,37 +2,18 @@ package com.stripe.net;
 
 import java.util.ArrayList;
 import java.util.List;
+
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
 import lombok.Getter;
-import lombok.Setter;
 
 @Getter
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
 class BaseApiRequest {
-  private BaseAddress baseAddress;
-  private ApiResource.RequestMethod method;
-  private String path;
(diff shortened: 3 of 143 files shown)


build.gradle
```diff
@@ -50,6 +50,10 @@ tasks.withType(JavaCompile) {
       // com.stripe.param.AccountUpdateParams.Individual.Address] within this file.)
       // We should fix this by having autogen use the fully-qualified class to eliminate ambiguity.
       check("SameNameButDifferent", net.ltgt.gradle.errorprone.CheckSeverity.OFF)
+
+      // InlineMe (https://errorprone.info/docs/inlineme) seems neat, but in order to add these annotations
+      // we would be imposing another dependency on `errorprone` to our users, not worth it.
+      check("InlineMeSuggester", net.ltgt.gradle.errorprone.CheckSeverity.OFF)
```
richardm-stripe, line 56: Our linter, "errorprone" has a linter rule that suggests adding an annotation (to the implementation) that would introduce an additional dependency. This change disables that linter rule.

src/main/java/com/stripe/net/BaseApiRequest.java
```diff
@@ -14,6 +15,14 @@ class BaseApiRequest {
 
   @Setter private List<String> usage;
 
+  /** Internal method. Adds a record of a tracked behavior identified by "used" to the request. */
+  public void addUsage(String used) {
```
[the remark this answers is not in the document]
pakrym-stripe: Lol, this is not php

src/main/java/com/stripe/net/BaseApiRequest.java
```diff
@@ -14,6 +15,14 @@ class BaseApiRequest {
 
   @Setter private List<String> usage;
```
pakrym-stripe, line 16: Should this class be immutable? Or are we ok with it being mutable?

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -85,6 +92,15 @@ public StripeRequest(
     }
   }
 
+  public StripeRequest(
+      ApiResource.RequestMethod method,
+      String url,
+      Map<String, Object> params,
+      RequestOptions options)
+      throws StripeException {
+    this(method, url, params, options, new ArrayList<String>());
```
pakrym-stripe, line 101: Collections.emptyList() ?

src/test/java/com/stripe/functional/StripeClientTest.java
```diff
@@ -0,0 +1,49 @@
...
+  public void tearDown() {
+    Stripe.enableTelemetry = originalTelemetry;
+  }
+
+  @Test
+  public void testReportsUsageTelemetry() throws StripeException {
+    mockClient.customers().create();
+    Mockito.reset(httpClientSpy);
+    mockClient.customers().update("cus_xyz");
+    Mockito.verify(httpClientSpy)
```
pakrym-stripe, line 33: Do we not have a cute wrapper around httpClientSpy to make these things easier?

src/test/java/com/stripe/functional/LiveStripeResponseGetterTest.java
```diff
@@ -43,4 +51,44 @@ public void testInvalidJson() throws StripeException {
...
+    Stripe.enableTelemetry = originalTelemetry;
+  }
+
+  @Test
+  public void testGlobalClientTelemetryTest() throws StripeException {
+    StripeResponseGetter responseGetterSpy =
+        Mockito.spy(new LiveStripeResponseGetter(null, httpClientSpy));
+    // We need to explicitly override the ApiResource.responseGetter = networkSpy that's set
+    // in BaseStripeTest. Unfortunately, BaseStripeTest shares the response getter spy across
+    // StripeClient and the Global Client, and so the global client will erroneously report stripe
```
pakrym-stripe, line 74: Why does the sharing matter? Isn't the usage per-request?

src/test/java/com/stripe/BaseStripeTest.java
```diff
@@ -125,6 +125,9 @@ public void setUpStripeMockUsage() {
     httpClientSpy = Mockito.spy(new HttpURLConnectionClient());
     networkSpy = Mockito.spy(new LiveStripeResponseGetter(null, httpClientSpy));
     mockClient = new StripeClient(networkSpy);
+    // The StripeClient constructor calls .setUsage on networkSpy, let's reset that so
```
pakrym-stripe, line 128: Does StripeClient call setUsage?

pakrym-stripe (changes requested):
Few Qs

src/main/java/com/stripe/net/BaseApiRequest.java
richardm-stripe: Really conflicted here. I'll experiment.

src/test/java/com/stripe/functional/StripeClientTest.java
richardm-stripe: BaseStripeTest's existing wrappers operate on `networkSpy` which uses a mock `ResponseGetter`. I wanted this test to be a little closer to the metal to exercise more of the changes in this PR, so it uses `httpClientSpy`, which didn't have any wrappers -- but I added a small wrapper in [Improve test infra](https://github.com/stripe/stripe-java/pull/1721/commits/43d6ac95304a35384a40aec78f7de42a64259924)

src/test/java/com/stripe/functional/LiveStripeResponseGetterTest.java
richardm-stripe: Ah, well spotted. This was left over from the previous approach #1698 where it did matter. (I abandoned that approach specifically because of this ugliness). In [Move and simplify global client telemetry test](https://github.com/stripe/stripe-java/pull/1721/commits/859e4d62dfa59788bad561414a656364a48fb65c) I removed the `setStripeResponseGetter` and simplified the test, and I moved it into `TelemetryTest.java` because with the new approach it makes less sense for this to live in LiveStripeResponseGetterTest.java.

src/main/java/com/stripe/net/StripeRequest.java
richardm-stripe: Good idea, fixed!

src/main/java/com/stripe/net/BaseApiRequest.java
richardm-stripe: So the moral of the story is that immutability plays really badly with subclassing. I think [Immutable BaseApiRequest](https://github.com/stripe/stripe-java/pull/1721/commits/4f8883c32f302d1fe9822550b5a32629da012c9b) is an improvement, but it's a little unsatisfying because the helper to add "Usage" has to live in the child class instead of the parent. WDYT?

src/main/java/com/stripe/net/BaseApiRequest.java
```diff
@@ -2,37 +2,18 @@
...
 import java.util.List;
+
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
 import lombok.Getter;
-import lombok.Setter;
 
 @Getter
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
 class BaseApiRequest {
```
richardm-stripe, line 10: This should be a compatible change because `BaseApiRequest` is protected visibility.

pakrym-stripe: Compat check seems to be unhappy for some reason.

richardm-stripe: Looks like `ApiRequestor.setUsage` accidentally snuck onto the PR where I introduced `BaseApiRequest` :'(.

We can [Fix compat](https://github.com/stripe/stripe-java/pull/1721/commits/58244cad01711d1d702564c66a901a3cb84bfd4d).

src/main/java/com/stripe/net/ApiRequest.java
```diff
@@ -1,20 +1,49 @@
...
       String path,
       Map<String, Object> params,
       RequestOptions options,
       ApiMode apiMode) {
-    super(baseAddress, method, path, options, apiMode);
-    this.params = params;
+    this(baseAddress, method, path, options, apiMode, null, params);
+  }
+
+  public ApiRequest withAddedUsage(String usage) {
```
pakrym-stripe, line 34:
```suggestion
  public ApiRequest addUsage(String usage) {
```
? 

We use `add` prefix for collections in param builders.

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -73,6 +78,7 @@ public StripeRequest(
       this.url = buildURL(method, url, params);
       this.content = buildContent(method, params);
       this.headers = buildHeaders(method, this.options);
+      this.usage = usage;
```
pakrym-stripe, line 81: How is this used?

pakrym-stripe: I see `request.usage()` in `sendWithTelemetry`

pakrym-stripe: Should we add a parameter to `requestWithTelemetry` instead? Feels dirty to leak usage into `StripeRequest` just to carry it around.

richardm-stripe: Can you say more about the advantages of that? My current mental model is that the purpose of `StripeRequest` is to "carry around" values between `StripeResponseGetter.request` and `HttpClient.requestWithRetries` (and to provide a handle on this for tests) so adding another value doesn't feel dirty to me.

richardm-stripe: FWIW I think maybe a problem here is that responsibility for setting telemetry would feel better living in `StripeResponseGetter` than in `HttpClient`. I'd prefer that not be in scope here, but I was able to restrict the scope of `StripeRequest.usage` to the `com.stripe.net` package in [Restrict StripeRequest.usage visibility](https://github.com/stripe/stripe-java/pull/1721/commits/27097f394c0a1102552d9b83bb13349682f3c651), to prevent a breaking change if we want to do that migration later.

src/main/java/com/stripe/net/ApiRequest.java
richardm-stripe: Added in [withAddedUsage -> addUsage](https://github.com/stripe/stripe-java/pull/1721/commits/90cf4b9005cf73532fe09693a4cfa94048c03c98)

src/main/java/com/stripe/net/StripeRequest.java
```diff
@@ -51,20 +52,26 @@ public class StripeRequest {
   /** The special modifiers of the request. */
   RequestOptions options;
 
+  /** List of tracked behaviors associated with this request. */
+  @Getter(AccessLevel.PACKAGE)
```
pakrym-stripe, line 56: Is this still needed with your latest change?

pakrym-stripe (changes requested):
Can we now remove the usage from the StripeRequest?

pakrym-stripe (approved):

PR #1725, opened 2024-01-16
Move request telemetry to LiveStripeResponseGetter

richardm-stripe:
## Summary
Moves responsibility for calculating queuing/calculating telemetry payloads and setting the `X-Stripe-Client-Telemetry` header from `HTTPClient` to `LiveStripeResponseGetter`.

## Details
There are public `requestWithTelemetry` methods on HTTPClient that I changed to not trigger any (additional) telemetry anymore. It's a significant behavior change for a hypothetical user calling these methods directly, but I don't consider this a supported use.

## Motivation
HTTPClient should be responsible for sending already-calculated HTTP requests, it shouldn't be responsible for calculating what to put in the HTTP request itself. See discussion at https://github.com/stripe/stripe-java/pull/1721/files#r1454137531

## Test strategy

Pretty good existing end-to-end telemetry test: https://github.com/stripe/stripe-java/blob/c044fabc4e5bf15bf2e8a717070511befab1b266/src/test/java/com/stripe/functional/TelemetryTest.java#L25

### DIFF
--- a/src/main/java/com/stripe/net/RequestTelemetry.java
+++ b/src/main/java/com/stripe/net/RequestTelemetry.java
@@ -26,12 +26,18 @@ class RequestTelemetry {
    * metrics, or if telemetry is disabled, then the returned {@code Optional} is empty.
    *
    * @param headers the request headers
+   * @deprecated Use {$ling {@link #pollPayload()} instead.
    */
+  @Deprecated
   public Optional<String> getHeaderValue(HttpHeaders headers) {
     if (headers.firstValue(HEADER_NAME).isPresent()) {
       return Optional.empty();
     }
 
+    return this.pollPayload();
+  }
+
+  public Optional<String> pollPayload() {
     RequestMetrics requestMetrics = prevRequestMetrics.poll();
     if (requestMetrics == null) {
       return Optional.empty();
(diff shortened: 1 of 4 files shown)


src/main/java/com/stripe/net/RequestTelemetry.java
```diff
@@ -26,12 +26,18 @@ class RequestTelemetry {
    * metrics, or if telemetry is disabled, then the returned {@code Optional} is empty.
    *
    * @param headers the request headers
+   * @deprecated Use {$ling {@link #pollPayload()} instead.
    */
+  @Deprecated
   public Optional<String> getHeaderValue(HttpHeaders headers) {
```
pakrym-stripe, line 32: Any reason we can't reuse the original method?

pakrym-stripe (approved):
So clean!

src/main/java/com/stripe/net/RequestTelemetry.java
richardm-stripe: Nothing *prohibitive*, but the original method is awkward and makes the calling code awkward.

First, I think checking for duplicate telemetry headers is odd in the first place since there's really only one place it gets set.

Second, if we're going to check for duplicate headers, the same class should be responsible for setting the header and checking for duplicates. It's very awkward that `getHeaderValue` returns an `Optional<String>` i.e. it leaves the responsibility for actually putting the value onto the `HEADER_NAME` header inside an HttpHeaders object to the caller -- but accepts a `HttpHeaders` object and checks `HEADER_NAME` there directly.

richardm-stripe: Merging, but we can continue discussing and if I'm totally off-base we can bring it back in a follow-up.

PR #1728, opened 2024-01-17
Beta: Stripe.rawRequest -> StripeClient.rawRequest

richardm-stripe:
## Summary
This PR copies `Stripe.rawRequest` to `StripeClient.rawRequest` and deprecates the former. The only difference is that `StripeClient.rawRequest` is an instance method and uses `this.getStripeResponseGetter()` vs. the global responseGetter.

It also copies `Stripe.deserialize` to `StripeClient.deserialize`.

## Motivation
I figured I would do this in advance of adding `usage` to `.rawRequest`.


src/main/java/com/stripe/StripeClient.java
```diff
@@ -570,4 +573,59 @@ StripeResponseGetterOptions buildOptions() {
...
+   * Send raw request to Stripe API. This is the lowest level method for interacting with the Stripe
+   * API. This method is useful for interacting with endpoints that are not covered yet in
+   * stripe-java.
+   *
+   * @param method the HTTP method
+   * @param relativeUrl the relative URL of the request, e.g. "/v1/charges"
+   * @param content the body of the request as a string
+   * @return the JSON response as a string
+   */
+  @Deprecated
```
pakrym-stripe, line 587: nit: rem

src/main/java/com/stripe/StripeClient.java
```diff
@@ -570,4 +573,59 @@ StripeResponseGetterOptions buildOptions() {
...
+   * @param relativeUrl the relative URL of the request, e.g. "/v1/charges"
+   * @param content the body of the request as a string
+   * @param options the special modifiers of the request
+   * @return the JSON response as a string
+   */
+  public StripeResponse rawRequest(
+      final ApiResource.RequestMethod method,
+      final String relativeUrl,
+      final String content,
+      final RawRequestOptions options)
```
pakrym-stripe, line 610: check options for null? or default

src/main/java/com/stripe/StripeClient.java
richardm-stripe, line 587:
```suggestion
```

src/main/java/com/stripe/StripeClient.java
```diff
@@ -570,4 +573,58 @@ StripeResponseGetterOptions buildOptions() {
...
+                BaseAddress.API, method, relativeUrl, content, options, options.getApiMode()));
+  }
+
+  /** Deserializes StripeResponse returned by rawRequest into a similar class. */
+  public StripeObject deserialize(String rawJson) throws StripeException {
+    if (rawJson == null) {
+      throw new IllegalArgumentException("rawJson cannot be null");
+    }
+
+    return StripeObject.deserializeStripeObject(
```
pakrym-stripe, line 627: pass a response getter in and use the overload that takes a string?

pakrym-stripe (changes requested):
I'm not sure if we are flowing the response getter correctly

src/main/java/com/stripe/StripeClient.java
richardm-stripe: Done in [Deserialize takes in responseGetter](https://github.com/stripe/stripe-java/pull/1728/commits/7ae40fb7f986ec72627da053615634d311a86057)

src/main/java/com/stripe/StripeClient.java
richardm-stripe: Done. Added a new overload that doesn't require `Type`. Also added a test that the responsegetter flows through.

src/main/java/com/stripe/model/StripeObject.java
```diff
@@ -110,6 +110,17 @@ static StripeObject deserializeStripeObject(
     return object;
   }
 
+  public static StripeObject deserializeStripeObject(
+      String payload, StripeResponseGetter responseGetter) {
+    JsonObject jsonObject = ApiResource.GSON.fromJson(payload, JsonObject.class).getAsJsonObject();
+    String type = jsonObject.get("object").getAsString();
```
pakrym-stripe, line 116: call the JsonObject overload here? Instead of copying it's implementation

src/test/java/com/stripe/ApiResource.java
```diff
@@ -0,0 +1 @@
+
```
pakrym-stripe, line 1: nit : remove?

src/main/java/com/stripe/StripeClient.java
pakrym-stripe: Was it? The `options.getApiMode` call will still nullref if options is null.

pakrym-stripe (changes requested):
https://github.com/stripe/stripe-java/pull/1728#discussion_r1457596209

src/main/java/com/stripe/StripeClient.java
richardm-stripe: Good catch - must have screwed up my git. Here's the change plus a test: [Handle null RawRequestOptions better](https://github.com/stripe/stripe-java/pull/1728/commits/1332ee7460b7cf7652fa57735b32550ddfe85230)

pakrym-stripe (approved):

PR #1752, opened 2024-02-23
Add helper for setting beta versions

helenye-stripe:
To prevent users from setting the same beta version many times.

### DIFF
--- a/src/main/java/com/stripe/Stripe.java
+++ b/src/main/java/com/stripe/Stripe.java
@@ -42,7 +42,10 @@ public abstract class Stripe {
    */
   public static void addBetaVersion(String betaName, String betaVersion) {
     if (stripeVersion.indexOf("; " + betaName) >= 0) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/Stripe.java
```diff
@@ -34,6 +34,18 @@ public abstract class Stripe {
    */
   public static volatile String stripeVersion = API_VERSION;
 
+  /** Add a specified beta to the global Stripe API Version. Only call this method once per beta. */
+  public static void addBetaVersion(String betaName, String betaVersion) {
+    if (stripeVersion.indexOf("; " + betaName) >= 0) {
```
richardm-stripe, line 39:
```suggestion
    if (stripeVersion.indexOf("; " + betaName) >= 0 + "=") {
```
worrying about the case where one beta name is the prefix of another

helenye-stripe: Good point!

richardm-stripe (approved):

PR #1760, opened 2024-03-04
Enhancing Readability and Efficiency with Java `List.of()`

yasirshabbir44:
## Updated Usage of Java `List.of()`

In this project, we've updated the usage of `List.of()` to replace the older `Arrays.asList()` method. This not only makes the code more readable but also enhances efficiency by leveraging the newer features introduced in Java.

### What's Changed?

Previously, we were using `Arrays.asList()` to create lists, which returns a fixed-size list backed by the array. With Java's introduction of `List.of()` in Java 9, we can now create immutable lists easily without relying on an intermediate array.

### Benefits:

- **Readability**: The usage of `List.of()` provides a more concise and clear way to create lists compared to `Arrays.asList()`.
- **Efficiency**: `List.of()` creates immutable lists directly without the need for an intermediate array, enhancing efficiency and reducing memory overhead.

### Example:

```java
// Using Arrays.asList()
headerMap.put("Accept", Arrays.asList("application/json"));
// Using List.of()
headerMap.put("Accept", List.of("application/json"));



By adopting this updated feature of Java, we aim to maintain code readability and efficiency standards in our project.

### DIFF
--- a/src/main/java/com/stripe/net/HttpHeaders.java
+++ b/src/main/java/com/stripe/net/HttpHeaders.java
@@ -3,7 +3,6 @@ package com.stripe.net;
 import static java.util.Objects.requireNonNull;
 
 import com.stripe.util.CaseInsensitiveMap;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


helenye-stripe:
Hi! Thanks for the submission. Unfortunately, `List.of()` isn't supported in Java 1.8, which we support, so we cannot use it.

PR #1805, opened 2024-05-13
Added missing log to changelog

prathmesh-stripe:
Updated changelog to add missing line.


anniel-stripe (approved):
Nice! If you haven't done so already, we should also update the release tag descriptions at https://github.com/stripe/stripe-java/releases

PR #1874, opened 2024-09-25
Fixes requests with null params

jar-stripe:
### Why?
When testing the meter event stream examples, I saw error code 400 returned from the stripe server when creating a MeterEventSession (using MeterEventSessionService.create()).  The issue was that we were sending the string "null" as the json body of the request.  The fix here is to ensure we never call the GSON encoder with a null object; we can do this by testing for null in createHttpContent and substituting an empty map if needed.

### What?
- changed JsonEncoder.createHttpContent to test for null and create a new HashMap if true.  this ensures we never encode a "null"


helenye-stripe:
CI is unhappy but lgtm

helenye-stripe (approved):

PR #1875, opened 2024-09-25
Make all tests pass in VS Code Test Runner

jar-stripe:
### Why
One of the ApiResourceTest tests relies on a system property that is set in our gradle test config.  This was not being set when running the tests through VS Code Test Runner, resulting in a test failure.  To fix this, I set the flag in settings.json for when we run tests in Test Runner

### What
- added java.test.config block to settings.json and added "-Dstripe.disallowGlobalResponseGetterFallback=true" as a vmarg for tests

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -18,5 +18,9 @@
   },
   "java.configuration.updateBuildConfiguration": "automatic",
   // LSP was ooming and it recommended this change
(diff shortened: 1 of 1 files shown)


jar-stripe:
@xavdid-stripe not a high priority, this exists in master today; that said, its a pretty tiny fix.

xavdid-stripe (approved):

PR #1882, opened 2024-09-28
Re-add getConnectBase

helenye-stripe:
This was accidentally removed.


xavdid-stripe (approved):
Easy fix, but curious why didn't any CI go off to warn us this got removed

helenye-stripe:
I think it did, but this PR also broke a bunch of other compat and I missed it.

PR #1883, opened 2024-10-01
Fixed raw request deserialization for v1 and v2 Api endpoints

prathmesh-stripe:
Currently StripeClient.deserialize method could not parse V2 endpoints correctly. This caused a type cast exception: 

```java
class com.stripe.model.StripeRawJsonObject cannot be cast to class com.stripe.model.v2.billing.MeterEventSession
```

Fix involved passing ApiMode through to the deserialize function, searching the right EventDataLookup class and updating existing event deserialization calls.


src/main/java/com/stripe/model/Event.java
```diff
@@ -363,7 +364,7 @@ public static class Data extends StripeObject implements StripeActiveObject {
      */
     @Deprecated
     public StripeObject getObject() {
-      return StripeObject.deserializeStripeObject(object, this.responseGetter);
+      return StripeObject.deserializeStripeObject(object, this.responseGetter, ApiMode.V1);
```
ramya-stripe, line 367: Where is the v2 counterpart to this?

src/main/java/com/stripe/model/EventDataObjectDeserializer.java
```diff
@@ -190,7 +192,8 @@ public StripeObject deserializeUnsafe() throws EventDataObjectDeserializationExc
   public StripeObject deserializeUnsafeWith(CompatibilityTransformer transformer) {
     return StripeObject.deserializeStripeObject(
         transformer.transform(rawJsonObject.deepCopy(), apiVersion, eventType),
-        this.responseGetter);
+        this.responseGetter,
+        ApiMode.V1);
```
ramya-stripe, line 196: Where are the v2 counterparts to the changes in this file?

src/main/java/com/stripe/model/EventDataClassLookup.java
```diff
@@ -7,8 +7,10 @@
 /**
  * Event data class look up used in event deserialization. The key to look up is `object` string of
  * the model.
+ *
+ * <p>For internal use by Stripe SDK only.
  */
-final class EventDataClassLookup {
+public final class EventDataClassLookup {
```
ramya-stripe, line 13: Why did we have to make this public? Is it so that we can access it from StripeObject using the fully qualified name which was needed to differentiate between the v1 and v2 EventDataClassLookup? If that was the case that how about renaming this class to `V1EventDataClassLookup` and the other one to `V2EventDataClassLookup` and not making them pubic?

src/main/java/com/stripe/model/EventDataObjectDeserializer.java
prathmesh-stripe: This is a V1 only thing for now. 
https://github.com/stripe/stripe-java/blob/33589343406af111bc977877f7ab96313bbe5ada/src/main/java/com/stripe/model/EventDataObjectDeserializer.java#L40-L50

V2 uses [Eventtypeadapterfactory](https://github.com/stripe/stripe-java/blob/5cd2f1b3547be92ccf3d84292be0df9eee9bb41e/src/main/java/com/stripe/model/v2/EventTypeAdapterFactory.java#L11)

src/main/java/com/stripe/model/EventDataClassLookup.java
prathmesh-stripe: Yes, I thought of it but for getting this out faster I picked this approach.

src/main/java/com/stripe/model/Event.java
prathmesh-stripe: V2 events don't have a getObject method that deserializes event into an object class. The equivalent would be `fetchRelatedObject` which is already typed.

helenye-stripe (approved):
lgtm for now but could we file a follow up to clean up the public/private for q4?

PR #1893, opened 2024-10-03
Merge from master to beta

ramya-stripe:
There have been more changes in the master branch since the last time we merged from it to beta. Doing on last merge from master before this week's beta releases

### DIFF
--- a/src/main/java/com/stripe/net/RawRequestOptions.java
+++ b/src/main/java/com/stripe/net/RawRequestOptions.java
@@ -48,11 +48,6 @@ public class RawRequestOptions extends RequestOptions {
   public static final class RawRequestOptionsBuilder extends RequestOptions.RequestOptionsBuilder {
     private Map<String, String> additionalHeaders;
 
-    /** Constructs a raw request options builder with default values. */
-    public RawRequestOptionsBuilder() {
-      super();
-    }
-
     public Map<String, String> getAdditionalHeaders() {
       return this.additionalHeaders;
     }
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -394,7 +394,7 @@ public class RequestOptions {
   }
 
   protected static String normalizeStripeVersion(String stripeVersion) {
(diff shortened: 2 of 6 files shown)


src/test/java/com/stripe/model/StandardizationTest.java
```diff
@@ -0,0 +1,156 @@
+package com.stripe.model;
+
```
jar-stripe, line 2: didn't this file get removed?  cc @prathmesh-stripe

src/main/java/com/stripe/net/RawRequestOptions.java
```diff
@@ -48,11 +48,6 @@ public static RawRequestOptionsBuilder builder() {
   public static final class RawRequestOptionsBuilder extends RequestOptions.RequestOptionsBuilder {
     private Map<String, String> additionalHeaders;
 
-    /** Constructs a raw request options builder with default values. */
```
jar-stripe, line 51: do we not need this constructor?

ramya-stripe: It wasnt in the master branch, so I assumed we don't need it here either

jar-stripe (approved):
LGTM

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -394,7 +394,7 @@ protected static String normalizeClientId(String clientId) {
   }
 
   protected static String normalizeStripeVersion(String stripeVersion) {
-    // null stripeVersions are considered "valid" and use Stripe.stripeVersion
+    // null stripeVersions are considered "valid" and use Stripe.apiVersion
```
prathmesh-stripe, line 397: Shouldn't it be `Stripe.stripeVersion`? https://github.com/stripe/stripe-java/blob/fb280de8ed50a12a16bf3bee2e358db9e0034b46/src/main/java/com/stripe/Stripe.java#L30

ramya-stripe: You are right. I assumed what was in master to be the source of truth - didnt check the actual property name

PR #1909, opened 2024-10-25
Do not allow setting Stripe.stripeVersion

jar-stripe:
## Why?

When we introduced beta SDKs, we allowed users to directly update the global configuration for API Version since they needed to pass beta headers in https://github.com/stripe/stripe-java/pull/1424. We soon realized that was not safe and was error prone if users didnt pass in the right format and so introduced a helper method in https://github.com/stripe/stripe-java/pull/1752 and advertised that as the right way of doing things in the README

Proper deserialization of classes from Events can be guaranteed only when the Webhook API version matches the API version used to generate the SDKs. Therefore, in this PR we are dropping the ability to directly update the API version.

## What?

- Make `Stripe.stripeVersion` final and initialized to the API version
- Add `Stripe.stripeVersionWithBetaHeaders` with public getter to hold the full api version + beta version string, and updated RequestOptions to use this by default
- Update `EventDataObjectDeserializer` to use the API_VERSION directly, and update api version mismatch to include the full stripe version with beta headers
- Update version mismatch tests to change the incoming event version instead of setting stripeVersion directly
- Fix package string on unrelated test so that tests pass without error in VS code

## Changelog
* `Stripe.stripeVersion` is no longer settable. If you were using this to set the beta headers, use the helper method `Stripe.addBetaVersion()` instead.

### DIFF
--- a/src/main/java/com/stripe/Stripe.java
+++ b/src/main/java/com/stripe/Stripe.java
@@ -4,6 +4,7 @@ import java.net.PasswordAuthentication;
 import java.net.Proxy;
 import java.util.HashMap;
 import java.util.Map;
+import lombok.Getter;
 
 public abstract class Stripe {
   public static final int DEFAULT_CONNECT_TIMEOUT = 30 * 1000;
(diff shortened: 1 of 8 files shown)


jar-stripe:
For the reviewer: The EventDataObjectDeserializer tests took most of the time to figure out and as a result is the part that I think could most use review to make sure we still have the same amount of coverage.

xavdid-stripe (approved):

Issue #1899, opened 2024-10-16
CardException.getUserMessage() returns null in v27.1.0


tucuxi:
### Describe the bug

In Java SDK v27.1.0, when a `CardException` `e` is thrown, `e.getUserMessage()` returns `null`.

Java SDK v26.12.0 returns `Your card has insufficient funds.` 

### To Reproduce

1. Create a `PaymentIntent` for a card with insufficient funds.
2. Catch the `CardException`.

### Expected behavior

`CardException.getUserMessage()` returns `Your card has insufficient funds.` as in the previous version.

### Code snippets

```Java
The implementation of StripeException.getUserMessage() is as follows:

  public String getUserMessage() {
    if (this.getStripeError() != null) {
      return this.getStripeError().getUserMessage();
    }
    return null;
  }

stripeError contains the expected message in message, but not in userMessage.
```

### OS

macOS 15.0.1

### Java version

Java 21

### stripe-java version

v21.1.0

### API version

2024-09-30.acacia

nithieshnaik:
can i work on this issue

jar-stripe:
Thanks @tucuxi for the report, and thanks @nithieshnaik for your interest in helping out!  We have a fix for this in PR https://github.com/stripe/stripe-java/pull/1911 and will be releasing it shortly.  Will update here when it's available to use!

jar-stripe:
@tucuxi, @nithieshnaik: [v27.1.2](https://github.com/stripe/stripe-java/releases/tag/v27.1.2) is available and fixes this issue. Please let us know if you continue to have any trouble here!

PR #1911, opened 2024-10-25
Fix StripeException#getUserMessage on v1 API errors

jar-stripe:
### Why?
This issue https://github.com/stripe/stripe-java/issues/1899 reports that CardException#getUserMessage() returns null after upgrading to SDK v27.1.  The root cause is that the underlying StripeError#getUserMessage is only valid for v2 errors.  Because it is not present on v1 errors, it will be null in the StripeError object.  This PR fixes the issue by ensuring we only access the user message in StripeError for v2 errors.

### What
- Adds stripeErrorApiMode internal field to StripeException to store which API version returned the error
- Adds setStripeV2Error method to set the error and set the stripeErrorApiMode to V2; modified setStripeError to set the stripeErrorApiMode to V1
- Modifies StripeException getUserMessage to return error.getMessage() if we've wrapped a v1 error
- Modifies StripeException getMessage to not include user message if we've wrapped a v1 error
- Adds unit tests for setting v1 and v2 errors in StripeException

### DIFF
--- a/src/main/java/com/stripe/exception/StripeException.java
+++ b/src/main/java/com/stripe/exception/StripeException.java
@@ -2,6 +2,7 @@ package com.stripe.exception;
 
 import com.google.gson.JsonObject;
 import com.stripe.model.StripeError;
+import com.stripe.net.ApiMode;
 import com.stripe.net.StripeResponseGetter;
 import lombok.Getter;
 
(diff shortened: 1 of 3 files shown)


jar-stripe:
I was able to reproduce the issue and then validate the fix with:
```java
try {
  PaymentIntentCreateParams params =
      PaymentIntentCreateParams.builder()
          .setCurrency("usd")
          .setAmount(2000L)
          .setConfirm(true)
          .setAutomaticPaymentMethods(
              PaymentIntentCreateParams.AutomaticPaymentMethods.builder()
                  .setAllowRedirects(
                      PaymentIntentCreateParams.AutomaticPaymentMethods.AllowRedirects.NEVER)
                  .setEnabled(true)
                  .build())
          .setPaymentMethod("pm_card_chargeDeclinedProcessingError")
          .build();
  StripeClient client = new StripeClient(this.apiKey);
  PaymentIntent paymentIntent = client.paymentIntents().create(params);
  System.out.println("PaymentIntent created: " + paymentIntent);
} catch (StripeException e) {
  e.printStackTrace();
  System.out.println(e.getUserMessage());
}
```

xavdid-stripe (approved):

PR #1919, opened 2024-11-06
Catch `JsonSyntaxException` when processing all errors

### LINKED ISSUE #1868: stripeClient throw JsonSyntaxException instead of StripeException
### Describe the bug

Hi
After upgrading the com.stripe:stripe-java dependency from version 26.9.0 to 26.11.0, a test in our project fails:

The test expects stripeClient to throw a StripeException, but instead, a com.google.gson.JsonSyntaxException is thrown.
It looks like the JsonSyntaxException is thrown directly instead of being encapsulated within a StripeException like in the previous version

### To Reproduce

Create a test:
Configure WireMock to return an invalid JSON response when the stripeClient do a call
a JsonSyntaxException should be thrown

`
java.lang.AssertionError: 
Expecting actual throwable to be an instance of:
  com.stripe.exception.StripeException
but was:
  com.google.gson.JsonSyntaxException: Expected a com.google.gson.JsonObject but was com.google.gson.JsonPrimitive; at path $
	at com.google.gson.internal.bind.TypeAdapters$34$1.read(TypeAdapters.java:1152)
	at com.google.gson.Gson.fromJson(Gson.java:1361)
	at com.google.gson.Gson.fromJson(Gson.java:1262)
	...(93 remaining lines not displayed - this can be changed with Assertions.setMaxStackTraceElementsDisplayed)
`

### Expected behavior

stripeClient should throw a StripeException

### Code snippets

_No response_

### OS

linux

### Java version

21

### stripe-java version

v26.11.0

### API version

2024-06-20

### Additional context

_No response_

helenye-stripe:
## Why?

Fixes #1868.

## What?

We changed error handling to understand apiMode from the shape of the error response, but did not wrap all of our Gson calls in a try-catch. This fixes that and also adds a test.

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -6,7 +6,6 @@ import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonSyntaxException;
 import com.stripe.Stripe;
 import com.stripe.exception.*;
-import com.stripe.exception.ApiKeyMissingException;
 import com.stripe.exception.oauth.InvalidClientException;
 import com.stripe.exception.oauth.InvalidGrantException;
 import com.stripe.exception.oauth.InvalidScopeException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -289,22 +288,27 @@ private StripeError parseStripeError(
...
-      JsonPrimitive error = responseBody.getAsJsonPrimitive("error");
-      if (error.isString()) {
-        handleOAuthError(response);
+    try {
+      /*
+      OAuth errors are JSON objects where `error` is a string. In
+      contrast, in API errors, `error` is a hash with sub-keys. We use
+      this property to distinguish between OAuth and API errors.
+
+      Try to read the response body to see if it must be
```
jar-stripe, line 297: Is this missing the second part of the sentence?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -289,22 +288,27 @@ private StripeError parseStripeError(
...
+      } else if (apiMode == ApiMode.V2) {
+        handleV2ApiError(response);
+      } else {
+        handleV1ApiError(response);
       }
-    } else if (apiMode == ApiMode.V2) {
-      handleV2ApiError(response);
-    } else {
-      handleV1ApiError(response);
+    } catch (JsonSyntaxException e) {
```
jar-stripe, line 310: Not directly related to the issue but are there any other exceptions that can be thrown here that we want to wrap in StripeException?

jar-stripe (approved):
Looks good w/ a couple of non-blocking comments (well, that sentence fragment would be good to fix before we merge :) )

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -289,22 +288,27 @@ private StripeError parseStripeError(
...
+      Try to read the response body to see if it must be
+      */
+      JsonObject responseBody = ApiResource.GSON.fromJson(response.body(), JsonObject.class);
+      if (responseBody.has("error") && responseBody.get("error").isJsonPrimitive()) {
+        JsonPrimitive error = responseBody.getAsJsonPrimitive("error");
+        if (error.isString()) {
+          handleOAuthError(response);
+        }
+      } else if (apiMode == ApiMode.V2) {
+        handleV2ApiError(response);
```
jar-stripe, line 306: does it make sense to pass the already parsed responseBody into handle*ApiError?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
helenye-stripe: My mistake, fixed

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
helenye-stripe: It only expects to throw `JsonSyntaxException` https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5/com/google/gson/JsonSyntaxException.html. Should we wrap for all unexpected exceptions as well?

jar-stripe: I'm not sure what the norm for our SDKs is here.  I think if we don't have a concrete answer now, maybe we leave it for later.

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
helenye-stripe: I'm going to leave it as is for now

PR #1934, opened 2025-01-10
add justfile tweak readme, remove coveralls

xavdid-stripe:
### Why?

In an effort to modernize and simplify our local tooling, we're moving our dev commands from makefiles to [justfiles](https://github.com/casey/just). This is intended to be mostly a drop-in replacement, but some command names may change per standardization efforts.

### What?

- add justfile
- add just commands to CI
- tweak readme
- remove coveralls

### See Also

[DEVSDK-2325](https://go/j/DEVSDK-2325)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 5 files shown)


jar-stripe (approved):
LGTM

PR #1940, opened 2025-01-23
don't return dev messages in `getUserMessage()`

xavdid-stripe:
### Why?

Per guidance in [V2 error guidelines](http://go/th/trh_doc_OxndOeFIEt3KHL), the `user_message` response field can be shown directly to users, but the `message` field is meant only for developers. The previous java implementation (https://github.com/stripe/stripe-java/pull/1911) conflated the two, which I worry may be confusing for users. I think the intended use case here is someting like:

1. stripe error happened!
1. if .getUserMessage() != null, return that to the user that triggered this action
1. otherwise, check which error class it is and handle/log accordingly

### What?

- return `null` from `getUserMessage()` for v1 errors
- tweak docstrings
- simplify version checking logic
- update test

### See Also

- [DEVSDK-2262](https://go/j/DEVSDK-2262)
- [V2 error guidelines](http://go/th/trh_doc_OxndOeFIEt3KHL)

## Changelog

- `stripeException.getUserMessage()` now returns `null` for v1 errors instead of the developer-facing message
    - `getUserMessage()`, if non-null, can be passed directly to your users to explain what went wrong
    - if `getUserMessage()` is null, then you should handle the error as usual, per the [error handling docs](https://docs.stripe.com/error-handling?lang=java#catch-exceptions).

### DIFF
--- a/src/main/java/com/stripe/exception/StripeException.java
+++ b/src/main/java/com/stripe/exception/StripeException.java
@@ -64,13 +64,24 @@ public abstract class StripeException extends Exception {
   }
 
   /**
-   * Returns a developer-facing description of the exception, including the HTTP status code and
-   * request ID (if applicable).
+   * Returns a description of the issue suitable to show an end-user (if available).
+   *
+   * @return a string representation of the user facing exception (or `null`).
+   */
+  public String getUserMessage() {
+    if (this.stripeError == null) {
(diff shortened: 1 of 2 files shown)


xavdid-stripe:
Tabling for now, but we'll revisit in the future.

PR #1952, opened 2025-03-10
Support for StripeAccount in StripeClient, and StripeContext on V1

jar-stripe:
### Why?
The SDK should support setting StripeAccount and StripeContext at the client level, and StripeClient should be accepted in the same way across V1 and V2 APIs.

### What?
- added StripeAccount to StripeClientBuilder
- removed v1 limitation on using StripeContext
- added tests

## Changelog
* ⚠️ Adds `getStripeAccount` to `StripeResponseGetterOptions`.  If you have a class that extends `StripeResponseGetterOptions` you will need to implement this method.
* Adds `setStripeAccount` to StripeClientBuilder, so you can specify the Stripe-Account header at the client (instead of the individual request) level.

### DIFF
--- a/src/main/java/com/stripe/StripeClient.java
+++ b/src/main/java/com/stripe/StripeClient.java
@@ -431,6 +431,9 @@ public class StripeClient {
     private final String meterEventsBase;
 
     @Getter(onMethod_ = {@Override})
+    private final String stripeAccount;
+
+    @Getter(onMethod_ = {@Override})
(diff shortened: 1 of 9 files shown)


prathmesh-stripe (approved):

prathmesh-stripe:
One test is failing. Otherwise LGTM

jar-stripe:
compat error is expected, because we are adding a new abstract method to a public class.

PR #1970, opened 2025-04-01
beta SDKs should match API version exactly

xavdid-stripe:
### Why?

During workshop, we discussed tightening the API version during Event deserialization in beta SDKs. Because breaking changes happen more frequently and more unexpectedly in beta environments, we decided that the beta SDKs should do an exact API version match instead of only matching on a release train.

### What?

- tweak `apiVersionMatch` to just do an exact string match
- update test to invert assertion


xavdid-stripe:
closing because we need a different approach

PR #1971, opened 2025-04-01
require beta SDKs match event versions exactly when deserializing

xavdid-stripe:
### Why?

We decided that beta SDKs should exactly match the API version for events they're deserializing to minimize the chances of accidental breaking changes. This PR tweaks the version matching function to accommodate that change

### What?

- add check for `preview` API release trains (and require an exact version match)
- add tests

## Changelog

- ⚠️ Public Preview SDKs now require an exact API version match when deserializing webhooks
    - To deserialize events from another API version, use `EventDataObjectDeserializer .deserializeUnsafe()` instead of `.deserialize()`.


mbroshi-stripe (approved):
Looks great!

PR #1975, opened 2025-04-07
Xavdid/merge java beta & fix version matching tests

xavdid-stripe:
### Why?

The merge script had a conflict which I fixed. But _then_ the tests I wrote in https://github.com/stripe/stripe-java/pull/1971 were failing. I added the helper methods we have in master and an extra test (noted in description)

### What?

- `just merge java-beta`
- copy code that was in master to support mocking API versions in tests
- fix some tests

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -19,8 +19,8 @@
   "java.configuration.updateBuildConfiguration": "automatic",
   // LSP was ooming and it recommended this change
   // (jar) added -Xss8m so lombok would run without stack overflowing
-  "java.jdt.ls.vmargs": "-XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xmx2G -Xms100m -Xlog:disable -Xss8m",
+  "java.jdt.ls.vmargs": "-XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xmx4G -Xms100m -Xlog:disable -Xss8m",
   "java.test.config": {
-    "vmargs": [ "-Dstripe.disallowGlobalResponseGetterFallback=true"]
-  },
+    "vmargs": ["-Dstripe.disallowGlobalResponseGetterFallback=true"]
+  }
 }
--- a/src/main/java/com/stripe/model/EventDataObjectDeserializer.java
+++ b/src/main/java/com/stripe/model/EventDataObjectDeserializer.java
@@ -203,9 +203,9 @@ public class EventDataObjectDeserializer {
 
     // Preserved for testing; we have tests that hook getIntegrationApiVersion
     // to test with other api versions.
-    String currentApiVersion = getIntegrationApiVersion();
-    if (!currentApiVersion.contains(".")) {
-      return eventApiVersion.equals(currentApiVersion);
+    String sdkApiVersion = getIntegrationApiVersion();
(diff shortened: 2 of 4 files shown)


.vscode/settings.json
```diff
@@ -19,8 +19,8 @@
   "java.configuration.updateBuildConfiguration": "automatic",
   // LSP was ooming and it recommended this change
   // (jar) added -Xss8m so lombok would run without stack overflowing
-  "java.jdt.ls.vmargs": "-XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xmx2G -Xms100m -Xlog:disable -Xss8m",
```
xavdid-stripe, line 22: auto-fix based on the java extension's suggestion after the java process was OOMing

src/test/java/com/stripe/model/EventDataObjectDeserializerTest.java
```diff
@@ -233,6 +234,13 @@ public void testGetRawJson() throws Exception {
     assertEquals(event.getData().object.toString(), rawJson);
   }
 
+  private EventDataObjectDeserializer stubIntegrationApiVersion(
```
xavdid-stripe, line 237: this is the same thing we have in master

xavdid-stripe (commented):
This _hopefully_ won't cause additional merge conflicts, but they should be easier to resolve going forward

.vscode/settings.json
jar-stripe: Any thoughts about bumping it up higher to say 8G?  Our dev machines have a lot more available memory and bumping this more would hedge against an issue where 4G technically works but is thrashing the GC

src/test/java/com/stripe/model/EventDataObjectDeserializerTest.java
```diff
@@ -244,4 +252,53 @@ public void testDeserializeSetsResponseGetter() throws Exception {
     assertTrue(deserializer.getObject().isPresent());
     verifyDeserializedStripeObject(deserializer.getObject().get());
   }
+
```
jar-stripe, line 255: should this live both in beta and master?  some of the tests may not make sense in GA but it could reduce future merge conflicts

src/main/java/com/stripe/model/EventDataObjectDeserializer.java
```diff
@@ -198,10 +198,14 @@ public StripeObject deserializeUnsafeWith(CompatibilityTransformer transformer)
   }
 
   private boolean apiVersionMatch() {
-
+    // the pure event API version, minus any beta signifiers
```
jar-stripe, line 201: were there any merge conflicts with the code changes in this method?  this was the one I was afraid of w.r.t. java, because its so different from master to beta

xavdid-stripe: I copied the GA version of this method over into beta, so besides some renames, there weren't conflicts. Most of the issue was from helper methods having never existed in beta

.vscode/settings.json
xavdid-stripe: no thoughts, happy to go higher! I just saw the error and clicked the "fix it" button.

The thing I don't know is if I should make that here or let this settle and try to make the fixes in master instead.

src/test/java/com/stripe/model/EventDataObjectDeserializerTest.java
xavdid-stripe: yes, I can copy the one new test over (most of them were already brought in from master). But, bold of you to assume we'll avoid any merge conflicts anyway 😂

PR #1987, opened 2025-04-15
Include Java versions 21, 22, 23, 24 in CI

prathmesh-stripe:
### Why?
In order to keep supporting new versions of PHP we need to test against them.

### What?
Added Java versions 21, 22, 23, 24 to our CI workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,10 @@ jobs:
           - "18"
           - "19"
           - "20"
+          - "21"
+          - "22"
+          - "23"
+          - "24"
 
     steps:
       - uses: extractions/setup-just@v2


xavdid-stripe (commented):
I think it makes sense to add these because they're all current. But we probably don't need them long term. Basically all of our users are on LTS java versions (8, 11, 17, and 21).

Given that, in this PR, I'd also:

- drop any older non-LTS versions we're currently testing
- add a link to the [java version support list](https://www.oracle.com/java/technologies/java-se-support-roadmap.html) so future readers know why we're testing 
- add a comment to codify which versions we should include in this list and why

EncriptionFJ (commented):
Balance account cards partner bank

PR #2010, opened 2025-05-30
build SDK w/ new spec

xavdid-stripe:
DON'T MERGE. This is just to see the output of snapshots built w/ openapi v2


src/main/java/com/stripe/events/V1BillingMeterErrorReportTriggeredEvent.java
```diff
@@ -35,7 +35,7 @@ public static final class EventData {
     public static final class Reason {
       /** The total error count within this window. */
       @SerializedName("error_count")
-      Integer errorCount;
+      Long errorCount;
```
xavdid-stripe, line 38: I think this is a bug: https://go/j/DEVSDK-2604

src/main/java/com/stripe/param/v2/core/EventDestinationUpdateParams.java
```diff
@@ -200,30 +200,12 @@ public Builder putMetadata(String key, String value) {
       return this;
     }
 
-    /**
-     * Add a key/value pair to `metadata` map. A map is initialized for the first `put/putAll` call,
-     * and subsequent calls add additional key/value pairs to the original map. See {@link
-     * EventDestinationUpdateParams#metadata} for the field documentation.
-     */
-    public Builder putMetadata(String key, EmptyParam value) {
```
xavdid-stripe, line 208: this is expected, since metadata is a `Map<String, String>` now

PR #2019, opened 2025-06-26
Update Sonatype endpoints in deploy.gradle

mbroshi-stripe:
### Why?

On June 30, 2025, the OSSRH publishing service run by Sonatype will be sunset: https://central.sonatype.org/publish/publish-guide/.

As part of their migration guide, we need to update the endpoints we use are part of publishing through Gradle: https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#configuration.

### What?

- Updates the hostnames in the Gradle publish enpoint from `oss.sonatype.org` to `ossrh-staging-api.central.sonatype.com`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,8 +1,6 @@
 buildscript {
     repositories {
         mavenCentral()
-        //Add only for SNAPSHOT versions
-        //maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
     }
 }
 
--- a/deploy.gradle
+++ b/deploy.gradle
@@ -15,12 +15,12 @@ def isReleaseBuild() {
 
 def getReleaseRepositoryUrl() {
     return hasProperty("RELEASE_REPOSITORY_URL") ? RELEASE_REPOSITORY_URL
-            : "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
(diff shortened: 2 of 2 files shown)


xavdid-stripe (approved):
It's interesting that it's a staging API! It's in their docs so I assume it's what they want, but I would have assumed like, a v2 or something.

Either way, thanks for grabbing!

mbroshi-stripe:
> It's interesting that it's a staging API! It's in their docs so I assume it's what they want, but I would have assumed like, a v2 or something.

It is interesting! From what I can tell, the "official" endpoint [doesn't support Gradle](https://central.sonatype.org/publish/publish-portal-gradle/)

> Currently, there is no official Gradle plugin for publishing to Maven Central via the Central Publishing Portal. We have received significant feedback that publishing via Gradle is important to the community so we wanted to communicate that Gradle support is on our roadmap.

So, I think this "staging" endpoint is a stopgap for the migration. From https://central.sonatype.org/publish/publish-portal-ossrh-staging-api:

> The Portal OSSRH Staging API is a partial reimplementation of the OSSRH / Nexus Repository Manager 2 Staging APIs. The intent of this service is to enable publishers who are using existing plugins to have a smooth migration to the Central Publisher Portal. We hope that over time plugins will adopt the [Portal API](https://central.sonatype.org/publish/publish-portal-api/) rather than rely on this service.

PR #2036, opened 2025-08-29
⚠️ Add strongly typed EventNotifications

xavdid-stripe:
]### Why?

As part of the Next-Gen Event Handling project, we're adding fully typed classes to represent all the ~~thin events~~ Event Notifications a user could receive. There are also helper methods that help with the event handling experience.

### What?

- (see changelog below for user-facing changes)
- add tests
- add `getMode` helper for `ApiMode`
- fix doc typos

### See Also

[DEVSDK-2662](https://go/j/browse/DEVSDK-2662)

## Changelog

We've overhauled how V2 Events are handled in the SDK! This approach should provide a lot more information at authoring and compile time, leading to more robust integrations. As part of this process, there are a number of changes to be aware of.

- Added matching `EventNotification` classes to every v2 `Event`. For example, there's now a `V1BillingMeterErrorReportTriggeredEventNotification` to match the existing `V1BillingMeterErrorReportTriggeredEvent`. Each notification class defines a `fetchEvent()` method to retrieve its corresponding event. For events with related objects, there's a `fetchRelatedObject()` method that performs the API call and casts the response to the correct type.
- ⚠️ Rename function `StripeClient.parseThinEvent` to `StripeClient.parseEventNotification` and remove the `Stripe.ThinEvent` class.
    - This function now returns a `Stripe.V2.Core.EventNotification` (which is the shared base class that all of the more specific `Stripe.Events.*EventNotification` classes  share) instead of `Stripe.ThinEvent`. When applicable, these event notifications will have the `relatedObject` property and a `fetchRelatedObject()` function. They also have a `fetchEvent()` method to retrieve their corresponding `Stripe.Event.*Event` instance.
    - If you parse an event the SDK doesn't have types for (e.g. it's newer than the SDK you're using), you'll get an instance of `Stripe.Events.UnknownEventNotification` instead of a more specific type. It has both the `relatedObject` property and the `fetchRelatedObject()` function (but they may be/return `null`)

### DIFF
--- a/src/main/java/com/stripe/model/v2/EventNotification.java
+++ b/src/main/java/com/stripe/model/v2/EventNotification.java
@@ -82,9 +82,9 @@ public abstract class EventNotification {
 
   /**
    * Helper for constructing an Event Notification. Doesn't perform signature validation, so you
-   * should use {@link StripeClient#parseEventNotification() parseEventNotification} instead for
-   * initial handling. This is useful in unit tests and working with EventNotifications that you've
-   * already validated the authenticity of.
+   * should use {@link com.stripe.StripeClient#parseEventNotification} instead for initial handling.
+   * This is useful in unit tests and working with EventNotifications that you've already validated
+   * the authenticity of.
    */
   public static EventNotification fromJson(String payload, StripeClient client) {
     // don't love the double json parse here, but I don't think we can avoid it?


src/main/java/com/stripe/events/V1BillingMeterErrorReportTriggeredEvent.java
```diff
@@ -5,6 +5,7 @@
 import com.stripe.exception.StripeException;
 import com.stripe.model.billing.Meter;
 import com.stripe.model.v2.Event;
+import com.stripe.model.v2.Event.RelatedObject;
```
xavdid-stripe, line 8: this was added to simplify some code in codegen. It has no effect despite not technically being wrong

src/main/java/com/stripe/model/v2/EventNotification.java
```diff
@@ -0,0 +1,103 @@
...
+  public com.stripe.model.v2.Event.Reason reason;
+
+  @Getter(AccessLevel.NONE)
+  protected transient StripeClient client;
+
+  /**
+   * Helper for constructing an Event Notification. Not recommended for production code (because it
+   * lacks signature validation) but useful in testing. StripeClient uses it under the hood.
+   */
+  public static EventNotification fromJson(String payload, StripeClient client) {
```
xavdid-stripe, line 58: this is new

xavdid-stripe: same comment as go - users might like to use it for their own tests. But we can start it protected and open it up later, if we want

xavdid-stripe: also it looks like we cant use it in tests if it's private. The more I think about it though, the more broadly useful it is. Like, if someone validates a webhook and then saves the string to a database. when they pull it out later, they'll want the full `EventNotification`, but don't need to validate it

src/main/java/com/stripe/events/V1BillingMeterErrorReportTriggeredEvent.java
jar-stripe: im curious what the issue is and if we can eventually remove it, but no harm for now

src/main/java/com/stripe/model/v2/EventNotification.java
xavdid-stripe: unfortunately, the client and the model are different packages, so it yells. public it is!

Issue #2012, opened 2025-06-03
StripeClientBuilder throws an exception with null apiKey


delor:
### Describe the bug

When initializing `StripeClient` using `StripeClient.builder()`, providing a `null` value for the API key via `.setApiKey(null)` is allowed by the builder's interface. However, calling `.build()` on this builder results in an `IllegalArgumentException` because a global API key is expected at build time.

In a multi-tenant environment, it's common to use a single `StripeClient` instance and provide the specific tenant's API key with each request using `RequestOptions.builder().setApiKey("sk_...")`. In such scenarios, a global API key for the `StripeClient` itself is not necessary and might even be undesirable.

Currently, the SDK mandates a non-null API key during `StripeClient` construction, even if `RequestOptions` will be used for every subsequent API call.

### To Reproduce

Attempt to build a `StripeClient` with a null API key:
    ```java
    try {
        StripeClient client = StripeClient.builder()
            .setApiKey(null) // Explicitly setting null, or not calling setApiKey if it defaults to null internally and is checked
            .build();
        // Potentially make a call, though the build() itself might fail
        // CustomerCollection customers = client.customers().list(RequestOptions.builder().setApiKey("sk_test_specific_tenant").build());
    } catch (Exception e) {
        e.printStackTrace(); // Observe the exception during build()
    }
    ```

### Expected behavior

It should be possible to construct a `StripeClient` instance without a globally configured API key (i.e., with `apiKey` being `null` or not set on the builder) if the intention is to always provide the API key per request via `RequestOptions`.

The `StripeClient.build()` method should not throw an exception if no global API key is provided, and the responsibility for providing the API key should then fall entirely on the `RequestOptions` for each API call. If an API call is made without an API key in `RequestOptions` (and no global key is set), then an exception at that point would be appropriate.

### Code snippets

```Java

```

### OS

macOS

### Java version

Java 21

### stripe-java version

29.0.0

### API version

2025-03-31.basil

### Additional context

`StripeClientBuilder.setApiKey` does handle `null` `apiKey` so I don't get it why it's not accepted when the client is build.
https://github.com/stripe/stripe-java/blob/d177cbc62bb595399e298288fe9803c1d213a3ac/src/main/java/com/stripe/StripeClient.java#L519-L526

helenye-stripe:
[...]

PR #2041, opened 2025-09-14
Throw `ApiKeyMissingException` instead of `IllegalArgumentException` for unset API keys

elnafateh:
### Why?
`BearerTokenAuthenticator` threw an IllegalArgumentException with a vague message when given a null API key. This was inconsistent with the rest of the SDK.

### What?
- Throw ApiKeyMissingException instead
- Improve the message to: "API key cannot be null. Set Stripe.apiKey"

### Notes
- This is an intentionally minimal and backward-compatible change.

### Future Work
A possible improvement for multi-tenant environments would be allowing StripeClient construction without requiring a global API key, instead relying solely on RequestOptions for per-request keys.
- Related issue: #2012 

## Changelog

- ⚠️ throw `ApiKeyMissingException` instead of `IllegalArgumentException` when making a request before setting an API key initializing a client without an API key. Tweak the wording of that message to better indicate how to fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ public class StripeExample {
                 .build();
 
         try {
-            Customer customer = client.customers().create(params);
+            Customer customer = client.v1().customers().create(params);
             System.out.println(customer);
         } catch (StripeException e) {
             e.printStackTrace();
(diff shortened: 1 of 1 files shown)


elnafateh:
@xavdid-stripe it's done

xavdid-stripe (approved):

PR #2057, opened 2025-09-25
Add `StripeContext` object

xavdid-stripe:
### Why?

As the `stripe-context` header is evolving, we want to give better tools for creating and managing contexts. This PR adds a new class, `StripeContext`. It can be used anywhere the context option is supplied and gets serialized to a string when making requests. It's also found on the `EventNotification.context` property.

### What?

- add `StripeContext` class
- allow it to be supplied for RequestOptions.context; serialize it into strings
- add tests

## Changelog

- Add the `StripeContext` class
- ⚠️ Change `EventNotification` (formerly known as `ThinEvent`)'s `context` property from `string` to `StripeContext`


src/main/java/com/stripe/StripeClient.java
```diff
@@ -1223,6 +1223,11 @@ public StripeClientBuilder setStripeContext(String context) {
       return this;
     }
 
+    public StripeClientBuilder setStripeContext(StripeContext context) {
+      this.stripeContext = context == null ? null : context.toString();
+      return this;
+    }
+
     public String getStripeContext() {
       return this.stripeContext;
```
mbroshi-stripe, line 1232: I find this a little surprising that the setter takes a different type than the getter returns. I would expect `StripeContext` to be serialized as a string, but would expect `getStripeContext` to return ` StripeContext`

src/main/java/com/stripe/StripeContext.java
```diff
@@ -0,0 +1,98 @@
...
+ * <p>You can use it whenever you're initializing a `StripeClient` or sending `stripe_context` with
+ * a request. It's also found in the `EventNotification.context` property.
+ */
+@EqualsAndHashCode
+public final class StripeContext {
+  private final List<String> segments;
+
+  /** Creates a new StripeContext with no segments. */
+  public StripeContext() {
+    this.segments = Collections.emptyList();
```
mbroshi-stripe, line 22: Nit: Maybe `this(null)` or `this(Collections.emptyList())`

src/main/java/com/stripe/model/v2/EventNotification.java
```diff
@@ -71,7 +72,10 @@ public static class Reason {
 
   /** [Optional] Authentication context needed to fetch the event or related object. */
   @SerializedName("context")
-  public String context;
+  private String contextString;
+
+  /** [Optional] Authentication context needed to fetch the event or related object. */
+  private transient StripeContext contextObject;
```
mbroshi-stripe, line 78: Why not just have a single internal representation?

mbroshi-stripe (commented):
Overall feedback is that I'd keep a single representation of `StripeContext` that does the right thing when it comes to serialization/deserialization rather than switching between object and string representations internally.

src/main/java/com/stripe/StripeClient.java
xavdid-stripe: There are 2 setters (string and `StripeContext`) and 1 getter (string). Changing the type of the `stripeContext` on the client itself had implications for the types for `RequestOptions` as well, which felt like a bigger breaking change.

Since I didn't figure folks will be retrieving context from anything but the EventNotification, keeping everything the same but being able to set a `StripeContext` object felt cleanest.

src/main/java/com/stripe/model/v2/EventNotification.java
```diff
@@ -70,8 +72,7 @@ public static class Reason {
   public Boolean livemode;
 
   /** [Optional] Authentication context needed to fetch the event or related object. */
-  @SerializedName("context")
-  public String context;
+  private transient StripeContext contextObject;
```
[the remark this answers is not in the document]
mbroshi-stripe: 1. The main change we should make before merging is to remove the `public parse` method. For now, you can make it a `private` method and move it into this class.
2. To clean up the current code, we should push everything into the deserialization layer. That means adding a new TypeAdapter [here](https://github.com/stripe/stripe-java/blob/281058289e4e045848ee22dc1aa5c6c1f9947006/src/main/java/com/stripe/net/ApiResource.java#L58). That can be a follow-on.
3. Also, I don't think you need the `transient` keyword unless I'm missing some subtlety.

PR #2104, opened 2025-11-05
Add EventNotificationHandler

xavdid-stripe:
### Why?

We've been designing a streamlined approach to handling incoming events that is easy to get right and hard to get wrong. This PR has the initial implementation of this new system.

The only other pending item is to add a method to allow handling a webhook without verifying the signature. This is good for testing and for Event Bridge, which doesn't use the signature-based verification. Otherwise, this is ready for review.

### What?

- add `EventHandler` class
- add event handler constructor on `StripeClient`
- add tests
- add method for changing the context of an existing client

### Example usage

```java
public class EventNotificationWebhookHandler {
  private static final String API_KEY = System.getenv("STRIPE_API_KEY");
  private static final String WEBHOOK_SECRET = System.getenv("WEBHOOK_SECRET");

  private static final StripeClient client = new StripeClient(API_KEY);
  private static final StripeEventNotificationHandler handler =
      client.notificationHandler(
          WEBHOOK_SECRET, EventNotificationWebhookHandler::fallbackCallback);

  public static void main(String[] args) throws IOException {
    handler.onV1BillingMeterErrorReportTriggered(
        EventNotificationWebhookHandler::handleMeterErrors);

    HttpServer server = HttpServer.create(new InetSocketAddress(4242), 0);
    server.createContext("/webhook", new WebhookHandler());
    server.setExecutor(null);
    server.start();
  }

  private static void fallbackCallback(
      EventNotification notif, StripeClient client, UnhandledNotificationDetails details) {
    System.out.println("Received unhandled event notification type: " + notif.getType());
  }

  private static void handleMeterErrors(
      V1BillingMeterErrorReportTriggeredEventNotification notif, StripeClient client) {
    Meter meter = notif.fetchRelatedObject();
    System.out.println("Handling meter error for meter: " + meter.getDisplayName());
  }

  static class WebhookHandler implements HttpHandler {
    @Override
    public void handle(HttpExchange exchange) throws IOException {
      if ("POST".equals(exchange.getRequestMethod())) {
        InputStream requestBody = exchange.getRequestBody();
        String webhookBody = new String(requestBody.readAllBytes(), StandardCharsets.UTF_8);
        String sigHeader = exchange.getRequestHeaders().getFirst("Stripe-Signature");

        try {
          handler.handle(webhookBody, sigHeader);

          exchange.sendResponseHeaders(200, -1);
        } catch (StripeException e) {
          exchange.sendResponseHeaders(400, -1);
        }
      } else {
        exchange.sendResponseHeaders(405, -1);
      }
      exchange.close();
    }
  }
}
```

### See Also

- [DEVSDK-2822](https://go/j/DEVSDK-2822)
- https://github.com/stripe/stripe-java/pull/2036

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Function;
 
 public class LiveStripeResponseGetter implements StripeResponseGetter {
   private final HttpClient httpClient;
@@ -87,8 +88,7 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
    * @return a new LiveStripeResponseGetter with the updated options and the same HTTP client
    */
   public LiveStripeResponseGetter withNewOptions(
-      java.util.function.Function<StripeResponseGetterOptions, StripeResponseGetterOptions>
-          contextCreator) {
+      Function<StripeResponseGetterOptions, StripeResponseGetterOptions> contextCreator) {
     StripeResponseGetterOptions newOptions = contextCreator.apply(this.options);
     return new LiveStripeResponseGetter(newOptions, this.httpClient);
   }


src/main/java/com/stripe/StripeClient.java
```diff
@@ -41,6 +41,50 @@ protected StripeResponseGetter getResponseGetter() {
...
+    return options.getStripeContext();
+  }
+
+  /**
+   * Sets the StripeContext for this client. This updates the client configuration to use the
+   * provided context for all subsequent requests.
+   */
+  protected void setContext(String context) {
+    // TODO(major): add getOptions to the StripeResponseGetter interface? that would simplify this
+    if (!(responseGetter instanceof LiveStripeResponseGetter)) {
```
xavdid-stripe: I don't love this and am all ears for a better suggestion!

mbroshi-stripe: Meh, not much you can do. Seems like the wrong abstractions were put in place (or they made the common Java mistake of creating an interface when there was only ever going to be a single implementation)

src/main/java/com/stripe/StripeClient.java
```diff
@@ -41,6 +41,73 @@ protected StripeResponseGetter getResponseGetter() {
...
+   * @return a new StripeClient with the custom context
+   * @throws IllegalStateException if this client doesn't use a LiveStripeResponseGetter
+   */
+  public StripeClient withStripeContext(StripeContext context) {
+    // Convert StripeContext to String
+    String contextString = (context == null) ? null : context.toString();
+
+    StripeResponseGetter responseGetter = this.getResponseGetter();
+
+    // We can only create a new client for LiveStripeResponseGetter
```
mbroshi-stripe: Nit, but can you clarify the reason for this in the comment? The following code is already self-documenting that you can't do it. I think the reason is that you need the `getOptions()` method, is that right?

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -76,6 +76,27 @@ public LiveStripeResponseGetter(StripeResponseGetterOptions options, HttpClient
...
+   * instance, but with a different stripe_context. This allows for efficient cloning when you need
+   * to make requests with different contexts (e.g., webhook processing) without reinitializing HTTP
+   * connections.
+   *
+   * @param contextCreator a function that takes the existing options and returns new options with
+   *     the desired context
+   * @return a new LiveStripeResponseGetter with the updated options and the same HTTP client
+   */
+  public LiveStripeResponseGetter withNewOptions(
+      java.util.function.Function<StripeResponseGetterOptions, StripeResponseGetterOptions>
```
mbroshi-stripe: Can't you import `java.util.function.Function`? Generally in Java, we import classes unless a name conflict forces you to use the fully qualified class name.

mbroshi-stripe (approved):
Beautiful! Added some minor feedback, but otherwise looks great!

PR #2131, opened 2025-12-18
⚠️ Use HTTP status code in V2 errors

mbroshi-stripe:
### Why?

We did an audit of V2 error-handling across our SDKs, and realized that in V2 APIs, Java users do not get typed exceptions based on HTTP Status Code like they do when calling a V1 API. E.g. a `429` response code from Stripe in V1 throws a `RateLimitException`, but against a V2 API, it will throw a more generic `ApiException`. Also, as part of our audit, we decided that it makes more sense to make `RateLimitException` extend `StripeException` rather than `ApiException` (since it does not come explicitly from the Stripe API). The reason it extends `ApiException` is that we anticipated that it would be in the API (it was briefly in Private Preview), which was a change we made in https://github.com/stripe/stripe-java/pull/2046.

### What?

- Refactors V1 and V2 code paths so they delegate to the same HTTP Status Code based logic
- `RateLimitException` now extends `StripeException` rather than `ApiException`
- Tests for all the edge cases of the new behavior

## Changelog
- ⚠️ `RateLimitException` now extends `StripeException` rather than `ApiException`
- ⚠️ Non-200 status codes from V2 endpoints will throw a `StripeException` (e.g. `RateLimitException`, `InvalidRequestException`, etc.) like in V1 instead of a generic `ApiException`


src/main/java/com/stripe/exception/RateLimitException.java
```diff
@@ -7,7 +7,7 @@
 import lombok.Getter;
 
 @Getter
-public class RateLimitException extends ApiException {
+public class RateLimitException extends StripeException {
```
prathmesh-stripe: Can you please add a link to the PR where this was changed so it is easier to track back this decision.

PR #2174, opened 2026-03-10
Tweak telemetry collection

xavdid-stripe:
### Why?

We've long collected detailed platform information about calls made using the SDK. But a user brought to our attention that we're probably collecting more than we need. To fix, we're no longer sending the full `uname` output and only collecting data we actually care about. Additionally, users can opt out of sending platform information using the existing telemetry opt-out methods.

### What?

- remove unused `publisher` User-Agent key
- condense important data from `uname` into the `platform` key
- tests

### See Also

- [stripe/stripe-php#2015](https://github.com/stripe/stripe-php/issues/2015)
- [doc](https://docs.google.com/document/d/13fq-DtC_6WTfMyyI8b-Gq6jensyECa5K0X3WPL9fZJU/edit?tab=t.0)
- [RUN_DEVSDK-2244](https://go/j/RUN_DEVSDK-2244)

### DIFF
--- a/src/main/java/com/stripe/net/HttpClient.java
+++ b/src/main/java/com/stripe/net/HttpClient.java
@@ -201,15 +201,7 @@ public abstract class HttpClient {
   }
 
   static String buildXStripeClientUserAgentString(String aiAgent) {
-    String[] propertyNames = {
-      "os.name",
-      "os.version",
-      "os.arch",
-      "java.version",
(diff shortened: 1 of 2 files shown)


mbroshi-stripe (approved):

bollie-dev:
Very useful. Thanks guys.

PR #2200, opened 2026-04-04
Add Ref<T> type for typed v2 object references

jar-stripe:
### Why?

Custom objects need typed references to other Stripe objects. A reference has the wire shape `{type, id, url}` and in the SDK becomes `Ref<T>` — a container that knows the target type and can fetch it.

### What?

- `com.stripe.model.v2.Ref<T extends StripeObject>` with `@SerializedName` annotations for `type`, `id`, `url`
- Implements `StripeActiveObject` — `responseGetter` is auto-injected by `StripeResponseGetterSettingTypeAdapterFactory` during GSON deserialization
- Transient `targetType` field set by `StripeCollectionItemTypeSettingFactory` to resolve type erasure
- `fetch()` calls `responseGetter.request(new ApiRequest(GET, url), targetType)` — same pattern as `Event.fetchRelatedObject()`
- Throws `IllegalStateException` when responseGetter is null

### See Also

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/com/stripe/model/v2/Ref.java
+++ b/src/main/java/com/stripe/model/v2/Ref.java
@@ -1,0 +1,52 @@
+package com.stripe.model.v2;
+
+import com.google.gson.annotations.SerializedName;
+import com.stripe.StripeClient;
+import com.stripe.exception.StripeException;
+import com.stripe.model.StripeObject;
+import com.stripe.net.ApiMode;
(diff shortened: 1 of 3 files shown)


mbroshi-stripe (approved):

jar-stripe:
Changing direction for now.

Issue #1845, opened 2024-08-06
`toJson` fails to create exactly the same json as the Stripe HTTP API


decoursin:
### Describe the bug

The [`InvoiceLineItem.getUnitAmountExcludingTax`](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/InvoiceLineItem.java#L167) returns a BigDecimal. While the Stripe API HTTP Server returns the [`unit_amount_excluding_tax`](https://docs.stripe.com/api/invoice-line_item/object#invoice_line_item_object-unit_amount_excluding_tax) as a string.

As a result, when the InvoiceLineItem is serialized using [`toJson`](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/StripeObject.java#L67), it returns a completely different format than the Stripe API Http Server. They should be the same.

### To Reproduce

1. Create an invoice and 1 invoice item.
2. Serialize that invoice using `toJson`.
3. View the serialize invoice item, and see that it's serialized as a json number rather than a string.

### Expected behavior

Serializing an InvoiceLineItem using `toJson` is suppose to replicate the Stripe API HTTP server 100% exactly, and it's not doing that.

### OS

All

### Java version

Java 17

### stripe-java version

v26.4.0

### API version

2024-06-20

xavdid-stripe:
Thanks for filing! We've copied to our internal bug tracker and will address it when we have a chance.

decoursin:
Great, thank you

parfentjev:
I'm using stripe sdk models to build mocks and stubs.

I observe the same issue with `unit_amount_decimal` and other similar fields. Moreover, _some_ `BigDecimal` values want to be treated as numbers (e.g., `percent_off` in coupon details). This dichotomy means that I cannot simply create a type adapter to serialize all `BigDecimal` values as `String`.

It's a major issue for automation testing.

Maybe there is some workaround I'm not aware of?

PR #2207, opened 2026-04-19
Fix BigDecimal serialization in StripeObject.toJson()

IamPritamAcharya:
### Why?

`StripeObject.toJson()` uses a Gson instance that does not serialize `BigDecimal`
values as strings. This leads to inconsistency with Stripe API expectations,
where decimal values are typically represented as strings to preserve precision.

Before this change:
{"amount": 25.50}

After this change:
{"amount": "25.50"}

This fix ensures consistent serialization behavior and avoids potential precision issues.

---

### What?

- Added a custom `TypeAdapter<BigDecimal>` to serialize values using `toPlainString()`
- Registered the adapter in `StripeObject.PRETTY_PRINT_GSON`
- Added a regression test (`StripeObjectTest`) to verify that `BigDecimal` values
  are serialized as JSON strings

---

### See Also

- Related issue: https://github.com/stripe/stripe-java/issues/1845

Note: Full test suite requires `stripe-mock` running locally. Unit tests pass and
this change does not affect API request logic.

### DIFF
--- a/src/main/java/com/stripe/model/StripeObject.java
+++ b/src/main/java/com/stripe/model/StripeObject.java
@@ -4,15 +4,37 @@ import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
 import com.stripe.net.ApiMode;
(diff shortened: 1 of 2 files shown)


IamPritamAcharya:
Closing this PR after further investigation.

The original approach used a global BigDecimal TypeAdapter, but it turns out
that Stripe models require field-specific handling — some BigDecimal fields
are expected as strings while others are numbers.

A global change would introduce incorrect serialization for certain fields.

Will revisit with a more targeted approach if needed.

Issue #2205, opened 2026-04-17
`toBuilderFullCopy()` does not copy all fields — `stripeContext` (and others) are missing


cchangenot:
### Describe the bug

**Library:** `stripe-java` `32.0.0`
**File:** `com/stripe/net/RequestOptions.java`

### Description

The `toBuilderFullCopy()` method is documented as *"Convert request options to builder, copying all options"*, but it does **not** copy all fields from the `RequestOptions` instance.
Specifically, the `stripeContext` and `stripeRequestTrigger` fields are **not copied**, which can lead to unexpected behavior when using the builder to create a modified copy of the original `RequestOptions`.

### Steps to Reproduce

```java
RequestOptions original = RequestOptions.builder()
    .setApiKey("sk_test_xxx")
    .setStripeContext("acct_context_123")
    .setStripeRequestTrigger("some_trigger")
    .build();

RequestOptions copy = original.toBuilderFullCopy().build();

System.out.println(original.getStripeContext());         // "acct_context_123"
System.out.println(copy.getStripeContext());             // null ← BUG
System.out.println(original.getStripeRequestTrigger()); // "some_trigger"
System.out.println(copy.getStripeRequestTrigger());      // null ← BUG
```

### Expected behavior

`toBuilderFullCopy()` should produce a builder that, when `.build()` is called, results in a `RequestOptions` **identical** to the original, including all fields.

### Code snippets

```Java

```

### OS

Fedora

### Java version

Java 25

### stripe-java version

v32.0.0

### API version

2026-03-25.dahlia

IamPritamAcharya:
Update: This issue has been fixed and merged!

mbroshi-stripe:
Just close the loop here, the fix was released in https://github.com/stripe/stripe-java/releases/tag/v32.1.0. Thanks again for your submission!

PR #2208, opened 2026-04-19
Fix missing stripeContext in RequestOptions.toBuilderFullCopy()

IamPritamAcharya:
### Why?

`RequestOptions.toBuilderFullCopy()` does not copy the `stripeContext` field,
causing it to be lost when rebuilding a `RequestOptions` instance.

This leads to unexpected behavior where request context metadata is silently dropped.

Before this change:
opts.toBuilderFullCopy().build().getStripeContext() == null

After this change:
opts.toBuilderFullCopy().build().getStripeContext() == original value

---

### What?

- Added `.setStripeContext(this.stripeContext)` in `toBuilderFullCopy()`
- Added a regression test to ensure `stripeContext` is preserved during copy

---

### See Also

- https://github.com/stripe/stripe-java/issues/2205

## Changelog
- Fixes a bug where an existing `stripeContext` was being reset to `null` when calling `RequestOptions#toBuilderFullCopy()`

### DIFF
--- a/src/main/java/com/stripe/net/RequestOptions.java
+++ b/src/main/java/com/stripe/net/RequestOptions.java
@@ -168,6 +168,7 @@ public class RequestOptions {
             .setClientId(this.clientId)
             .setIdempotencyKey(this.idempotencyKey)
             .setStripeAccount(this.stripeAccount)
+            .setStripeContext(this.stripeContext)
             .setStripeRequestTrigger(this.stripeRequestTrigger)
             .setConnectTimeout(this.connectTimeout)
             .setReadTimeout(this.readTimeout)
--- a/src/test/java/com/stripe/net/RequestOptionsTest.java
+++ b/src/test/java/com/stripe/net/RequestOptionsTest.java
@@ -47,25 +47,25 @@ public class RequestOptionsTest {
     assertNull(optsRebuilt.getBaseUrl());
   }
 
-    @Test
-    public void testToBuilderFullCopy() {
+  @Test
+  public void testToBuilderFullCopy() {
     RequestOptions opts =
         RequestOptionsBuilder.unsafeSetStripeVersionOverride(
-            RequestOptions.builder()
-                .setApiKey("sk_foo")
-                .setBaseUrl("http://example.com")
-                .setClientId("123")
-                .setIdempotencyKey("123")
-                .setStripeAccount("acct_bar")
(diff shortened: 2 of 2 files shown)


src/test/java/com/stripe/net/RequestOptionsTest.java
```diff
@@ -47,29 +47,48 @@ public void testPersistentValuesInToBuilder() {
...
     RequestOptions optsRebuilt = opts.toBuilderFullCopy().build();
 
     assertEquals(opts, optsRebuilt);
-  }
+    assertEquals("acct_context_123", optsRebuilt.getStripeContext());
+    assertEquals("some_trigger", optsRebuilt.getStripeRequestTrigger());
+    }
+
+    @Test
+    public void testToBuilderFullCopy_preservesStripeContext() {
```
mbroshi-stripe: Nit: This test is redundant now that you've added `setStripeContext` to `testToBuilderFullCopy`

src/test/java/com/stripe/net/RequestOptionsTest.java
```diff
@@ -47,29 +47,48 @@ public void testPersistentValuesInToBuilder() {
     assertNull(optsRebuilt.getBaseUrl());
   }
 
-  @Test
-  public void testToBuilderFullCopy() {
+    @Test
+    public void testToBuilderFullCopy() {
```
mbroshi-stripe: Remove extraneous whitespace. This will fail linting. You can run `just format` to fix it

src/main/java/com/stripe/net/RequestOptions.java
```diff
@@ -168,6 +168,7 @@ public RequestOptionsBuilder toBuilderFullCopy() {
             .setClientId(this.clientId)
             .setIdempotencyKey(this.idempotencyKey)
             .setStripeAccount(this.stripeAccount)
+            .setStripeContext(this.stripeContext)
```
mbroshi-stripe, line 171: Great catch!

mbroshi-stripe (commented):
Thanks for catching this and submitting a fix! I left a couple minor comments to address

IamPritamAcharya:
@mbroshi-stripe 
Addressed feedback:
- Removed redundant test
- Fixed formatting issues

Thanks for the review!

mbroshi-stripe (approved):

PR #2219, opened 2026-05-21
Emit warning when `stripe-notify` header is present in response

xavdid:
### Why?

We've noticed a trend where users (and their agents) will sometimes reach for suboptimal API endpoints/methods because of outdated information. For example, using `/v1/accounts` when `/v2/core/accounts` is available and has a superset of the functionality from v1.

In an effort to help steer users towards best practices, we're introducing a new header to provide feedback during the development process. The SDK is responsible for surfacing this header, if present.

Once enabled internally, the header will be returned by the server for:

- all testmode calls
- livemode calls made by an agent

We may tweak that going forwards. But no matter the conditions, the SDKs will act as a thin pipe.

### What?

- Emit warning when we see the `Stripe-Notice` header in a response. **Note** this is the first warning we're emitting in this SDK, so make sure it's the correct way to do that!
- add tests

### See Also

- https://go/agent-steering
- [DEVSDK-2430](https://go/j/browse/DEVSDK-2430)

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -21,8 +21,11 @@ import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.logging.Logger;
 
 public class LiveStripeResponseGetter implements StripeResponseGetter {
+  private static final Logger logger = Logger.getLogger("Stripe");
+
   private final HttpClient httpClient;
   private final StripeResponseGetterOptions options;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -255,6 +262,10 @@ public InputStream requestStream(
     return this.requestStream(new ApiRequest(baseAddress, method, path, params, options));
   }
 
+  private static void maybeEmitStripeNotice(HttpHeaders headers) {
+    headers.firstValue("Stripe-Notice").ifPresent(logger::warning);
```
mbroshi-stripe, line 266: Very slick!

mbroshi-stripe (approved):

PR #2229, opened 2026-06-02
Fix V2 namespace singularization

jar-stripe:
### Why?

The codegen incorrectly singularized V2 namespace-promoted service names (`PaymentService` instead of `PaymentsService`). This was a latent bug that only manifested when new V2 namespaces were added.

### What?

- Renames `v2.PaymentService` to `v2.PaymentsService`

## Changelog

- ⚠️ Renames `v2.PaymentService` to `v2.PaymentsService` to more closely match our API naming


xavdid (approved):

xavdid:
As far as I can tell, all the changes in this PR are already captured in the auto PR; closing.

I'll make sure this is reflected in the final java changelog

PR #2244, opened 2026-07-10
Replace source hash with Telemetry UUID

xavdid:
### Why?

Recently, we introduced the `source` hash in the user-agent header to help our support team track down exactly where incoming requests originated from.

They've reported back to us that this hasn't been as useful as we wanted, so we're going a different direction.

Instead, we're adopting an industry-standard method of persisting a small UUID on disk that we include with subsequent requests. It's easy for users to read this file and it contains no PII. It also honors our existing telemetry opt outs.

It's meant to be lightweight and failure tolerant.

### What?

- remove `source` key from user-agent header
- add `telemetry_id` to the user-agent header
- adjust tests accordingly

### See Also

- original request: [DEVSDK-3131](https://go/j/DEVSDK-3131)
- update request: [RUN_DEVSDK-2563](https://go/j/RUN_DEVSDK-2563)

### DIFF
--- a/src/test/java/com/stripe/net/TelemetryIdTest.java
+++ b/src/test/java/com/stripe/net/TelemetryIdTest.java
@@ -1,0 +1,118 @@
+package com.stripe.net;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.stripe.BaseStripeTest;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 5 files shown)


src/test/java/com/stripe/net/TelemetryIdTest.java
```diff
@@ -0,0 +1,106 @@
...
+    Files.write(filePath, existingId.getBytes("UTF-8"));
+
+    String id = TelemetryId.get();
+    assertEquals(existingId, id);
+  }
+
+  @Test
+  public void testGetConfigDirReturnsPathContainingStripe() {
+    // On any supported platform, the config dir should contain "stripe" (or "Stripe" on Windows)
+    Path configDir = TelemetryId.getConfigDir();
```
jar-stripe, line 87: may be a small nit, but it looks like these tests are a bit inconsistent; dotnet for example contains different tests and will explicitly filter out windows, and ruby and php follow that test pattern but without the windows tests.  this is probably ok for now but worth considering a follow up (and maybe some CI tests on windows)

jar-stripe (approved):

PR #2250, opened 2026-07-16
Fix flakey testTelemetryEnabled test

jar-stripe:
### Why?
The `testTelemetryEnabled` test will fail frequently which can delay releases and other PR merges.  This has happened more frequently in the past several months.  The condition that is failing tests that the request timeout ms is _strictly greater than_ the 30ms body delay configured in the test.  After adding an assertion message that prints out the actual value, it looks like the mock request is completing in 30ms
```
org.opentest4j.AssertionFailedError: 30 is not greater than 30 ==> expected: <true> but was: <false>
```

I suspect this is a rounding/truncating issue, but either way we should accept any number equal or greater to the delay here.

### What?
- Modifies assert calls that test request duration to be greater than or equal to the delay value
- Adds assert message to help troubleshoot any future issues (if this is not the complete fix).


xavdid (approved):
nice!

jar-stripe:
> nice!

thanks!  we'll see if this was entirely the issue 🤞

Issue #2201, opened 2026-04-06
Provide a way for deserializing EventBridge events


AlessandroBagnoli:
Hi there!

As noted in the [Stripe EventBridge documentation](https://docs.stripe.com/event-destinations/eventbridge#event-structure), events delivered via EventBridge have a different schema than standard webhooks due to an additional wrapping layer.

Currently, it seems the SDK's deserialization logic is tightly coupled with signature verification. Since EventBridge events don't use signatures in the same way, there isn't a straightforward way to use the SDK to parse these events into strongly-typed objects.

Proposed Solution:
It would be great if the SDK exposed a method to deserialize these specific EventBridge payloads directly.

Question:
Is there currently a supported way to do this that I might have missed, or should this be considered a new feature request?

Many thanks!

xavdid:
Hey @AlessandroBagnoli! Thanks for the suggestion.

The current recommendation is to use `EventNotification.fromJson(webhookBody, client)` instead of `client.parseEventNotification(webhookBody, sigHeader, WEBHOOK_SECRET)`. The latter calls the former under the hood, but assumes validation is already complete (or unnecessary). The rest of the calling code (casting from an `EventNotification` to a more specific type) is the same.

This isn't documented well, so we'll work on surfacing that!

AlessandroBagnoli:
Thanks @xavdid ! It would be lovely to have a way to do this without having to pass an instance of a client (similarly to how the `Webhook` class allows this with static methods)

xavdid:
That's currently not planned because `EventNotification`s need to be able to fetch their corresponding events and related objects without relying on global state. So they take a client, allowing users to customize retrieval behavior separate from the rest of an integration.

That said, we're always open to ideas. Can you talk more about your use case?

xavdid:
@AlessandroBagnoli It was pointed out to me internally that you were probably using Snapshot events (the events that we've had for a long time) not `EventNotifications` (new "thin" events). If that's the case, `fromJson` won't help you since it's the wrong type-- sorry about that!

In that case, we do need better tools to work with Event bridge and that work is already planned 😁

AlessandroBagnoli:
> @AlessandroBagnoli It was pointed out to me internally that you were probably using Snapshot events (the events that we've had for a long time) not `EventNotifications` (new "thin" events). If that's the case, `fromJson` won't help you since it's the wrong type-- sorry about that!
> 
> In that case, we do need better tools to work with Event bridge and that work is already planned 😁

[...]

PR #2256, opened 2026-07-24
add/adjust event parsing helpers

xavdid:
### Why?

Historically, our webhook parsing logic has been closely tied to signature validation (because the expectation is that users verify webhooks' authenticity before use). But, as the webhook handling experience has grown in complexity over time, our helper methods haven't kept up.

In addition to the classic parse+verify & handle, we also support / encourage :

- using cloud providers (like [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid)), which don't require validation at all
- validating & ACKing an event quickly, then queuing it up for later processing
- unit tests w/ signed test payloads

This gives us 4 use cases that each need a subset of helper methods:

| case                                  | `verify` | `parse` | `parseAndVerify` | `signTestPayload` |
| ------------------------------------- | -------- | ------- | ---------------- | ----------------- |
| process immediately                   |          |         | ✅               | ✅                |
| verify, queue, process later          | ✅       | ✅      |                  | ✅                |
| process immediately (cloud provider)  |          | ✅      |                  |                   |
| queue, process later (cloud provider) |          | ✅      |                  |                   |

To support these workflows, we need all of these methods publicly available in every SDK:

1. `constructEvent(payload, signature, ...)` (parse + verify)
2. `parseEventNotification(payload, signature, ...)` (parse + verify)
3. `constructEventWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
4. `parseEventNotificationWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
5. `verify(payload, signature, secret, ...)`
6. `signTestPayload(payload, secret)`

The exact naming & arguments vary by language, but each SDK should provide roughly the same capabilities.

> [!IMPORTANT]
> For reviewers: the exact error messages are a little inconsistent between SDKs right now. Everything throws in the right place, but we're still iterating on some of the naming. We're trying out "thin event notification" as a descriptor on some of these surfaces. We're planning to have this all cleaned up by the GA launch of interop events; ignore exact messages for now.

### What?

> [!NOTE]
> these notes generic across all the SDKs except where noted; see `Changelog` below for the SDK-specific list of changes

- add missing event handling methods to `StripeClient` & `Webhook` classes, where relevant.
- Add an internal-facing `maybe_extract_from_cloud_provider_envelope` for handling wrapped events without the user having to think about it
- centralize some event parsing logic; adjusting accepted types for some methods
- adjust some error messages
- add tests
- update docstrings

### See Also

- [stripe/stripe-java#2201](https://github.com/stripe/stripe-java/issues/2201)
- [DEVSDK-3085](https://go/j/DEVSDK-3085)
- [DEVSDK-3210](https://go/j/DEVSDK-3210)

## Changelog

- Added methods that return their respective `Event`/`EventNotification` class instances without verifying authenticity. Use them when you've previously verified an event (e.g. you verified, put the event in a queue, and are now processing). Supports events from [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid) natively.
  - `Webhook.constructEventWithoutVerification(payload)`
  - `StripeClient.constructEventWithoutVerification(payload)`
  - `StripeClient.parseEventNotificationWithoutVerification(payload)`
- Added `Webhook.Signature.generateSignatureHeader(payload, secret)`, which computes a full `Stripe-Signature` header for the given payload. Useful for unit tests!


src/main/java/com/stripe/model/v2/core/EventNotification.java
```diff
@@ -91,12 +91,20 @@ public static class Reason {
...
-    JsonObject jsonObject = ApiResource.GSON.fromJson(payload, JsonObject.class).getAsJsonObject();
+    return EventNotification.fromJson(
+        ApiResource.GSON.fromJson(payload, JsonObject.class).getAsJsonObject(), client);
+  }
 
+  /**
+   * Helper for constructing an Event Notification. Doesn't perform signature validation, so you
+   * should use {@link com.stripe.StripeClient#parseEventNotification} instead for initial handling.
+   * This is useful in unit tests and working with EventNotifications that you've already validated
+   * the authenticity of.
```
jar-stripe: very small nit: sentence ends in a preposition.

src/main/java/com/stripe/model/v2/core/EventNotification.java
```diff
@@ -91,12 +91,20 @@ public static class Reason {
...
+   */
+  public static EventNotification fromJson(JsonObject jsonObject, StripeClient client) {
     if (jsonObject.has("object") && "event".equals(jsonObject.get("object").getAsString())) {
```
jar-stripe: this wasn't introduced in this PR, but just to confirm, thin events dont have an object field?

src/main/java/com/stripe/net/Webhook.java
```diff
@@ -69,28 +71,90 @@ public static Event constructEvent(
...
+    // AWS
+    // https://docs.stripe.com/event-destinations/eventbridge#event-structure
+    if (root.has("detail")) {
+      return root.get("detail").getAsJsonObject();
+    }
+
+    // Azure
+    // https://docs.stripe.com/event-destinations/eventgrid#event-structure
+    if (root.has("specversion")) {
+      return root.get("data").getAsJsonObject();
```
jar-stripe, line 143: is there any way data could be null here?  if so, add root.has("data") to the conditinal

jar-stripe (approved):
approving to unblock, but the comments are worth considering (esp the potential null dereference)

src/main/java/com/stripe/model/v2/core/EventNotification.java
xavdid: they do, but it's set to `v2.core.event`, not `event`

jar-stripe: It would be safer here to check for `v2.core.event` and reject all others; JsonObjects are exposed by other resources via `getRawJsonObject()`, so someone could accidentally pass arbitrary JsonObject objects into EventNotification.fromJson and we want to catch that early.

xavdid: fair! i'll update

src/main/java/com/stripe/model/v2/core/EventNotification.java
xavdid: haha i'll see if I can fix it in a way I don't hate

src/main/java/com/stripe/net/Webhook.java
xavdid: we chatted in slack- it's not possible for a well-formed envelope, but we're not guaranteed that we'll get those. So we'll be more defensive.

PR #2281, opened 2026-08-27
Merge to beta

jar-stripe:
### Why?
Routine merge of master changes into the `beta` release channel.

### What?
Merged master into `beta` using sdk-codegen's automerge tooling (`tools/automerge/java-beta.yaml`).
Net change is limited to `.github/workflows/ci.yml`, the new `.github/workflows/zizmor.yml` and `.github/zizmor.yml`, and the GA CHANGELOG entries.

One manual resolution: took `beta`'s copy of `src/main/java/com/stripe/param/CustomerSessionCreateParams.java`. The automerge chunk merge emitted a second copy of the `ActiveEntitlements` param class and its builder setter, which both branches already had — a duplicate class and method in one scope, which does not compile. `beta`'s copy is a strict superset of master's (it also carries `TaxIdElement`), so taking it loses nothing.


jar-stripe:
Closing — created with the wrong head branch (`master` instead of `jar/merge-*`) due to a missing `--head` flag. Reopening the correct PR now.

PR #2284, opened 2026-08-28
Harden API requestor code against malicious URLs

xavdid:
### Why?

If an attacker got access to a user's webhook secret (or the user wasn't validating incoming webhook signatures) _and_ a user was calling the built-in `event_notification.fetch_related_object()` method, a malicious webhook payload could cause a user integration to make an authenticated request to an attacker-controlled domain (leaking the secret key)

To solve, we're being more careful during url construction to ensure user input can't be used to send requests to non-`stripe.com` domains (on a per-request basis).

### What?

- add path validation method and call it before the final URL is built
- Encode the `event.id` path param in `fetch_event()` to match the sanitation in the rest of the generated endpoints
- add tests

### See Also

- [RUN_DEVSDK-2808](https://go/j/RUN_DEVSDK-2808)

### DIFF
--- a/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
+++ b/src/main/java/com/stripe/net/LiveStripeResponseGetter.java
@@ -476,6 +476,31 @@ public class LiveStripeResponseGetter implements StripeResponseGetter {
     }
   }
 
+  /**
+   * Asserts that a request path is origin-relative: that it begins with a single {@code "/"}.
+   *
+   * <p>The absolute URL is built by concatenating a base URL onto this path, and no base URL ends
+   * in a slash. A path like {@code "@evil.example/v1/x"} or {@code ".evil.example/v1/x"} would
+   * therefore land inside the authority component and send the request -- {@code Authorization}
+   * header included -- to a host of the path's choosing. Some request paths originate in remote
+   * data (a webhook body's {@code related_object.url}, a collection's {@code url}, a response's
+   * {@code next_page_url}), so the path cannot be assumed to be well-formed.
+   *
+   * <p>A single leading slash is sufficient: it terminates the authority component, after which
+   * nothing in the path can extend it. Deliberately not using {@link java.net.URI} to parse -- it
+   * enforces RFC 2396 strictly and would reject paths containing characters that callers have
+   * always been able to send.
+   *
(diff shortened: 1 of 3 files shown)


src/main/java/com/stripe/net/LiveStripeResponseGetter.java
```diff
@@ -476,6 +476,31 @@ public void validateRequestOptions(RequestOptions options) {
...
+  /**
+   * Asserts that a request path is origin-relative: that it begins with a single {@code "/"}.
+   *
+   * <p>The absolute URL is built by concatenating a base URL onto this path, and no base URL ends
+   * in a slash. A path like {@code "@evil.example/v1/x"} or {@code ".evil.example/v1/x"} would
+   * therefore land inside the authority component and send the request -- {@code Authorization}
+   * header included -- to a host of the path's choosing. Some request paths originate in remote
+   * data (a webhook body's {@code related_object.url}, a collection's {@code url}, a response's
+   * {@code next_page_url}), so the path cannot be assumed to be well-formed.
+   *
```
jar-stripe, line 485: small nit, this feels like a lot of exposition above a method that just checks if a path starts with a slash.

src/test/java/com/stripe/net/OriginRelativePathTest.java
```diff
@@ -0,0 +1,74 @@
+package com.stripe.net;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import com.stripe.BaseStripeTest;
+import org.junit.jupiter.api.Test;
+
+/**
+ * The absolute request URL is built by concatenating a base URL onto a relative path, and no base
```
jar-stripe: repeats comment content from LiveStripeRequestGetter

jar-stripe (approved):
Approving to unblock.  Change looks good, but the comments are a bit length and repetative

src/main/java/com/stripe/net/LiveStripeResponseGetter.java
xavdid: On the one hand, sure. On the other hand, this reasoning behind the validation is pretty useful (and feels like a "why" worth documenting). But, i'll shorten it.

